PR #3, opened 2016-07-11
Add DCO check

duglin:
Add a travis-ci check to ensure all PR's commits are signed with a DCO.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,4 @@
+language: bash
+
+before_install:  
+  - ./.validateDCO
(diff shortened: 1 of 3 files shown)


MHBauer:
Can you add a definition of what this means?

duglin:
@MHBauer Done

MHBauer:
I'm not sure what you changed, but to be explicit:

1. define DCO
1. explain or link to what it means to agree to it
1. explain how to do it (git commit -s)
1. explain that signing off means you are agreeing to the dco

I can do research, but we shouldn't make people do research, or blow off the significance of what signing off is intending to mean.

PR #4, opened 2016-07-11
Initial commit of service controller.

bmelville:


MHBauer:
let's not commit binaries. `cf/service_controller/service_controller`

`go build` in the `cf/service_controller` works for me.

bmelville:
Yikes, totally missed that, updated in the commit.

duglin:
can you move this to be under the "kube" dir since this will be a kube-focused impl? I was saving the 'cf' dir for the ruby/CF extracted code.

bmelville:
Understood, please take another look.

duglin:
LGTM

PR #10, opened 2016-07-14
split up the struct by domain

MHBauer:
- also ensure we satisfy the interface

### DIFF
--- a/k8s/service_controller/server/service_storage.go
+++ b/k8s/service_controller/server/service_storage.go
@@ -12,21 +12,38 @@ type ServiceBroker struct {
 	Password string `json:password""`
 }
 
-type ServiceStorage interface {
+// The Broker interface provides functions to deal with brokers.
+type Broker interface {
 	ListBrokers() ([]*ServiceBroker, error)
 	GetBroker(string) (*ServiceBroker, error)
 	GetInventory(string) (*model.Catalog, error)
 	AddBroker(*ServiceBroker, *model.Catalog) error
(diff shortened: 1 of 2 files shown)


MHBauer:
gonna need a bot to tell me I need to sign it @duglin

duglin:
LOL just get in the habit of signing all of your commits  :-)

k8s/service_controller/server/service_storage.go
```diff
@@ -12,21 +12,38 @@ type ServiceBroker struct {
...
+type Broker interface {
 	ListBrokers() ([]*ServiceBroker, error)
 	GetBroker(string) (*ServiceBroker, error)
 	GetInventory(string) (*model.Catalog, error)
 	AddBroker(*ServiceBroker, *model.Catalog) error
 	DeleteBroker(string) error
+}
 
+// The Servicer interface provides functions to deal with services.
+type Servicer interface {
```
duglin: I'd prefer if `Servicer` was based on the word `Instance` instead of `Service` because the diff between a service and service broker is thin to a newbie and to avoid confusion I'd prefer to use the terms we use in other spots: ServiceBroker, ServiceInstance and ServiceBinding.

MHBauer: Agree. Thanks.

MHBauer:
Split up, because I believe in modularity, and while we may not agree on the totality of the interface yet, let's try to stay organized.

duglin:
LGTM

PR #13, opened 2016-07-14
start a k8s storage backend

MHBauer:
compiles and runs successfully but does not actually do anything.

nil and not-implemented backends for testing purposes later, as well as to provide a template for anyone else wishing to implement a backend.


MHBauer:
with this in, we can start implementing each endpoint individually.

duglin:
LGTM

PR #14, opened 2016-07-14
Minor changes so we use the local image

duglin:


k8s/service_controller/service-controller.yaml
```diff
@@ -14,7 +14,7 @@ spec:
     spec:
       containers:
       - name: service-controller
-        image: gcr.io/bmelville-test/service-controller:latest
```
MHBauer: not LGTM if you haven't pulled this image already.

PR #16, opened 2016-07-15
Try using pullapprove

duglin:


duglin:
LGTM

duglin:
Approved

.pullapprove.yml
```diff
@@ -0,0 +1,9 @@
+approve_by_comment: true
+approve_regex: LGTM
```
MHBauer: according to docs [here](http://docs.pullapprove.com/yaml/approve_regex/)
NOT LGTM

PR #31, opened 2016-07-20
First pass at aligning with `cf create-service-broker`

duglin:

### DIFF
--- a/k8s/service_controller/Makefile
+++ b/k8s/service_controller/Makefile
@@ -1,9 +1,9 @@
-# .PHONY: all binary test-unit clean server
+# .PHONY: all binary test-unit test clean server
 
 EXE:=service_controller
 IMAGE:=service-controller
 GO_FILES=$(shell find . -name \*.go -print | grep -v _test.go)
-TEST_DIRS=$(shell find . -name \*_test.go -exec dirname {} \; | sort -u)
+TEST_UNIT_DIRS=$(shell find . -name \*_test.go -exec dirname {} \; | grep -v ./tests | sort -u)
 
 all: binary
 
@@ -15,7 +15,7 @@ $(EXE): $(GO_FILES)
 	  -v ${CURDIR}:/src/github.com/cncf/servicebroker/k8s/service_controller \
 	  -e GOPATH=/ \
 	  golang \
-	  sh -c "go get ./... && go build ."
+	  sh -c "go get -d ./... && go build ."
 	docker build -t $(IMAGE):latest .
 
 test-unit: binary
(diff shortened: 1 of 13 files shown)


k8s/service_controller/model/service_broker.go
```diff
@@ -0,0 +1,11 @@
+package model
+
+type ServiceBroker struct {
+	// json info from create docs
+	// http://apidocs.cloudfoundry.org/239/service_brokers/create_a_service_broker.html
+	Name         string `json:"name"`
+	BrokerURL    string `json:"broker_url"`
```
MHBauer: is there going to be any negotiation on this? It seems overly verbose to call something 'broker_url' in a 'broker' object.

duglin: I think Google is listing some things they would like to change, but my assumption is that for the immediate future we should align with what's in the CF docs/specs just to get things working - then we can see how people feel about changing things.

k8s/service_controller/Makefile
```diff
@@ -24,12 +24,25 @@ test-unit: binary
...
+	  sh -c "go get -d ./... && make --no-print-directory test-unit-local"
 	
 test-unit-local:
-	@TEST_DIRS=$(shell find . -name \*_test.go -exec dirname {} \; | sort -u)
+	@TEST_UNIT_DIRS="$(shell find . -name \*_test.go -exec dirname {} \; | sort -u)"
 	@echo "\n### Running the unit tests"
-	@go test $(TEST_DIRS)
+	@go test $(TEST_UNIT_DIRS)
+
+test: binary
```
MHBauer: why not `test-docker` and `test`? I would expect more people to have go installed than docker.

k8s/service_controller/server/controller.go
```diff
@@ -63,18 +63,20 @@ func (c *Controller) Inventory(w http.ResponseWriter, r *http.Request) {
...
-	var b ServiceBroker
+	var b model.ServiceBroker
 	err := utils.BodyToObject(r, &b)
 	if err != nil {
 		fmt.Printf("Error unmarshaling: %#v", err)
 		utils.WriteResponse(w, 400, err)
 		return
 	}
 
+	fmt.Printf("REQ: %v\n", b)
```
MHBauer: this leads to another issue/potential pr. We need a logging service/library.

k8s/service_controller/tests/sb_create_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"testing"
+
+	"bytes"
+	"net/http"
+	"os/exec"
+)
+
+func Test_SB_create_basic(t *testing.T) {
```
MHBauer: delete or uncomment

MHBauer: in fact, you can get rid of this whole test file if this doesn't get uncommented.

k8s/service_controller/Makefile
duglin: let's deal with that in a separate PR - I want to get some of these re-org changes in.

duglin:
LGTM

PR #34, opened 2016-07-21
Setup a go broker

duglin:
and move the node one into the brokers folder


duglin:
LGTM

PR #35, opened 2016-07-21
Add a real create test with the go broker

duglin:
This should support the CF cli->CC request - need to generate good data for the response next.


duglin:
LGTM

PR #40, opened 2016-07-22
Coverage

MHBauer:
not sure this would be worth it, other than the makefile cleanup.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,7 @@
-sudo: required
 language: go
-
-services:
-  - docker
 
 script:
   - make all
-
   
 after_success:
   - bash generate-coverage.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,13 @@
 sudo: required
-language: bash
+language: go
 
 services:
   - docker
 
 script:
   - make all
+
+  
+after_success:
+  - bash generate-coverage.sh
+  - bash <(curl -s https://codecov.io/bash)
--- a/k8s/service_controller/Makefile
+++ b/k8s/service_controller/Makefile
@@ -14,10 +14,10 @@ binary: $(EXE)
 $(EXE): $(SRC_FILES)
 ifeq (,$(shell which go))
 	@$(MAKE) --no-print-directory binary-docker
+	docker build -t $(IMAGE):latest .
 else
 	@$(MAKE) --no-print-directory binary-local
 endif
-	docker build -t $(IMAGE):latest .
 
 binary-docker:
 	docker run \
@@ -28,6 +28,7 @@ binary-docker:
 	  sh -c "go get -d ./... && make binary-local"
 
 binary-local:
+	go get -v -d ./...
 	go build -o $(EXE) .
 
 #####################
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -1,8 +1,8 @@
-sudo: required
-language: bash
-
-services:
-  - docker
+language: go
```
duglin: Do we need this for codecov? One of the reasons I didn't want to tag this as a "go" project is because at some point I'm hoping to have the CF ruby code move into here too - which means we're not just go.

.travis.yml
```diff
@@ -1,8 +1,8 @@
...
 
 script:
   - make all
+  
+after_success:
+  - bash generate-coverage.sh
```
duglin: Could this be part of the lower-level Makefile so its go-specific?

k8s/service_controller/Makefile
```diff
@@ -14,10 +14,10 @@ binary: $(EXE)
 $(EXE): $(SRC_FILES)
 ifeq (,$(shell which go))
 	@$(MAKE) --no-print-directory binary-docker
+	docker build -t $(IMAGE):latest .
 else
 	@$(MAKE) --no-print-directory binary-local
 endif
-	docker build -t $(IMAGE):latest .
```
duglin: Why did you move this up?

duglin: Even on a local build I still want to build the image since at some point I expect it to be run via docker in the tests.

k8s/service_controller/brokers/go/Makefile
```diff
@@ -5,10 +5,10 @@ all: $(EXE)
 $(EXE): gobroker.go
 ifeq (,$(shell which go))
 	@$(MAKE) --no-print-directory binary-docker
+	docker build -t gobroker .
 else
 	@$(MAKE) --no-print-directory binary-local
 endif
-	docker build -t gobroker .
```
duglin: Ditto here - irrespective of whether we build with docker or not we should still create an image.

.travis.yml
MHBauer: can try using `gimme` manually, but travis is not set up to do multi language projects.

.travis.yml
MHBauer: I don't understand. It is a script.

k8s/service_controller/Makefile
MHBauer: because I don't want to need docker in travis.

duglin: we won't be able to avoid it at some point since we'll need to run kube for our tests and right now I think that means hyperkube in a container(s).  Embrace the inevitable  :-)

.travis.yml
duglin: But let's run it outside of the travis stuff since the code coverage will be go, or ruby, specific depending on which sub-project we're building/testing.  Moving it into the k8s portion will allow us to run the go version for just the k8s side of things.

.travis.yml
duglin: the current setup is setup to do multi languages - its doing go w/o travis knowing it.  So we should be able to add ruby for the CF side of the project as well.

Issue #28, opened 2016-07-19
Verify the cf cli APIs for user-defined services aligns with CF's defintion


duglin:
- [ ] Service Brokers
- [ ] Service Instances
- [ ] Service Bindings
- [ ] Services
- [ ] Service Plans

duglin:
add unit testcases too

bmelville:
I'm not sure if this is the right issue, but I have a PR in progress that aligns the service_controller API with the cloud controller API as defined here: https://apidocs.cloudfoundry.org/239/

bmelville:
The RESTful path surface is aligned (for the methods we support). Remaining work I've seen is:

1) Align Request/Response object model,
2) Identify remaining methods to support in the open service controller.

This is probably not a comprehensive list though.

shalako:
This does not appear to be related to the Open Broker API specification. Can we close it?

duglin:
if we remove the code then yes, but let's resolve https://github.com/openservicebrokerapi/servicebroker/issues/102 first

PR #44, opened 2016-07-26
First step to align the RESTful path surface of controller API with CF.

bmelville:
This is a work in progress, but has broad changes affecting the controller and all storage interfaces, so wanted to make sure it's out there for awareness.

References #28


vaikas-google:
LGTM

PR #45, opened 2016-08-02
get a broker and list all brokers

MHBauer:
- stuffed the created int64 into a string because it does not
   round-trip correctly through k8s
 - left the debugging print statements in
 - k8s type that extends the model servicebroker type

Signed-off-by: Morgan Bauer <mbauer@us.ibm.com>

### DIFF
--- a/k8s/service_controller/server/k8s/storage.go
+++ b/k8s/service_controller/server/k8s/storage.go
@@ -29,10 +29,11 @@ func CreateServiceStorage() server.ServiceStorage {
 // listSB is only used for unmarshalling the list of service brokers
 // for returning to the client
 type listSB struct {
-	Items []*model.ServiceBroker
+	Items []*k8sServiceBroker `json:"items"`
 }
 
 func (kss *K8sServiceStorage) ListBrokers() ([]*model.ServiceBroker, error) {
+	fmt.Println("listing all brokers")
 	// get the ServiceBroker
 
 	c := exec.Command("kubectl", "get", "ServiceBrokers", "-ojson")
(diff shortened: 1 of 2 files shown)


k8s/service_controller/server/k8s/storage.go
```diff
@@ -44,16 +45,36 @@ func (kss *K8sServiceStorage) ListBrokers() ([]*model.ServiceBroker, error) {
...
+	ret := make([]*model.ServiceBroker, 0)
+	for _, v := range lsb.Items {
+		ret = append(ret, v.ServiceBroker)
+	}
+	return ret, nil
 }
 
-func (kss *K8sServiceStorage) GetBroker(id string) (*model.ServiceBroker, error) {
-	return nil, fmt.Errorf("Not implemented yet")
+func (kss *K8sServiceStorage) GetBroker(name string) (*model.ServiceBroker, error) {
```
bmelville: Was this a bad merge? At this level the broker key is going to be the GUID rather than broker name.

k8s/service_controller/server/k8s/storage.go
```diff
@@ -29,10 +29,11 @@ func CreateServiceStorage() server.ServiceStorage {
 // listSB is only used for unmarshalling the list of service brokers
 // for returning to the client
 type listSB struct {
-	Items []*model.ServiceBroker
+	Items []k8sServiceBroker `json:"items"`
```
bmelville: May be my go naivety, why not list of pointers here?

k8s/service_controller/server/k8s/storage.go
MHBauer: That's actually an interesting issue. We can't lookup by guid in k8s (or I do not know how), so I'm using the name. You can see what I did in AddBroker, using the broker name twice.

k8s/service_controller/server/k8s/storage.go
MHBauer: That seems to work too. Not sure how big of a difference it is in this case.

(also, it's a *slice* of whatevers)

duglin:
LGTM

Issue #43, opened 2016-07-23
Create a CLI


duglin:
We probably can't use the CF cli for all operations so we should just create a cli for just the user facing service related operations we're going to support.

bmelville:
I'm going to be working on a CLI for at least the operations the service controller API supports right now.

bmelville:
FYI @vaikas-google

vaikas-google:
FYI, I have a simple CLI showing how it would look like with Cobra. Take a look and let me know your thoughts, it handles much of the cruft for us for free and is quite neat.

vaikas-google:
FYI, I'm going to be adding more complete API commands this week. Just wanted to make sure nobody else was working on it also :)

MHBauer:
@vaikas-google if you mean in the cli side, it's all yours. ping me if you want a review of something, even if it is incomplete.

vaikas-google:
I'd like to close this because I think the CLI is there. It's not complete, but I'd rather have smaller issues tracking specific items left to do?

bmelville:
SGTM, I'd love to see more service instance operations than create :)

PR #48, opened 2016-08-05
Add start of command line with example commands using cobra

vaikas-google:
#43 simple start showing how this would look like with the cobra.

### DIFF
--- a/cli/cmd/brokers.go
+++ b/cli/cmd/brokers.go
@@ -1,0 +1,61 @@
+package cmd
+
+import (
+	"fmt"
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	RootCmd.AddCommand(brokersCmd)
+	brokersCmd.AddCommand(createCmd)
+	brokersCmd.AddCommand(describeCmd)
+	brokersCmd.AddCommand(listCmd)
+	brokersCmd.AddCommand(deleteCmd)
+}
+
+var brokersCmd = &cobra.Command{
+	Use:   "brokers",
+	Short: "manage brokers associated with a service controller",
+	Long:  "Manage brokers associated with a service controller",
+}
+
+var createCmd = &cobra.Command{
+	Use:   "create",
+	Short: "Add a service broker to service controller",
+	Long:  "Add a service broker to service controller",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		return fmt.Errorf("Not implemented yet")
+	},
+}
+
+var describeCmd = &cobra.Command{
+	Use:   "describe",
+	Short: "Describe the specified service broker",
+	Long:  "Describe the specified service broker",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if len(args) == 0 {
+			return fmt.Errorf("need name of the broker")
+		}
+		return callService(fmt.Sprintf("/v2/service_brokers/%s", args[0]), "GET", "describe broker", nil)
+	},
+}
+
+var listCmd = &cobra.Command{
+	Use:   "list",
+	Short: "List brokers the service controller knows about",
+	Long:  "List brokers the service controller knows about",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		fmt.Printf("doing list %s %d\n", controller, timeout)
+		return callService("/v2/service_brokers", "GET", "list service brokers", nil)
+	},
+}
+
+var deleteCmd = &cobra.Command{
+	Use:   "delete",
+	Short: "Delete a broker from the service controller",
+	Long:  "Delete a broker from the service controller",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		fmt.Printf("doing delete %s %d\n", controller, timeout)
+		return nil
+	},
+}
--- a/cli/cmd/http_util.go
+++ b/cli/cmd/http_util.go
@@ -1,0 +1,61 @@
+package cmd
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"time"
+
+	"github.com/ghodss/yaml"
+)
+
+func callService(path, method, action string, reader io.ReadCloser) error {
+	u := fmt.Sprintf("%s%s", controller, path)
+
+	resp, err := callHttp(u, method, action, reader)
+	if err != nil {
+		return err
+	}
+	var j interface{}
+	if err := json.Unmarshal([]byte(resp), &j); err != nil {
+		return fmt.Errorf("Failed to parse JSON response from service: %s : %v", resp, err)
+	}
+
+	y, err := yaml.Marshal(j)
+	if err != nil {
+		return fmt.Errorf("Failed to serialize JSON response from service: %s : %v", resp, err)
+	}
+
+	fmt.Println(string(y))
+	return nil
+}
+
+func callHttp(path, method, action string, reader io.ReadCloser) (string, error) {
+	request, err := http.NewRequest(method, path, reader)
+	request.Header.Add("Content-Type", "application/json")
+
+	client := http.Client{
+		Timeout: time.Duration(time.Duration(timeout) * time.Second),
+	}
+
+	response, err := client.Do(request)
+	if err != nil {
+		return "", err
+	}
+
+	defer response.Body.Close()
+	body, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		return "", err
+	}
+
+	if response.StatusCode < http.StatusOK ||
+		response.StatusCode >= http.StatusMultipleChoices {
+		message := fmt.Sprintf("status code: %d status: %s : %s", response.StatusCode, response.Status, body)
+		return "", fmt.Errorf("cannot %s: %s\n", action, message)
+	}
+
+	return string(body), nil
+}
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -1,0 +1,22 @@
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+)
+
+// These are shared between all the commands
+var (
+	controller string
+	timeout    int
+)
+
+func init() {
+	RootCmd.PersistentFlags().StringVar(&controller, "controller", "http://localhost:10000", "URL for service controller")
+	RootCmd.PersistentFlags().IntVar(&timeout, "timeout", 10, "http timeout (in seconds) for interaction with service controller")
+}
+
+var RootCmd = &cobra.Command{
+	Use:   "sc-cli",
+	Short: "CLI for Service Controller operations",
+	Long:  "Command Line Interface for the Service Controller",
+}
--- a/Makefile
+++ b/Makefile
@@ -3,12 +3,16 @@
 
 all:
 	$(MAKE) --no-print-directory -C k8s
+	$(MAKE) --no-print-directory -C cli
 
 test-unit:
 	$(MAKE) --no-print-directory -C k8s test-unit
+	$(MAKE) --no-print-directory -C cli test-unit
 
 test:
 	$(MAKE) --no-print-directory -C k8s test
+	$(MAKE) --no-print-directory -C cli test
 
 clean:
 	$(MAKE) --no-print-directory -C k8s clean
+	$(MAKE) --no-print-directory -C cli clean
--- a/cli/Makefile
+++ b/cli/Makefile
@@ -1,0 +1,101 @@
+.PHONY: all cli test-unit test clean validate
+
+EXE=sc
+SRC_FILES=$(shell find . -name \*.go -print | grep -v _test.go | grep -v \\./tests)
+TEST_UNIT_DIRS=$(shell find . -name \*_test.go -exec dirname {} \; | grep -v \\./tests | sort -u)
+
+all: validate cli test-unit test
+
+#####################
+
+cli: $(EXE)
+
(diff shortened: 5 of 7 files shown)


sc-cli.go
```diff
@@ -0,0 +1,9 @@
+package main
```
duglin: can you move this into the "cmd" dir since its local to that stuff?
You should probably also copy the Makefile from the k8s/service_controller to build an exe

duglin:
Nice first step.
Let's add some testcases in k8s/service_controller that use this cli to do some stuff - along side the existing HTTP based ones.

sc-cli.go
vaikas-google: I can, I was just following their recommended set up.
https://github.com/spf13/cobra#getting-started

Perhaps, we could have
cli/
 sc-cli.go
 Makefile
 cmd/
    commands then go here.

duglin: Ah, I was thinking that there might be more than one client (go vs ruby as an example) - so having the go version in the root felt odd to me.  But perhaps I was assuming too much.

vaikas-google: I can totesmcgoats move it to something like cli/ as described above, what do you think about that? In either case, I don't think I can have the commands (which would be part of the cmd package) and the binary (which should be part of main iirc) in the same directory.

duglin: that's such an odd "feature" of go that I can't have a global variable that can be used across a package if the package is "main".  IMO there's no reason why all of the files for this cmd line couldn't be in one dir - its a pretty small package/exe.  oh well.

duglin: actually, I take that back - I was doing it wrong.
You can move sc-cli.go int "cmd" dir and it works just fine.
Just one code change to get it to compile.

vaikas-google: When you say there was only one code change, which one are you talking about? Changing the package from main -> cmd? 
Can you cut & paste your sc-cli.go that creates a binary from the cmd directory? I'm having a hard time understanding / replicating a working binary from there :)

Just to get it to compile, I had to:
 - change package main->cmd
 - remove the import statement
 - cmd.RootCmd => RootCmd

duglin: Sorry - yes it was more than one line - I forgot about the first two bullets you listed.  Only the last one was "real" to me so I blocked out the other 2

duglin: actually, I changed package cmd->main in the other files  - kept it as main in sc-cli.go

vaikas-google: Ok, so we could do that :) Question is do we want to. As per cobra user documentation, they suggest a layout that the main binary is NOT in the same directory as the commands, and they suggest:
 sc-cli.go
 /cmd
    commands go here

So, I really don't like lumping everything into the cmd directory and having them in the main package without a good reason. Anybody else who comes into the project with cobra might be surprised by this, I also think there are tools that assume this structure. 
So, I see two viable options:
  - leave the sc-cli.go at the top, you don't like that I know :)
 - introduce something like cli/ directory and have sc-cli.go there, with cmd/<commands> underneath that.

Are you fine with the second option? Seems like it would allow us to conform to suggested cobra structure, and move the sc-cli.go out of the root which I understood was your biggest concern.

Thoughts?

duglin: if 2 makes you happy go for it :-)
Just a commentary.... 
1) I don't like tools telling me how to structure my code - if I want one HUGE dir, let me
2) I don't expect this cli to be anything more than a test tool so I can't get too worked up over sticking to a formal (long-term-pretty) structure, and on the flip side, its the reason I wanted to keep it simple and in one dir.

vaikas-google: fair points :) Ok, I'll proceed with #2 then.

vaikas-google:
Yep, I'll try to get something out in the next hour or two.

vaikas-google:
Ok, I'm trying to squash my commits. And I've now gone from 3 commits to 5 commits. WTF? Figuring out what's going on now...

MHBauer:
@vaikas-google if you're on the cf slack I can help with git. 

It looks like you're committing directly to the master branch of your local repo, so you might have to rebase against a specific remote. Something like `git rebase -i origin/master` if this repo is your origin remote.

vaikas-google:
Thanks, I think I got it, take a look. As far as signing commits, I don't know what / how to do that. As far as Makefile goes, the one in service_controller involves things like building docker containers, etc. we won't be doing that here. How about I get this in, then at least you can start using parts of it. I'll then add:
 - more commands
 - tests using this instead of post?

duglin:
To sign just add `-s` to your commit, so in this case just do: `git commit -s --amend --no-edit` and then repush.

As for the Makefile, the one you were looking at uses Docker only when golang isn't installed locally, so you probably shouldn't ever need that.  I can fix-up the Makefile for you once we merge if you want, but the pattern is pretty simple.

cli/cmd/brokers.go
```diff
@@ -0,0 +1,61 @@
+package cmd
+
+import (
+	"fmt"
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	RootCmd.AddCommand(brokersCmd)
```
duglin: Do you see us having more than one top-level cmd (I don't consider "version" a real cmd)? If not do we need to require the "brokers" word on the cmd line?

cli/cmd/brokers.go
```diff
@@ -0,0 +1,61 @@
...
+	},
+}
+
+var describeCmd = &cobra.Command{
+	Use:   "describe",
+	Short: "Describe the specified service broker",
+	Long:  "Describe the specified service broker",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if len(args) == 0 {
+			return fmt.Errorf("need name of the broker")
```
duglin: s/name/ID/
Format is: `GET /v2/service_brokers/6d019023-17e6-488e-a50a-4ce1f35a9524`
https://apidocs.cloudfoundry.org/237/service_brokers/retrieve_a_particular_service_broker.html

cli/cmd/http_util.go
```diff
@@ -0,0 +1,61 @@
+package cmd
+
+import (
+	"fmt"
+	"github.com/ghodss/yaml"
+
+	"encoding/json"
```
duglin: Can you re-order these so that the glang ones are in the top section and then "others" are after the blank line?
eg.  https://github.com/cncf/servicebroker/blob/master/k8s/service_controller/service_controller.go

cli/cmd/http_util.go
```diff
@@ -0,0 +1,61 @@
...
+func callService(path, method, action string, reader io.ReadCloser) error {
+	u := fmt.Sprintf("%s%s", controller, path)
+
+	resp, err := callHttp(u, method, action, reader)
+	if err != nil {
+		return err
+	}
+	var j interface{}
+	if err := json.Unmarshal([]byte(resp), &j); err != nil {
+		return fmt.Errorf("Failed to parse JSON response from service: %s", resp)
```
duglin: Can you include the err in the output?

cli/cmd/http_util.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	y, err := yaml.Marshal(j)
+	if err != nil {
+		return fmt.Errorf("Failed to serialize JSON response from service: %s", resp)
```
duglin: please include the error in the output here too.

cli/cmd/root.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/spf13/cobra"
+)
+
+// These are shared between all the commands
+var (
+	controller string
+	timeout    int
+)
+
+func init() {
```
duglin: Probably more of a golang question, but is there an advantage to doing this in an init() rather than just in a normal func?

duglin: oh never mind.

duglin:
For the Makefile, edit the top one like this:
```
# git diff
diff --git a/Makefile b/Makefile
index a3db122..443c64c 100644
--- a/Makefile
+++ b/Makefile
@@ -3,12 +3,16 @@
 
 all:
        $(MAKE) --no-print-directory -C k8s
+       $(MAKE) --no-print-directory -C cli
 
 test-unit:
        $(MAKE) --no-print-directory -C k8s test-unit
+       $(MAKE) --no-print-directory -C cli test-unit
 
 test:
        $(MAKE) --no-print-directory -C k8s test
+       $(MAKE) --no-print-directory -C cli test
 
 clean:
        $(MAKE) --no-print-directory -C k8s clean
+       $(MAKE) --no-print-directory -C cli clean
```

and then add cli/Makefile:
```
.PHONY: all cli test-unit test clean validate

EXE=sc
SRC_FILES=$(shell find . -name \*.go -print | grep -v _test.go | grep -v \\./tests)
TEST_UNIT_DIRS=$(shell find . -name \*_test.go -exec dirname {} \; | grep -v \\./tests | sort -u)

all: validate cli test-unit test

#####################

cli: $(EXE)

$(EXE): $(SRC_FILES)
ifeq (,$(shell which go))
	@$(MAKE) --no-print-directory cli-docker
else
	@$(MAKE) --no-print-directory cli-local
endif

cli-docker:
	docker run -ti \
	  -w /src/github.com/cncf/servicebroker/cli \
	  -v ${CURDIR}:/src/github.com/cncf/servicebroker/cli \
	  -e GOPATH=/ \
	  golang \
	  sh -c "go get -d ./... && make cli-local"

cli-local:
	go build -o $(EXE) .

#####################

test-unit: cli
	@echo "\n### Running the unit tests"
ifeq (,$(shell which go))
	@$(MAKE) --no-print-directory test-unit-docker
else
	@$(MAKE) --no-print-directory test-unit-local
endif

test-unit-docker:
	docker run -ti \
	  -w /src/github.com/cncf/servicebroker/cli \
	  -v ${CURDIR}:/src/github.com/cncf/servicebroker/cli \
	  -e GOPATH=/ \
	  golang \
	  sh -c "go get -d ./... && make --no-print-directory test-unit-local"

test-unit-local:
	@go test $(TEST_UNIT_DIRS)

#####################

test: cli
	@echo "\n### Running the integration tests"
ifeq (,$(shell which go))
	@$(MAKE) --no-print-directory test-docker
else
	@$(MAKE) --no-print-directory test-local
endif

test-docker:
	docker run -ti \
	  -w /src/github.com/cncf/servicebroker/cli \
	  -v ${CURDIR}:/src/github.com/cncf/servicebroker/cli \
	  -e GOPATH=/ \
	  --net host \
	  golang \
	  sh -c "go get -d ./... && make --no-print-directory test-local"

test-local: cli
	echo No tests yet
	# @go test -v ./tests

#####################

clean:
	-rm -f $(EXE)

#####################

validate: 
ifeq (,$(shell which go))
	@$(MAKE) --no-print-directory validate-docker
else
	@$(MAKE) --no-print-directory validate-local
endif

validate-docker:
	docker run -ti \
	  -w /src/github.com/cncf/servicebroker/cli \
	  -v ${CURDIR}:/src/github.com/cncf/servicebroker/cli \
	  -e GOPATH=/ \
	  golang \
	  sh -c "go get -d ./... && make --no-print-directory validate-local"

validate-local:
	@if ! [ "`find . -name \*.go -exec gofmt -s -l {} \;`" = "" ]; then \
		echo Please gofmt the following files: \\n$(shell find . -name \*.go -exec gofmt -s -l {} \;) \\n ; \
		exit 1 ; \
	fi
```

cli/cmd/brokers.go
vaikas-google: Yes, my understanding was that you should be able to do the following:
 - manage brokers [brokers.go]
 - manage service instances [services.go]
 - manage bindings [bindings.go]

Did I misunderstand?

cli/cmd/brokers.go
MHBauer: can't retrieve buy ID in k8s. not sure what to do about the disconnect.

vaikas-google: That's interesting in a sense that when you create a service broker, it takes a name. I would expect the CLI to use the same name, but apparently the user has to type the GUID. Their CLI doesn't seem to have the describe anyways:
https://docs.cloudfoundry.org/cf-cli/cf-help.html#service-admin

cli/cmd/http_util.go
vaikas-google: done

cli/cmd/http_util.go
vaikas-google: done

cli/cmd/http_util.go
vaikas-google: done.

cli/cmd/root.go
vaikas-google: ack

vaikas-google:
Thanks, PTAL.

vaikas-google:
Anything else that needs to be addressed before submitting?

duglin:
gimme a bit - stuck in calls....

cli/cmd/brokers.go
duglin: Nope, my bad - silly question

cli/cmd/brokers.go
duglin: I guess if we could make it user-friendly, and not just a simpler way to call the REST API, then we should, so I think the create would be like cf's:
```
sc broker create SB_NAME USER PWD URL
```
so I take it back, it should be name instead of ID.  If this doesn't workout we can change it later.

duglin:
LGTM

PR #50, opened 2016-08-09
Add sample service instance create to go broker

duglin:
Other misc changes


duglin:
LGTM

PR #52, opened 2016-08-10
Make it so json parsing errors tell you where the issue is

duglin:
Make service controller be more explicit when it starts


duglin:
LGTM

PR #54, opened 2016-08-12
raw http to unauth server

MHBauer:
no more shelling out to kubectl

### DIFF
--- a/k8s/service_controller/server/k8s/storage.go
+++ b/k8s/service_controller/server/k8s/storage.go
@@ -4,12 +4,10 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
-	"os/exec"
-	"strings"
+	"net/http"
 
 	"github.com/cncf/servicebroker/k8s/service_controller/model"
 	"github.com/cncf/servicebroker/k8s/service_controller/server"
-	"github.com/cncf/servicebroker/k8s/service_controller/utils"
 )
 
 type K8sServiceStorage struct {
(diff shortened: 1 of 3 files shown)


MHBauer:
@duglin why is travis blowing up?

duglin:
I have no idea.  Its building w/o docker for some reason - its had golang installed but we didn't ask for it.  But even so, I can't reproduce this locally.

duglin:
I think travis is messing up by making golang available, but if that's it then we probably need to set GOPATH.
Try adding:
```
GOAPTH=$(shell cd ../../.. && pwd)
```
to the root Makefile right before the "all:" target and see if that helps.

MHBauer:
@duglin travis documentation is awful, but it does say the environments come with go installed by default. No idea what's going on. Tried a couple things. It's green, but it doesn't look quite right to me. I don't think it built the new client?

I'll let you fix it next week.

MHBauer:
@duglin https://www.traviscistatus.com/incidents/11hp8bhkrkn7 might be why stuff started breaking. They made changes.Added go 1.6.3 by default.

.travis.yml
```diff
@@ -5,4 +5,6 @@ services:
   - docker
 
 script:
+  - go version
+  - go get -d ./...
```
duglin: you can delete these lines,they're not needed any more.

MHBauer:
@duglin okay. it worked.

MHBauer:
lgtm myself

PR #55, opened 2016-08-17
Add new commands to the CLI

vaikas-google:
for creating service_instances, getting inventory, service_plans, deleting brokers.
Add delete broker to in memory storage service + tests.

### DIFF
--- a/cli/cmd/brokers.go
+++ b/cli/cmd/brokers.go
@@ -1,16 +1,35 @@
 package cmd
 
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
+	"io/ioutil"
+
+	"github.com/cncf/servicebroker/k8s/service_controller/model"
 	"github.com/spf13/cobra"
+)
+
+const (
+	BROKERS_URL     = "/v2/service_brokers"
+	BROKERS_FMT_STR = "/v2/service_brokers/%s"
+)
+
+var (
+	user     string
+	password string
 )
 
 func init() {
 	RootCmd.AddCommand(brokersCmd)
-	brokersCmd.AddCommand(createCmd)
-	brokersCmd.AddCommand(describeCmd)
-	brokersCmd.AddCommand(listCmd)
-	brokersCmd.AddCommand(deleteCmd)
+	brokersCmd.AddCommand(createBrokersCmd)
+	brokersCmd.AddCommand(describeBrokersCmd)
+	brokersCmd.AddCommand(listBrokersCmd)
+	brokersCmd.AddCommand(deleteBrokersCmd)
+	createBrokersCmd.Flags().StringVarP(&user, "user", "u", "", "user name to use for broker auth")
+	createBrokersCmd.Flags().StringVarP(&password, "password", "p", "", "password to use for broker auth")
+	createBrokersCmd.Flags().StringVarP(&spaceGUID, "spaceGUID", "s", "default", "Space GUID to use for broker")
+
 }
 
 var brokersCmd = &cobra.Command{
@@ -19,43 +38,73 @@ var brokersCmd = &cobra.Command{
 	Long:  "Manage brokers associated with a service controller",
 }
 
-var createCmd = &cobra.Command{
-	Use:   "create",
+var createBrokersCmd = &cobra.Command{
+	Use:   "create <NAME> <BROKER_URL>",
 	Short: "Add a service broker to service controller",
 	Long:  "Add a service broker to service controller",
 	RunE: func(cmd *cobra.Command, args []string) error {
-		return fmt.Errorf("Not implemented yet")
+		if len(args) != 2 {
+			return fmt.Errorf("need NAME and BROKER_URL")
+		}
+		req := model.CreateServiceBrokerRequest{
+			Name:      args[0],
+			BrokerURL: args[1],
+		}
+		if len(user) > 0 {
+			req.AuthUsername = user
+		}
+		if len(password) > 0 {
+			req.AuthPassword = password
+		}
+		if len(spaceGUID) > 0 {
+			req.SpaceGUID = spaceGUID
+		}
+		body, err := json.Marshal(req)
+		if err != nil {
+			return err
+		}
+		return callService(BROKERS_URL, "POST", "create broker", ioutil.NopCloser(bytes.NewReader(body)))
 	},
 }
 
-var describeCmd = &cobra.Command{
-	Use:   "describe",
+var describeBrokersCmd = &cobra.Command{
+	Use:   "describe <NAME>",
 	Short: "Describe the specified service broker",
 	Long:  "Describe the specified service broker",
 	RunE: func(cmd *cobra.Command, args []string) error {
-		if len(args) == 0 {
-			return fmt.Errorf("need name of the broker")
+		if len(args) != 1 {
+			return fmt.Errorf("need NAME of the broker")
 		}
-		return callService(fmt.Sprintf("/v2/service_brokers/%s", args[0]), "GET", "describe broker", nil)
+		guid, err := fetchBrokerGUID(args[0])
+		if err != nil {
+			return err
+		}
+		return callService(fmt.Sprintf(BROKERS_FMT_STR, guid), "GET", "describe broker", nil)
 	},
 }
 
-var listCmd = &cobra.Command{
+var listBrokersCmd = &cobra.Command{
 	Use:   "list",
 	Short: "List brokers the service controller knows about",
 	Long:  "List brokers the service controller knows about",
 	RunE: func(cmd *cobra.Command, args []string) error {
-		fmt.Printf("doing list %s %d\n", controller, timeout)
-		return callService("/v2/service_brokers", "GET", "list service brokers", nil)
+		return callService(BROKERS_URL, "GET", "list service brokers", nil)
 	},
 }
 
-var deleteCmd = &cobra.Command{
-	Use:   "delete",
+var deleteBrokersCmd = &cobra.Command{
+	Use:   "delete <NAME>",
 	Short: "Delete a broker from the service controller",
 	Long:  "Delete a broker from the service controller",
 	RunE: func(cmd *cobra.Command, args []string) error {
-		fmt.Printf("doing delete %s %d\n", controller, timeout)
-		return nil
+		if len(args) == 0 {
+			return fmt.Errorf("need NAME of the broker to delete")
+		}
+
+		guid, err := fetchBrokerGUID(args[0])
+		if err != nil {
+			return err
+		}
+		return callService(fmt.Sprintf(BROKERS_FMT_STR, guid), "DELETE", "delete broker", nil)
 	},
 }
--- a/cli/cmd/inventory.go
+++ b/cli/cmd/inventory.go
@@ -1,0 +1,38 @@
+package cmd
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/spf13/cobra"
+)
+
+const (
+	INVENTORY_FMT_STR = "%-20s %-25s %-30s\n"
+)
+
+func init() {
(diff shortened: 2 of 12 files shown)


duglin:
very excited by this.  I'll review and test later tonight.
You'll need to sign the commit

vaikas-google:
ARGH :) I always forget. How do I do this after the fact?

duglin:
`git commit -s --amend --no-edit`
`git push origin brokers -f`

vaikas-google:
Great, thanks! Done. Please take a look.

cli/cmd/brokers.go
```diff
@@ -19,16 +41,39 @@ var brokersCmd = &cobra.Command{
...
 }
 
-var createCmd = &cobra.Command{
+var createBrokersCmd = &cobra.Command{
 	Use:   "create",
 	Short: "Add a service broker to service controller",
 	Long:  "Add a service broker to service controller",
 	RunE: func(cmd *cobra.Command, args []string) error {
-		return fmt.Errorf("Not implemented yet")
+		if len(name) == 0 {
```
duglin: its a pet-peeve of mine, but, to me required args shouldn't require a flag. Can we make it so that its:
`sc brokers create NAME BROKER-URL`  ?

cli/cmd/brokers.go
```diff
@@ -19,16 +41,39 @@ var brokersCmd = &cobra.Command{
...
+		body, err := json.Marshal(req)
+		if err != nil {
+			return err
+		}
+		return callService(BROKERS_URL, "POST", "create broker", ioutil.NopCloser(bytes.NewReader(body)))
 	},
 }
 
-var describeCmd = &cobra.Command{
+var describeBrokersCmd = &cobra.Command{
```
duglin: so this one would be: `sc brokers describe NAME`.
Actually, this one is kind of odd. The help doesn't doesn't show a -n option, nor does it show NAME as an arg.  When I use -n I get an error, bad flag. When I use just NAME w/o -n then I don't get an error, just the help text.

duglin: Ah, I see, it requires a uuid. The error text on line 82 threw me.

cli/cmd/brokers.go
```diff
@@ -40,22 +85,25 @@ var describeCmd = &cobra.Command{
...
-var deleteCmd = &cobra.Command{
+var deleteBrokersCmd = &cobra.Command{
 	Use:   "delete",
 	Short: "Delete a broker from the service controller",
 	Long:  "Delete a broker from the service controller",
 	RunE: func(cmd *cobra.Command, args []string) error {
-		fmt.Printf("doing delete %s %d\n", controller, timeout)
-		return nil
+		if len(args) == 0 {
+			return fmt.Errorf("need uuid of the broker to delete")
```
duglin: hmm, this is kind of interesting. The other cmds take a name but here we ask for a uuid. Perhaps we should, under the covers, convert the name to a uuid for them?

cli/cmd/service_instances.go
```diff
@@ -0,0 +1,82 @@
...
+	Short: "Manage service instances",
+	Long:  "Manage service instances",
+}
+
+var createServiceInstancesCmd = &cobra.Command{
+	Use:   "create",
+	Short: "Create a service instance",
+	Long:  "Create a service instance",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if len(name) == 0 {
```
duglin: `sc service-instances create NAME SERVICE PLAN SPACE-GUID`

k8s/service_controller/model/service_instance.go
```diff
@@ -22,9 +22,9 @@ type LastOperation struct {
 }
 
 type CreateServiceInstanceRequest struct {
```
duglin: I think we need to have two different models. This one was written with the API to the SB in mind, which isn't the same as the API to the CC.  Your edits are for the CC API.

Perhaps we should prefix these with `SB_` and prefix the CC ones with `SC_` ??

k8s/service_controller/server/controller.go
```diff
@@ -360,11 +360,9 @@ func (c *Controller) getBroker(serviceID string) (*model.ServiceBroker, error) {
 }
 
 // This is what we get sent to us
-type CreateServiceInstanceRequest struct {
-	Name            string                 `json:"name"`
-	ServicePlanGUID string                 `json:"service_plan_guid"`
-	Parameters      map[string]interface{} `json:"parameters,omitempty"`
-}
+// TODO: Get rid of these and use the ones from:
```
duglin: Yea, this goes to what I mentioned above.  if I'm remembering correctly, I think this struct is for the CC API.
https://apidocs.cloudfoundry.org/240/service_instances/creating_a_service_instance.html

cli/cmd/brokers.go
```diff
@@ -1,16 +1,38 @@
...
-	brokersCmd.AddCommand(deleteCmd)
+	brokersCmd.AddCommand(createBrokersCmd)
+	brokersCmd.AddCommand(describeBrokersCmd)
+	brokersCmd.AddCommand(listBrokersCmd)
+	brokersCmd.AddCommand(deleteBrokersCmd)
+	createBrokersCmd.Flags().StringVarP(&name, "name", "n", "", "Name to give to Service Broker")
+	createBrokersCmd.Flags().StringVarP(&brokerURL, "brokerURL", "b", "", "URL of the broker")
+	createBrokersCmd.Flags().StringVarP(&user, "user", "u", "", "user name to use for broker auth")
+	createBrokersCmd.Flags().StringVarP(&password, "password", "p", "", "password to use for broker auth")
+	createBrokersCmd.Flags().StringVarP(&spaceGUID, "spaceGUID", "s", "", "Space GUID to use for broker")
```
bmelville: Is it worth removing all space references? It seems like they're unused and unnecessary in the generic model.

cli/cmd/inventory.go
```diff
@@ -0,0 +1,38 @@
+package cmd
```
bmelville: Should this cmd be a service-plans command instead? i.e., is inventory just service-plans list?

cli/cmd/service_instances.go
```diff
@@ -0,0 +1,82 @@
...
+	parameters      string
+)
+
+func init() {
+	RootCmd.AddCommand(serviceInstancesCmd)
+	serviceInstancesCmd.AddCommand(createServiceInstancesCmd)
+	createServiceInstancesCmd.Flags().StringVarP(&name, "name", "n", "", "Name to give to Service Instance")
+	createServiceInstancesCmd.Flags().StringVarP(&service, "service", "s", "", "Name of the service to instantiate")
+	createServiceInstancesCmd.Flags().StringVarP(&plan, "plan", "p", "", "Name of the service plan to instantiate")
+	createServiceInstancesCmd.Flags().StringVarP(&spaceGUID, "space_guid", "g", "default", "Space GUID on which to instantiate the service to")
```
bmelville: same here re: spaces

cli/cmd/util.go
```diff
@@ -0,0 +1,65 @@
...
+	i, err := callHttp(u, "GET", "inventory", nil)
+	if err != nil {
+		return "", err
+	}
+	var c model.Catalog
+	err = json.Unmarshal([]byte(i), &c)
+	if err != nil {
+		return "", err
+	}
+	for _, s := range c.Services {
```
bmelville: Can you put a TODO here to move this into a library? We have a ton of these methods that get a plan, or plan ID, or service, given a name and a catalog.

k8s/service_controller/model/service_instance.go
bmelville: Doug, this is broken in a bunch of places and I think we're going to follow up with a full model cleanup, which separates sc API model from sb API model.

vaikas-google: Yeah, I think we should clean this model up a bit with things like schemas in there, etc. I'd rather do the clean up of the model in a separate PR, since it will have ramifications to the entire stack.

cli/cmd/service_instances.go
vaikas-google: fine, but no space-guid, it should be a flag like in the other cases. Also, let's figure out what this even means, it's a CFism, and seems like it should not be a required argument as per other comments.

cli/cmd/inventory.go
vaikas-google: I don't know yet. Seems like this is the place where one would do things like searches, etc.? One could argue that they also will surface on the list commands? If you feel strongly, I can move it.

cli/cmd/brokers.go
vaikas-google: done

cli/cmd/brokers.go
vaikas-google: done

cli/cmd/brokers.go
vaikas-google: done

cli/cmd/brokers.go
vaikas-google: While we're still figuring out what it is, I'll default it to "default", right now it's required by the CF, so I'd vote to leave it in.

vaikas-google:
Ok, I think I addressed all the comments, PTAL.

k8s/service_controller/model/service_instance.go
vaikas-google: Rather than use SB_ SC_ prefixes, perhaps a better model would be to use golang package for this? That seems cleaner to me. So, perhaps we'll have model. Maybe something like:
model/sc/...
model/sb/...

bmelville: Even a step further, the SB model should be defined in a separate folder, e.g.,

service_controller/model/...
service_broker/model/...

Since the broker model is really the model for the broker's API.

vaikas-google: Sure, I just put them backwards in there depending on how to organize. I think:
sc/model/...
sb/model/...
makes sense, not wedded to either one. Just that we should use go packages for these rather than baking in the name of the objects (sc_ & sb_)

cli/cmd/util.go
```diff
@@ -0,0 +1,88 @@
...
+	"io"
+
+	"github.com/cncf/servicebroker/k8s/service_controller/model"
+)
+
+// TODO(vaikas): Move these into a ../lib or ../../lib?
+
+// from: https://play.golang.org/p/4FkNSiUDMg
+// newUUID generates a random UUID according to RFC 4122
+func NewUUID() (string, error) {
```
bmelville: github.com/satori/go.uuid is a good replacement for this. You can then just do 'uuid.NewV4().String()'

k8s/service_controller/server/mem/in_mem_service_storage.go
```diff
@@ -79,8 +79,14 @@ func (s *InMemServiceStorage) AddBroker(broker *model.ServiceBroker, catalog *mo
 	return nil
 }
 
-func (s *InMemServiceStorage) DeleteBroker(string) error {
-	return fmt.Errorf("Not implemented yet")
+func (s *InMemServiceStorage) DeleteBroker(id string) error {
```
bmelville: What happens to the instances/bindings associated with this broker when it's gone? Do we need to clear anything associated with the broker?

vaikas-google: Added a TODO to delete bindings. It's bit more complex and since nothing is persisted any ways, seems like not a big deal atm.

cli/cmd/util.go
vaikas-google: Removed it, it's not used yet. Once we need uuids, that's a good one, I did see that before too, just forgot to remove this since I didn't end up needing UUIDs after all.

vaikas-google:
Thanks! PTAL.

bmelville:
LGTM, please squash before committing.

bmelville:
LGTM

PR #63, opened 2016-08-24
First cut of defining separate SC / SB models with changes reflected …

vaikas-google:
…in the controller

Signed-off-by: vaikas-google <vaikas@google.com>


k8s/service_controller/server/controller.go
```diff
@@ -8,8 +8,9 @@ import (
 	"strings"
 	"time"
 
-	"github.com/cncf/servicebroker/k8s/service_controller/model"
 	"github.com/cncf/servicebroker/k8s/service_controller/utils"
+	sb_model "github.com/cncf/servicebroker/model/service_broker"
```
bmelville: i think typically in go it's just sbmodel and scmodel

model/service_broker/service_broker.go
```diff
@@ -0,0 +1,43 @@
+package model
```
bmelville: This is a SC model I think. Service broker doesn't actually know about itself :)

bmelville:
I'm assuming that this is just separating them out without modifying the models to align with the actual SC model? If so, just the one service broker comment, otherwise looks good.

k8s/service_controller/server/controller.go
vaikas-google: done.

bmelville:
LGTM

vaikas-google:
Thanks for the review, PTAL.

duglin:
let's not merge any more PRs until we get travis running.  Will also review this later today on the plane.

vaikas-google:
Who's in charge of setting up the travis and what's the expected ETA?

duglin:
I'm working with @caniszczyk to try to get it fixed.

duglin:
@caniszczyk added travis back so we should be good now

bmelville:
LGTM

PR #64, opened 2016-08-24
allow a host specifier on the cmd line

MHBauer:
move the k8s specific stuff into k8s location

assumes k8s at the local proxy 8001.

### DIFF
--- a/k8s/service_controller/server/k8s/storage.go
+++ b/k8s/service_controller/server/k8s/storage.go
@@ -11,20 +11,72 @@ import (
 )
 
 type K8sServiceStorage struct {
+	// Host is the location where we'll talk to k8s
+	Host string
 }
 
 const serviceDomain string = "cncf.org"
 const apiVersion string = "v1alpha1"
 const brokerResource string = "servicebrokers"
-const defaultUri string = "http://localhost:8080/apis/" + serviceDomain + "/" + apiVersion + "/namespaces/default/" + brokerResource
+const defaultUri string = "http://%v/apis/" + serviceDomain + "/" + apiVersion + "/namespaces/default/" + brokerResource
 
 // The k8s implementation should leverage Third Party Resources
 // https://github.com/kubernetes/kubernetes/blob/master/docs/design/extending-api.md
 
 var _ server.ServiceStorage = (*K8sServiceStorage)(nil)
 
-func CreateServiceStorage() server.ServiceStorage {
-	return &K8sServiceStorage{}
+type Meta struct {
+	Name string `json:"name"`
+}
+
+type k8sServiceBroker struct {
+	*model.ServiceBroker
+	ApiVersion string `json:"apiVersion"`
+	Kind       string `json:"kind"`
+	Metadata   Meta   `json:"metadata"`
+}
+
+type VName struct {
+	Name string `json:"name"`
+}
+
+type TPR struct {
+	Meta       `json:"metadata"`
+	ApiVersion string  `json:"apiVersion"`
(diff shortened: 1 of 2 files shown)


MHBauer:
Can someone test this?

duglin:
change the imports from cncf to servicebroker and it should compile

MHBauer:
@duglin @vaikas-google @bmelville Any ideas? Merge it anyway?

vaikas-google:
# _/home/travis/build/servicebroker/servicebroker/k8s/service_controller
./service_controller.go:34: too many arguments in call to k8s.CreateServiceStorage

This is what Travis is complaining about? 

Or is there some other issue that you're running into?

MHBauer:
Yes. I might be having a case of tunnel vision, but that code should work.

duglin:
ok we gotta fix this travis stuff, I'm convinced that its using code from master during imports and not from the branch.  Let me play around....

MHBauer:
my guess, don't do 'go get' at the top level.

MHBauer:
Why does that work dug? Doesn't make any sense.

duglin:
There actually may have been an easier way to do it for the official SB builds but it would never work for forked builds.  In the case of forked builds the src was cloned into a path with our username in it instead of "servicebroker" which is why the "go get ./..." would always grab the master files and never see our local/changed files.  The `ln` forces our current clone to the right path with servicebroker as the dir name.

duglin:
Ideally travis should checkout master and then merge in our branch so the paths are ok.

k8s/service_controller/server/k8s/storage.go
```diff
@@ -11,20 +11,72 @@ import (
...
+	kind       string  `json:"kind"`
+	Versions   []VName `json:"versions"`
+}
+
+const TPRapiVersion string = "extensions/v1beta1"
+const thirdPartyResourceString string = "ThirdPartyResource"
+
+var versionMap []VName = []VName{{"v1alpha1"}}
+var serviceBrokerMeta Meta = Meta{"service-broker.cncf.org"}
+var serviceBindingMeta Meta = Meta{"service-binding.cncf.org"}
```
duglin: Missing a "services" entry but I'm assuming you'll get that in a follow-on PR.

k8s/service_controller/service_controller.go
```diff
@@ -15,65 +12,26 @@ import (
...
 	Backend    string
+	Host       string
 }
 
 var options Options
 
 func init() {
 	flag.StringVar(&options.ConfigPath, "c", ".", "use '-c' option to specify the config file path")
 	flag.StringVar(&options.Backend, "backend", "mem", "backend to use for storing info. 'mem' for in memory backend (default), 'k8s' for kubernetes based backend")
+	flag.StringVar(&options.Host, "host", "localhost:8001", "the host of the backend to connect to")
```
duglin: Can you change this back to "8080" to match how it used to be and how Kube defaults?

k8s/service_controller/server/k8s/storage.go
MHBauer: har har dug.

k8s/service_controller/service_controller.go
MHBauer: is it defined anywhere? `kubectl proxy` usually brings it up on 8001 for me.

duglin:
that's odd its not 8080 for you, I don't think I set anything up to make it do that.

duglin:
LGTM

PR #65, opened 2016-08-24
switch to guid to avoid mismatch between k8s and cf

MHBauer:


k8s/service_controller/server/k8s/storage.go
```diff
@@ -120,7 +120,7 @@ func (kss *K8sServiceStorage) AddBroker(broker *model.ServiceBroker, catalog *mo
 	// versions:
 	// - name: v1alpha1
 	ksb := NewK8sSB()
-	ksb.Metadata = Meta{Name: broker.Name}
+	ksb.Metadata = Meta{Name: broker.GUID}
```
duglin: I think this is right but just want to make sure, where is this info used? If its used internally and its required to be a constant even in the face of a SB rename then I agree.

MHBauer: https://github.com/kubernetes/kubernetes/blob/master/docs/devel/api-conventions.md#metadata

duglin: any reason you're not using Name for Name and using GUID on the uid field?

MHBauer: uid cannot be set. k8s internally generated.

MHBauer: name is the 'unique identifier in a namespace'

duglin:
LGTM

PR #67, opened 2016-08-26
Bindings command line 

vaikas-google:


vaikas-google:
Ok, this has now been updated to use the new models. Please take a look.

cli/cmd/service_bindings.go
```diff
@@ -0,0 +1,120 @@
...
+		if err != nil {
+			return err
+		}
+		var si model.ServiceInstance
+		err = json.Unmarshal([]byte(i), &si)
+		if err != nil {
+			return err
+		}
+
+		fmt.Printf("%s -> %s\n\t%+v\n", sb.FromServiceInstanceName, si.Name, sb.Parameters)
```
bmelville: Might be good to illustrate here whether the from service instance has been instantiated yet or not. (can be a follow up)

bmelville:
Please sign and squash

cli/cmd/service_bindings.go
vaikas-google: As in something like
 foo (not instantiated yet) -> mysql

bmelville: Yeah something like that, or even just a * next to it or (foo) instead of foo

vaikas-google:
Ok, I did squash and sign:
git commit -s --amend --no-edit

bmelville:
LGTM

PR #71, opened 2016-09-01
create get all services endpoint and stub out

MHBauer:
Needed this for my understanding so I stubbed it out. Will use soon.

### DIFF
--- a/k8s/service_controller/server/nil/storage.go
+++ b/k8s/service_controller/server/nil/storage.go
@@ -119,6 +119,10 @@ func (s *NotImplementedYetServiceStorage) ListServices() ([]*model.ServiceInstan
 	return nil, fmt.Errorf("Not implemented yet")
 }
 
+func (s *NotImplementedYetServiceStorage) GetServices() ([]*model.Service, error) {
+	return nil, fmt.Errorf("Not implemented yet")
+}
+
 func (s *NotImplementedYetServiceStorage) GetService(id string) (*model.ServiceInstance, error) {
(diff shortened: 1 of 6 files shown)


k8s/service_controller/server/nil/storage.go
```diff
@@ -79,6 +79,10 @@ func (s *NilServiceStorage) GetBrokerByService(id string) (*model.ServiceBroker,
 	return nil, nil
 }
 
+func (s *NilServiceStorage) AllServices() ([]*model.Service, error) {
```
duglin: Why do we have this if we never call it?

MHBauer: at some point in the past I was thinking of using it for testing or as a template for other storage implementations. Might be more trouble than it's worth.

duglin: can we add it later when needed? Just feels odd if its not used.
We still need the rest of the PR though, right?

MHBauer: won't compile without a full implementation.

duglin: that's why I'm a bit confused. It compiles for me w/o this method. I didn't see any ref to AllServices() in any interface.

MHBauer: I see what you mean. I'll take a closer look.

MHBauer:
@duglin fixed

duglin:
LGTM

PR #72, opened 2016-09-03
add initial api spec

duglin:
- took the files from: https://docs.cloudfoundry.org/services/
- converted from html.erb to just .md files
- still lots of clean-up to be done but its a starting point


duglin:
We should probably also move this to a new top-level repo called "specs" but I don't have access to do that, so I'll move this later.

duglin:
LGTM

PR #73, opened 2016-09-06
missed an instance of cncf import

MHBauer:
not sure how this was missed. might have been causing our build crazyness.


duglin:
LGTM

Issue #59, opened 2016-08-22
Service broker can declare supported schema for parameters field


bmelville:
Service broker will provide schema information for both service instances and bindings for each plan. The schema information is for both parameter inputs and output properties. Currently the proposal is that the schema format will be JSON schema v4.

Proposal: https://docs.google.com/document/d/1-IKI-PwPnhtbK0su1UzWA_UoyaCjL8S2d6ky3Xmuwvg/edit?usp=sharing

shalako:
This issue seems more appropriate for the Kubernetes service-catalog project. The openservicebrokerapi org is not meant for development of platform controllers.

bmelville:
@shalako the intention of this issue is for the parameter schema PR. I've renamed to make it more appropriate.

bmelville:
Proposal doc: https://docs.google.com/document/d/1-IKI-PwPnhtbK0su1UzWA_UoyaCjL8S2d6ky3Xmuwvg/edit?usp=sharing

shalako:
@avade I know you're planning to draft a proposal for the CF community for this feature. You may want to consider Brendan's existing google doc; link to it and dedup or duplicate the content in yours? I expect going forward we'll have to have a design proposal for the API and another for CF changes.

avade:
I had already lifted much of the content form Brendan's doc :)

Brendan are you going to update the proposed API spec part of this doc or
sharing more as an example?

If this is going to be "the" proosal doc for this feature, can I get rights
to suggest edits, please?

bmelville:
Hey Alex,

The API spec part was more an example section, but I'm happy to make this
more the canonical doc if you'd like. I believe that you can suggest edits
if you have comment rights by switching to suggest mode. If not I can give
you edit rights.

B.

avade:
I have had a read through the comments and found it difficult to work out the state of the discussion.

Perhaps @bmelville / @gberche-orange could provide a summary and what needs to happen before we can `accept` the proposal and get started on a Cloud Foundry implementation.

gberche-orange:
To my understanding, the core specification needs a decision on where schemas should be attached at :
- object level (service instance, service binding) to the expense of restricting some fields only to some verbs: create/update/delete/actions)
- method level (provision, bind, update, delete, actions) to the expense of duplicating field definitions for the ones common between verbs, such as create and update

[...]

PR #74, opened 2016-09-07
Service broker can declare supported schema for parameters field

bmelville:
Signed-off-by: Brendan Melville bmelville@google.com

Refs #59

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -791,7 +791,7 @@ existing system.
 <tr>
   <td>parameters</td>
   <td>JSON object</td>
-  <td>Cloud Foundry API clients can provide a JSON object of configuration parameters with their request and this value will be passed through to the service broker. Brokers are responsible for validation.</td>
+  <td>Cloud Foundry API clients can provide a JSON object of configuration parameters with their request and this value will be passed through to the service broker. Brokers are responsible for validation. Valid parameters for a request may be specified by the broker catalog for each plan in [schemas.service_instances](#SchemaObject).</td>
 </tr>
 <tr>
   <td>accepts_incomplete</td>
@@ -942,7 +942,7 @@ To enable this functionality, a broker declares support for each service by incl
 <tr>
   <td>parameters</td>
   <td>JSON object</td>
-  <td>Cloud Foundry API clients can provide a JSON object of configuration parameters with their request and this value will be passed through to the service broker. Brokers are responsible for validation.</td>
+  <td>Cloud Foundry API clients can provide a JSON object of configuration parameters with their request and this value will be passed through to the service broker. Brokers are responsible for validation. Valid parameters for a request may be specified by the broker catalog for each plan in [schemas.service_instances](#SchemaObject).</td>
 </tr>
 <tr>
   <td>previous_values</td>
@@ -1145,7 +1145,7 @@ the resource it creates.</p>
 <tr>
   <td>parameters</td>
   <td>JSON object</td>
-  <td>Cloud Foundry API clients can provide a JSON object of configuration parameters with their request and this value will be passed through to the service broker. Brokers are responsible for validation.</td>
+  <td>Cloud Foundry API clients can provide a JSON object of configuration parameters with their request and this value will be passed through to the service broker. Brokers are responsible for validation. Valid parameters for a request may be specified by the broker catalog for each plan in [schemas.service_bindings](#SchemaObject).</td>
 </tr>
 </tbody>
 </table>
--- a/spec.md
+++ b/spec.md
@@ -367,8 +367,57 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
   <td>boolean</td>
   <td>This field allows the plan to be limited by the non_basic_services_allowed field in a Cloud Foundry Quota, see <a href="http://docs.cloudfoundry.org/running/managing-cf/quota-plans.md">Quota Plans</a>. Default: true</td>
 </tr>
+<tr>
+  <td><a href="#SchemasObject">&nbsp;&nbsp;&nbsp;schemas</a></td>
+  <td>object</td>
+  <td>The schema definitions for the input properties for service instances and bindings of the plan.</td>
+</tr>
 </tbody>
 </table>
+
+<h5> Schemas Object <a name="SchemasObject"></a> </h5>
+
+<table border="1" class="nice">
+<thead>
+<tr>
+  <th>Response field</th>
+  <th>Type</th>
+  <th>Description</th>
+</tr>
+</thead>
+<tbody>
+<tr>
+  <td><a href="#SchemaObject">&nbsp;&nbsp;&nbsp;service_instance</a></td>
+  <td>object</td>
+  <td>The schema definitions for the input properties for a service instance.</td>
+</tr>
+<tr>
+  <td><a href="#SchemaObject">&nbsp;&nbsp;&nbsp;service_binding</a></td>
+  <td>object</td>
+  <td>The schema definitions for the input properties for a service binding. Used only if the service is bindable.</td>
+</tr>
+</tbody>
+</table>
+
+<h5> Schema Object <a name="SchemaObject"></a> </h5>
+
+<table border="1" class="nice">
+<thead>
+<tr>
+  <th>Response field</th>
+  <th>Type</th>
+  <th>Description</th>
+</tr>
+</thead>
+<tbody>
+<tr>
+  <td>&nbsp;&nbsp;&nbsp;parameters</td>
+  <td>object</td>
+  <td>The schema definitions for the input parameters. Schema definitions must be valid <a href="http://json-schema.org/">JSON Schema draft v4</a></td>
+</tr>
+</tbody>
+</table>
+
 
 \* Fields with an asterisk are required.
 
@@ -413,6 +462,29 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
         }, {
           "content": "40 concurrent connections"
         }]
+      },
+      "schemas": {
+        "service_instance": {
+          "parameters": {
+            "required": [
+              "fake_required_property"
+            ],
+            "properties": {
+              "fake_required_property": {
+                "type": "string"
+              }
+            }
+          }
+        },
+        "service_binding": {
+          "parameters": {
+            "properties": {
+              "fake_binding_property": {
+                "type": "string"
+              }
+            }
+          }
+        }
       }
     }, {
       "name": "fake-async-plan",
@@ -424,6 +496,29 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
         "bullets": [{
           "content": "40 concurrent connections"
         }]
+      },
+      "schemas": {
+        "service_instance": {
+          "parameters": {
+            "required": [
+              "fake_required_property"
+            ],
+            "properties": {
+              "fake_required_property": {
+                "type": "string"
+              }
+            }
+          }
+        },
+        "service_binding": {
+          "parameters": {
+            "properties": {
+              "fake_binding_property": {
+                "type": "string"
+              }
+            }
+          }
+        }
       }
     }, {
       "name": "fake-async-only-plan",
@@ -435,6 +530,29 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
         "bullets": [{
           "content": "40 concurrent connections"
         }]
+      },
+      "schemas": {
+        "service_instance": {
+          "parameters": {
+            "required": [
+              "fake_required_property"
+            ],
+            "properties": {
+              "fake_required_property": {
+                "type": "string"
+              }
+            }
+          }
+        },
+        "service_binding": {
+          "parameters": {
+            "properties": {
+              "fake_binding_property": {
+                "type": "string"
+              }
+            }
+          }
+        }
       }
     }]
   }]


specs/sb/api.md
```diff
@@ -342,6 +400,38 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
...
+      "schemas": {
+        "service_instance": {
+          "parameters": {
+            "required": [
+              "fake_required_property"
+            ],
+            "properties": {
+              "fake_required_property": {
+                "type": "string"
+              }
```
vaikas-google: Maybe add an example of what an output schema looks like also?

bmelville: Added for both spots. PTAL at the new commit.

specs/sb/api.md
```diff
@@ -342,6 +400,45 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
...
+            }
+          },
+          "outputs": {
+            "properties": {
```
pmorie: I'm not super-clear on why this `properties` element is needed, since the `outputs` and `credentials` are described as being properties -- is there something that I'm missing?

specs/sb/api.md
gokrokvertskhov: What are differences between properties and parameters of an instance? I think it may be confusing. What if simplify this to
"schemas":{
  "service_instance":{
    "properties": {},
    "required": []
  }
}

specs/sb/api.md
bmelville: Yes, as we discussed before, schema is defined as JSON schema, which consists of a "properties" section, possibly in addition to "required" and "definitions". As such, each different schema defined in the schemas section must include at least a subfield "properties" to match JSON schema.

specs/sb/api.md
```diff
@@ -296,6 +296,64 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
...
 </tr>
+<tr>
+  <td>&nbsp;&nbsp;&nbsp;schemas</td>
+  <td>object</td>
+  <td>The schema definitions for the input and output properties for service instances and bindings of the plan.</td>
+</tr>
+</tbody>
+</table>
+
+<h5> Schemas Object <a name="PObject"></a> </h5>
```
duglin: I think this should be "Schemas Field" to align with the previous table.

specs/sb/api.md
```diff
@@ -296,6 +296,64 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
...
+</tr>
+<tr>
+  <td>&nbsp;&nbsp;&nbsp;service_binding</td>
+  <td>object</td>
+  <td>The schema definitions for the input and output properties for a service binding. Used only if the service is bindable.</td>
+</tr>
+</tbody>
+</table>
+
+<h5> Schema Object <a name="PObject"></a> </h5>
```
duglin: I think it would read better if this table were nested under the previous one and "parameters" and "output" were under "service_instance" row, and "credentials" was under the "service_binding" row. Having two tables right next to each other with almost the exact same titles is kind of confusing for a newbie.  Nesting them will help scope things.

robertgmoss:
LGTM

specs/sb/api.md
pmorie: Got you, thanks for the clarification.

pmorie:
LGTM aside from minor formatting nits which @duglin has already pointed out.

specs/sb/api.md
```diff
@@ -364,6 +461,45 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
...
+            "properties": {
+              "fake_output": {
+                "type": "string"
+              }
+            }
+          }
+        },
+        "service_binding": {
+          "credentials": {
+            "properties": {
```
gokrokvertskhov: In current form "credentials" entry looks like a required input parameter which used during a service binding. If this is an output, then it should be inside "outputs" entry just like it is done in "service_instance" example.

specs/sb/api.md
shalako: `schemas.service_instance.parameters.properties` has defined the valid inputs. What is `outputs` for? Could you please provide an example use case? How would a client use it and for what purpose?

specs/sb/api.md
bmelville: Doug, I was going for consistency with how the rest of the doc is. For example, a Plan Object is a sub-section of the Service Object, and so the Schemas Object acts in the same manner.

specs/sb/api.md
bmelville: See my comments above Doug, I chose to go with consistence within the doc, where sub objects are broken out into their own section, e.g., with the Plan Object. I have however added hyperlinks as was done with the plan field.

specs/sb/api.md
bmelville: There is an example of an output for both service instance and service binding in the example section. I have updated them to be more clear, please let me know if you still feel them lacking.

shalako: Your example value of `outputs` is `dashboard_url`, which is an optional top-level field. `schemas` is effectively an opaque object specific to a broker. Are you proposing that `dashboard_url` is no longer part of the API spec? 

IMO, `outputs` is beyond the scope of the discussed use case. We're not trying to schemify all the things. The scope is very specific: expose what fields a client could send in the `parameters` field on create, update, and bind. This `outputs` field defines a broker's response for those operations. Arguably this is not necessary, because the broker's response is already defined by the API spec.

bmelville: Your observation for service instances outputs is correct. However it is within scope to schematize outputs for service bindings, and I didn't want to define separate objects for each. I can either separate the objects into nearly identical objects, or specify in the docs that outputs schema is only valid for bindings. How do these approaches seem to you?

shalako: I would suggest PRs representing proposed changes to the broker API spec address individual use cases so they can be discussed discreetly. This PR seems to conflate several use cases. Best I can tell there are several things going on in this PR:
- UI clients should be able to dynamically display UI elements for capturing inputs to construct the `parameters` field which can be optionally sent with the `PUT /v2/service_instances/:guid`, `PATCH /v2/service_instances/:guid`, and `PUT /v2/service_instances/:guid/service_instances/:guid` endpoints.
- Some use case related to outputs for service bindings. I don't understand what the use case is, but it is IMO independent from the first one.

> However it is within scope to schematize outputs for service bindings, 
Schematizing outputs for service bindings is a separate use case from schematizing inputs for the parameters field. The former is not well defined and IMO needs further discussion. The latter is more well defined and actionable. Conflating the two is confusing. 

> and I didn't want to define separate objects for each. I can either separate the objects into nearly identical objects, or specify in the docs that outputs schema is only valid for bindings. How do these approaches seem to you?
I don't understand what you're referring to.

bmelville: I am happy to split these up and discuss further in the other PR. I have updated to remove outputs from this PR.

shalako: Thank you!

bmelville: Please see this commit from the new PR to open discussion on outputs :)

https://github.com/servicebroker/servicebroker/pull/96/commits/aba7c242838ed11d446cecc0d0c5f1b3c40bd2d6

specs/sb/api.md
```diff
@@ -342,6 +390,29 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
         }, {
           "content": "40 concurrent connections"
         }]
+      },
+      "schemas": {
+        "service_instance": {
```
shalako: "service_instances" (plural) to match the api resource

specs/sb/api.md
```diff
@@ -342,6 +390,29 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
...
+              "fake_required_property"
+            ],
+            "properties": {
+              "fake_required_property": {
+                "type": "string"
+              }
+            }
+          }
+        },
+        "service_binding": {
```
shalako: `service_bindings` (plural) to match the API resource we're providing the schema for

avade:
Sometimes a service may want to offer certain parameters at creation time only.

For example, you may allow the number of nodes in the cluster to be set at `provision` time only, but the total connections to be changed with an `update`.

How could this be represented in the schema?

bmelville:
@avade we've been thinking about this extension as well. The beauty of JSON schema is that we can add fields as extensions as long as they don't conflict with names. Typically, you add fields like x-serviceBroker-updatable: false, and it is the responsibility of the broker to provide validation of that.

specs/sb/api.md
bmelville: Good catch, thanks, I've updated for both service_instances and service_bindings.

bmelville:
@avade I've added some links to the original commit message that may help you as you're playing with this. See the following:

Documentation on JSON Schema can be found here:

- https://tools.ietf.org/html/draft-zyp-json-schema-04
- https://tools.ietf.org/html/draft-fge-json-schema-validation-00

Examples of JSON Schema can be found here:

- http://json-schema.org/examples.html
- http://json-schema.org/example1.html
- http://json-schema.org/example2.html

avade:
Awesome. Thanks, Brendan.

shalako:
Brendan,

Would you consider updating this PR so that the schema objects are scoped to operation, rather than API resources? The API resource is implicit.

```
plans
  schemas
    provision
      parameters
        properties
        required
    update
      parameters
        properties
        required
    bind
      parameters
        properties
        required
```

st3v:
Here's a first take on a simple meta-schema for arbitrary params:

```json
{
  "properties": {
	"$schema": {
      "type": "string",
      "enum": [
        "http://json-schema.org/draft-04/schema#"
      ]
    },
    "type": {
      "type": "string",
      "enum": [
        "object"
      ]
    },
    "properties": {
      "type": "object"
    },
    "required": {
      "type": "array",
      "items": { "type": "string" },
      "uniqueItems": true
    }
  },
  "additionalProperties": false,
  "required": [
    "type",
    "$schema"
  ]
}
```

This currently requires the schema provided by the broker to:

* follow JSON-schema Draft 4 and explicitly specify that using a `$schema` constraint
* specify a top-level `type` constraint that must be set to `object`
* optionally specify a `properties` constraint of type `object`
* optionally specify a `required` constraint  of type `array` containing unique strings
* not specify any additional constraints (e.g. $ref, $id, ....)

This is just a first example. We should discuss what features of JSON-Schema 4 we would like to allow and what features we don't support.

specs/sb/api.md
```diff
@@ -364,6 +435,29 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
         "bullets": [{
           "content": "40 concurrent connections"
         }]
+      },
+      "schemas": {
+        "service_instances": {
```
st3v: As mentioned during the call these examples would actually validate successfully against everything that is not of type `object`, e.g. `42` or `foo`. 

This is because we they don't specify a `type` constraint, i.e. they allow any type. The `properties` and `required` constraints are only enforced for types for which they apply, i.e. the `object` type. They don’t apply for other types. In other words, the says I’m okay with any type, but if it is an `object` it has to have a property with name `fake_required_property` and type `string`.

You can try that yourself using one of the openly available JSON-schema validators, e.g. http://www.jsonschemavalidator.net and using the following two schemas ...

```
{"properties": { "a": {} }, "required": [ "a" ]}
```
and
```
{"type": "string", "properties": { "a": {} }, "required": [ "a" ]}
```

Number `42` successfully validates against the first but not the second schema.

I agree that this behavior is kinda weird. But it works as designed. See https://github.com/json-schema/json-schema/issues/172.

bmelville:
@shalako Re: making this a property of the operation

I thought through this a lot because it was initially very appealing (similar to Swagger API spec), but I don't think it makes sense in this context. The problem is that the broker API does not provide different call paths for or differentiate in any way between creation and updating of resources. There is only PUT, so it seems very confusing and possibly in opposition to the spirit of the method that the same PUT call take different parameters depending on the state of the underlying resource, especially give that there is no GET for resources, so the caller can never know whether the PUT will be a create or update.

What's the goal of this? Is it to prevent some properties from being updatable? My thoughts are that having a single schema, paired with PATCH and a per-property extension field to set immutability would satisfy all of our needs. An example of this is:

```
service_instances:
  properties:
    foo:
      type: string
    bar:
      type: string
      x-servicebroker-updatable: false
```

I could create the resource with:

```
PUT
foo: hello
bar: world
```

and I could update with:

```
PATCH
foo: I love
```

but could not update with:

```
PATCH
bar: icecream
```

bmelville:
@shalako Re: making this a property of the operation

I thought through this a lot because it was initially very appealing (similar to Swagger API spec), but I don't think it makes sense in this context. The problem is that the broker API does not provide different call paths for or differentiate in any way between creation and updating of resources. There is only PUT, so it seems very confusing and possibly in opposition to the spirit of the method that the same PUT call take different parameters depending on the state of the underlying resource, especially give that there is no GET for resources, so the caller can never know whether the PUT will be a create or update.

What's the goal of this? Is it to prevent some properties from being updatable? My thoughts are that having a single schema, paired with PATCH and a per-property extension field to set immutability would satisfy all of our needs. An example of this is:

service_instances:
  properties:
    foo:
      type: string
    bar:
      type: string
      x-servicebroker-updatable: false
I could create the resource with:

PUT
foo: hello
bar: world
and I could update with:

PATCH
foo: I love
but could not update with:

PATCH
bar: icecream

EDIT:
I had misconstrued the API a little bit. However, I think my point is still valid. It seems that updating an instance using PATCH allows you to change the parameters, but you must still provide the full set of parameters. Is this true? If so then I believe the schema would remain the same and it is still a question of immutability of particular fields.

bmelville:
@st3v Re: meta-schema

My concern with introducing our own metaschema is that there is a lot to JSON schema worth supporting in addition to this. In looking, it appears JSON Schema provides their own meta-schemas. I think that we could use the core meta-schema, which provides all of the common fields like definitions, required, properties, patternProperties, etc., and explicitly disallow the hyper meta-schema, which adds support for URL features.

These two meta-schemas are defined under the Meta-schemas section here: http://json-schema.org/documentation.html

Let me know what you think about that.

B.

shalako:
@bmelville I wouldn't expect all possible keys within the parameters field would be required with PUT, but that would be left up to individual brokers. The parameters field itself is optional, per the API spec. But an individual broker could reject the request if some required keys of a PUT or PATCH are not provided. The required field for each key would suggest how the broker will validate the request, and not all keys should be required for PATCH (only those the broker says are required). The broker should be able to declare which fields are required for each operation. Maybe a field is required on PATCH but not PUT (weird, but why not?). Many brokers will have no required keys within the parameters field; many brokers won't support any parameters at all.

st3v:
@bmelville With regards to meta-schemas. Please excuse the long message. If you prefer, I would be happy to jump on a call to discuss this further. It might be a bit too implementation-specific for the general call on Thursday.

I totally see where you are coming from in your comment above. I also don't want to restrict all the flexibility JSON-schema gives broker authors but there are a few features and characteristics of the standard that might turn out to be problematic for our use case. I think we need to figure out how we want to deal with them.

At the risk of repeating myself, it might be worth taking a closer look at the two potential issues I mentioned before. There might be others.

### 1. Potential inconsistencies between Broker API and Broker-provided schema

As mentioned in [this comment](https://github.com/openservicebrokerapi/servicebroker/pull/74#pullrequestreview-7051548), the current examples expose an interesting problem for schemas that don't specify a top-level `type` constraint. Most people assume that for those schemas validators would use the `object` type as default. But that's not the case. Instead, such schemas allow any type. This seems to be a pretty common misunderstanding for everyone new JSON-schema. So it is rather likely that we will see some brokers do the same, i.e. they don't specify a top-level `type` in the schema they expose. 

As mentioned, such schemas would allow pretty much any value for the `parameters` property. Numbers, strings, booleans, etc. However, according to our own API spec `parameters` must be an object:

> Cloud Foundry API clients can provide a JSON object of configuration parameters with their request and this value will be passed through to the service broker. Brokers are responsible for validation. 

So if the broker provides a schema without a top-level `type`, the client might use that schema to validate a `parameters` value that is not an object, let's say it's the string `foo`. Depending on the implementation of the controller, this would then either lead to an error when the client sends the request to the controller or it would result in a request from the controller to the broker that looks something like this:

```
{
  "organization_guid": "org-guid-here",
  "plan_id":           "plan-guid-here",
  "service_id":        "service-guid-here",
  "space_guid":        "space-guid-here",
  "parameters":        "foo"
}
```  

At that point our controller would no longer adhere to our own API spec.

We would have a very similar issue if we allowed the broker to specify a schema that explicitly defines a `type` constraint with a value other than`object`. E.g.:

```
{ "type": "string" }
```

Without a meta-schema we wouldn't really have a way to detect this problem at broker registration-time. Instead we would have to deal with the issue of an invalid `parameters` value at provisioning-, update-, or bind-time.

### 2. URI resolution scopes and dereferencing

You mentioned that we could just *"... disallow the hyper meta-schema ..."*. But all the Hyper Meta-schema does is defining the `links`, `fragmentResolution`, `media`, `readOnly`, and `pathStart` keywords ([see here](http://json-schema.org/hyper-schema)). Sorry if I wasn't clear enough, but this has nothing to do with JSON-references and JSON-pointers I was referring to during last weeks call. 

Take a look at sections [7](http://json-schema.org/latest/json-schema-core.html#rfc.section.7) and [8](http://json-schema.org/latest/json-schema-core.html#rfc.section.8) of the standard. They explain the `$ref` and `id` keywords that can be used to scope and reference other schemas through a URI. The URI can point to a schema that is defined within the same document (e.g. `"$ref" = "#/definitions/foo"`) or to a schema that lives outside the document (e.g. `"$ref": "foo.json#/definitions/foo"` or `"$ref": "http://example.com/foo.json#/definitions/foo"`). 

It's up to the implementation of the standard (in our case a schema validator on the controller-side or a form generator on the client-side) to resolve those URIs when parsing the schema. I don't think we're going to have problems with inline-references (i.e. references within the same document). But what about external references? Let's assume our controller wants to validate the Broker-provided schema at registration time. Can we assume that the controller always has access to those external schemas? What about a client that uses those schemas to render forms?

It's not uncommon for these components to run on-premise without access to the internet. Are we okay seeing some brokers that might not work in an offline environment because they specify external references in their schemas? If we are okay with that, how do we handle those situations and how do we communicate this to broker authors? Wouldn't it just be easier to disallow external references and verify that during registration?

### Approaches

Having a meta-schema is one way to address these issues. The meta-schema doesn't have to be as restrictive as in my previous example. It could just enforce a top-level `type` constraint with value `object` (addresses issue 1) and restrict references to be inline (addresses issue 2). In general, I think it might be better to be somewhat restrictive to begin with and add more JSON-schema features as we come across use-cases that need them. It's usually easier to add stuff than to take it away, especially if you have brokers implementing it. But that's up for discussion.

Documenting these potential pitfalls as part of the spec and having ways to deal with the resulting issues in the controller when they occur might be another approach that doesn't involve having a meta-schema. Happy to discuss that approach.

bmelville:
@st3v I very much agree with all of your points. Is it safe to say that meta-schema is just an implementation detail of what we're trying to accomplish?

I would word our version of the spec as:

* Conforms to draft v4 core JSON Schema spec
* Disallows external references

I think that "type: object" and "$schema" are encapsulated in the first statement. We can choose to validate the requirements using a meta-schema.

bmelville:
@shalako Thanks for the clarification Shannon. In my mind, I have been thinking of service instances and service bindings as RESTful resources with properties of their own, which are defined by the input parameters. Based on what you are saying, this is not the case, and I might manipulate an instance like the following:

```
PUT
a: 1
b: 2
c: 3

PATCH
d: 4
e: 5
f: 6
```

This would have two schemas with completely different properties, one for creation, one for update. Is that the spirit of the API? It seems a little weird to me that in the broker API these resources aren't treated as objects with properties.

When I think of PUT vs. PATCH, I think that a PATCH operation allows me to modify a subset of the properties available to modify in PUT, which will be patched into the full set of existing values, but the functionality is very different here. I am a little concerned as people will be using the broker API to build declarative systems to configure and manage instances and bindings (e.g., in the service catalog SIG), and this model of allowing different properties during update than insert contradicts that model and will make it more challenging and confusing to use.

shalako:
@bmelville 

Using your example PUT and PATCH, after the PUT the service instance might have configuration
```
a: 1
b: 2
c: 3
d: default
e: default
f: default
```
and after the PATCH
```
a: 1
b: 2
c: 3
d: 4
e: 5
f: 6
```


Consider the following stories:

- As a service provider, I would like to offer app developers a service plan that does not support configuration options for create or update
- As a service provider, I would like to offer app developers a service plan that offers a few non-required configuration options for create only, as I don't yet support changing these properties once the instance is provisioned. The developer should be able to choose values for none, one, or many of the available options.
- As a service provider, I would like to offer app developers a service plan that allows developers to update some, but not all of the configuration options the plan supports for create. There are some properties I still don't support changing once the service instance is created. The developer should be able to change the value of one property on update for which they accepted the default on create.
- As a service provider, I would like to require a few configuration options on create, in addition to optional parameters. Once the service instance has been created, a subset of those required options may optionally be updated, but a few of them cannot be changed.
- As a marketplace catalog provider, I want to enable my API clients to discover what parameters a broker supports for create, update, and bind so that they can appropriately prompt users for values of this parameters and construct the complete value of the parameters field sent with create, update, and bind requests
- As a marketplace catalog provider, I do not want to be responsible for validation of the parameters field that my clients may send with create, update, and bind requests. I expect my API clients to have used the schema the broker supplied and to have done their own client-side validation. 
- As a marketplace catalog provider, I want to enable my APIs client to discover the value of the parameters field last successfully sent to the broker for the service instance or service binding, so that my API clients can display to app developers the current configuration of these resources.

st3v:
@bmelville 

> I would word our version of the spec as:
>
>* Conforms to draft v4 core JSON Schema spec
>* Disallows external references

Agreed, adding this to the spec makes sense. I still believe it would help developers implementing the spec to have a meta-schema but we can discuss that later and handle it in a separate PR. 

I would add one more point ...

* Explicitly defines the top-level type to be an object, i.e. `"type": "object"`

I don't think that this is already covered by your first statement.

avade:
@bmelville @st3v @shalako - Is the plan to wait until we have extracted the docs from CF, as discussed in last week's call, before updating the proposal? It may be an overhead, but I would be interested to see it updated before that happens, then rebased.

It would be great if we can be in a place to decide if we are happy with the proposal to go into the API (2.12?) in tomorrow's call.

angarg12:
@avade we agreed to continue working on the PR until the merge.
@bmelville @st3v should we accept this PR and open new ones to discuss technical/implementation/wording details? This one is dragging and is becoming too long and hard to follow. Except for minor details (that we can discuss separately), it LGTM.

shalako:
I don't believe that a thread being long and hard to follow is a good reason to merge a PR. That would be like signing a bill into law without having read it, because it is long (sadly, I suspect it happens all the time). 

I would prefer not to merge any PR that contains known issues or controversial content, with the expectation of "fixing it later". Instead I would would recommend one of the following:
- Improve the PR to the point where it is acceptable
- Reduce the scope of the PR (that may or may not be applicable for this PR) so that the new scope can be agreed on
- Use a tool other than Github to collaborate on proposals. In CF we have used Google docs for more than three years to discuss modifications to the broker API with great success.

bmelville:
Agree with Shannon, this PR has some issues still in discussion. It is also
not migrated to the single document format yet.

I think we've already reduced the scope sufficiently, but Shannon, if you
feel like we should move the discussion out of the PR into another venue I
am happy to start a doc or have phone conversation, or whatever is most
conducive.

shalako:
Brendan,

I agree that the scope of this PR currently seems appropriate; I only
mentioned the second option as (1) we've already employed it with this PR,
and (2) we should consider it in discussion of future PRs.

Shannon Coen
Product Manager, Cloud Foundry
Pivotal, Inc.

On Mon, Nov 28, 2016 at 11:07 AM, Brendan Melville <notifications@github.com

bmelville:
@shalako 

I love these stories and agree whole-heartedly with most of them. However, I think that separating schemas into create vs update doesn't quite match in my mind. Additionally, I think some of the stories don't work out-right with the broker API (but please correct any of my misplaced assumptions). I'll go into them in-line below.

>As a service provider, I would like to offer app developers a service plan that does not support
>configuration options for create or update

This just requires schemas to be optional, which I think is satisfied by the current spec proposal.

>As a service provider, I would like to offer app developers a service plan that offers a few non->required configuration options for create only, as I don't yet support changing these properties >once the instance is provisioned. The developer should be able to choose values for none, one, or >many of the available options.

This seems to require two things:

* **defaults**: satisfied by JSON schema itself
* **non-updatable parameters**: not satisfied by JSON schema and the crux of what we are discussing.

My proposal was that we have a single service instance schema, and add an extension field that allows the provider to specify immutable fields (i.e., fields that cannot be changed during update). This allows for a single schema for the parameters a service instance takes, which is a much more declarative paradigm in my mind.

Your alternative proposal is that we separate creation and update into two schemas, and my concern there is that these schemas could be wildly different (i.e., the provider could easily shoot him/herself in the foot).

To me it makes sense that a particular plan has a set of inputs--some of which are immutable--which is defined by a schema. Contrast this with the more imperative approach where each method on a service instance has a set of input parameters.

If the imperative approach is more in spirit with the existing SB API, then that is a fine answer, but it concerns me, as I think catalog systems will be building declarative software on top of the SB API, and will run into trouble with brokers that do not have identical or near identical schemas between create and update.

>As a service provider, I would like to offer app developers a service plan that allows developers to >update some, but not all of the configuration options the plan supports for create. There are some >properties I still don't support changing once the service instance is created. The developer should >be able to change the value of one property on update for which they accepted the default on >create.

I think this is in line with the above discussion, requirements being defaults and patching changes.

>As a service provider, I would like to require a few configuration options on create, in addition to >optional parameters. Once the service instance has been created, a subset of those required >options may optionally be updated, but a few of them cannot be changed.

The requirements for this seem to be required fields, which are satisfied by JSON schema, and the ability to differentiate between updatable and non-updatable fields, as in the discussion above.

>As a marketplace catalog provider, I want to enable my API clients to discover what parameters a >broker supports for create, update, and bind so that they can appropriately prompt users for values >of this parameters and construct the complete value of the parameters field sent with create, >update, and bind requests

This is just the requirement of schemas themselves, so definitely yes! :)

>As a marketplace catalog provider, I do not want to be responsible for validation of the parameters >field that my clients may send with create, update, and bind requests. I expect my API clients to >have used the schema the broker supplied and to have done their own client-side validation.

I don't think I agree with this. While there is value in doing client side validation and using schemas on the client side for visualization, etc., I don't think it is safe to say a catalog provider can rely on clients alone. One of the core values to schemas is the ability to do easy server-side validation and error handling, which to me is a necessity when building a broker (or any service).

>As a marketplace catalog provider, I want to enable my APIs client to discover the value of the >parameters field last successfully sent to the broker for the service instance or service binding, so >that my API clients can display to app developers the current configuration of these resources.

This I want to reflect back to you as a question. It seems that the SB API is **not** a RESTful API, and does not provide get/list methods for instances and bindings, so how is this story satisfied? It is one of the concerns at the top of my mind, i.e., not being able to query a broker for state.

shalako:
> My proposal was that we have a single service instance schema, and add an extension field that allows the provider to specify immutable fields (i.e., fields that cannot be changed during update). This allows for a single schema for the parameters a service instance takes, which is a much more declarative paradigm in my mind.

This seems like a reasonable alternative.

> > As a marketplace catalog provider, I do not want to be responsible for validation of the parameters >field that my clients may send with create, update, and bind requests. I expect my API clients to >have used the schema the broker supplied and to have done their own client-side validation.
>
> I don't think I agree with this. While there is value in doing client side validation and using schemas on the client side for visualization, etc., I don't think it is safe to say a catalog provider can rely on clients alone. One of the core values to schemas is the ability to do easy server-side validation and error handling, which to me is a necessity when building a broker (or any service).

Do not confuse Marketplace Catalog Provider with Broker Author. In CF, the Marketplace Provider is the CF operator and Broker authors are third parties. I am not suggesting that validation of the user inputs is entirely the user-client's responsibility. I'm just not sure whether it should be CF's responsibility (the marketplace). This puts more of a burden on the service-catalog (as it is known in K8, I believe), to understand the scheme and validate client requests. In CF, the parameters field is opaque; we just pass it through to the broker for validation. The broker should (of course) validate the value of this field.

> > As a marketplace catalog provider, I want to enable my APIs client to discover the value of the >parameters field last successfully sent to the broker for the service instance or service binding, so >that my API clients can display to app developers the current configuration of these resources.
> 
> This I want to reflect back to you as a question. It seems that the SB API is not a RESTful API, and does not provide get/list methods for instances and bindings, so how is this story satisfied? It is one of the concerns at the top of my mind, i.e., not being able to query a broker for state.

I have long since tired of discussions on strict definitions of REST, so I will take a pass on your declaration of what the broker API is and isn't. In CF we take a lean, product-drive approach to software; if the broker API doesn't have a GET endpoint it is because we never needed one. That isn't to say we couldn't add one later.

As above, when I mention the Marketplace Provider persona I am not talking about brokers. I'm talking about CF, or the service-catalog in K8s. In CF, we intend to enable user-clients to discover the value of the parameters field. This could be achieved by storing the value of the parameters field (CF does not currently) and enabling a client to retrieve it (CF already supports a GET on the service instance resource). Or we could add a GET to the broker API and have CF start calling the broker when its own GET service instance endpoint was hit. TBD.

bmelville:
This is now rebased onto spec.md

bmelville:
Ah yes, I apologize, I misinterpreted your stories as broker author stories, thanks for clarification.

I think the latter points in our discussion don't have specific bearing on the schemas, so based on your first comment, is it fair to say that we have agreement that we can maintain object-oriented schemas rather than method-oriented schemas?

shalako:
Seems like a mutable/immutable flag for each field or a list of the immutable fields would work. The only use case I imagine it doesn't cover are attributes that are not configurable on update but not create; I'm not aware of any examples of this and wouldn't expect this to be a common use case.

@avade @st3v ?

angarg12:
An 'immutable' extension seems clean and functional.

@shalako if I understood your use case, Marketplace is a client of SB API is that right? Then most likely it makes no sense for them to validate data. 

In any case the behaviour of this component would be outside the scope of this spec, wouldn't it? 

In fact, should the spec dictate if components MUST use schema for validation, or just define the existence of the schema and leave validation as an implementation detail by clients-servers?

bmelville:
In my mind it is important to define the ability to specify schema, but it
will be up to an individual broker or client to make use of it in a
meaningful way.

It is also up to us to define any standard tooling or examples to make it
easy for broker authors or clients to use it if they so choose.

Does that sound appropriate?
B.

avade:
> "It is also up to us to define any standard tooling or examples to make it
easy for broker authors or clients to use it if they so choose" - @bmelville 

👍 

> "An 'immutable' extension seems clean and functional" @angarg12 

How would that look in jsonschema a custom type? We could make this a "recommend" way in the spec via examples, as Brendan suggests.

There are going to be many weird and wonderful combinations of when and how parameters can be updated, so I expect we will need to come up with recommendations as they appear.

Service controllers that validate brokers with a meta-schema can then choose to accept registrations from brokers that use only certain jsonschema custom types.

angarg12:
> How would that look in jsonschema? A custom type?

I'm a little bit out of the loop with JSON schema but I was thinking about something like how "required" works now, where you list the required fields. Does that sound plausible and convenient?

shalako:
Closing in favor of proposal in #59, per our contribution process

PR #78, opened 2016-09-08
Trying to get the gobroker to work - first step of several

duglin:
But also did some clean-up.
Tried to clean-up the model stuff too - we were using the wrong stuff in places.

I'm trying to follow a pattern w.r.t. our schema/model:
- "scmodel" refers to /models/service_controler/*
- "sbmodel" refers to /models/service_broker/*
- "model" refers to "this exe's model".

both scmodel and sbmodel should JUST define what's at the API level.  Internal data structures should NOT be in those dirs. Internal data structures go into the exe's "model" dir.  While in some cases they are close, we can't assume they are the same.  And in cases where they are the same (if there are any) then the internal one can just inherit from the s*model one.

Technically "scmodel" shouldn't be at the top level since its our impl specific but that's ok - let's hope that perhaps those API will be useful to other/non-k8s impls  :-)

Also, trying to get into the habit of using Request/Response on the structs that go on the wire so its very clear that these are not meant to be internal data structure, they're for message flows. Its way too easy right now to grab the wrong struct because the names are the same.


MHBauer:
Why the change of `Credential` to `Interface{}`? Is it still going to serialize correctly?

brokers/go/Makefile
```diff
@@ -22,8 +22,8 @@ binary-local:
 	go get -d ./...
 	go build -o $(EXE) -tags netgo -installsuffix netgo gobroker.go
 
-broker:
-	docker run --net-host -ti -p 9090:9090 gobroker
+broker: $(EXE)
+	docker run --net=host -ti -p 9090:9090 gobroker
```
MHBauer: isn't `net=host` and allocating a specific port exclusive?

MHBauer:
do you want me to try and run this to see if it works with the core?

brokers/go/Makefile
duglin: hmm maybe not sure - will try w/o it...

duglin:
I switch from Credentials to interface because the spec doesn't mandate the Credential's struct's fields.  It needs to be wide-open from a spec perspective. As long as the struct behind the interface is done correctly then yes it should serialize ok.

duglin:
I updated the main comment - please see that for my ideas about cleaning up the model dirs.

brokers/go/Makefile
duglin: yep - you’re right - fixed it

MHBauer:
can't hurt anything at the moment. I'll try it when I get on site.

lgtm

duglin:
LGTM

PR #80, opened 2016-09-15
model/storage revamp

duglin:
Lots of changes...
- try to return the proper structs on the SC responses
- switch from pointers to IDs/strings for the "model" so that we don't have to have the entire catalog in memory. This means we'll have a lazy load policy.
- there was a lot of confusion in the code where we'd talk about "services" but we really meant "service instance", or visa-versa.  Tried to clean that up.
- complete support for creating a binding
- more clean-up w.r.t. aligning our structs with the API specs
- make it so that the storage plug-in never call back up into the server. I also would like to, at some point, make it so that all interactions with the storage layer is done thru a model.* object/func. But I saved that for later, but did clean up some stuff to make it easier to do.
- changed it so that when we look for something we don't return an error if its not there, we just return (nil,nil) so that we can easily know a 404 from a 500 type of situation - meaning 404 is when the return object is nil, not when 'err' is a certain type of error.
- got rid of the nil storage back-end since the in-memory should be good enough

I'm sure there's more but I can't remember  :-)


MHBauer:
too much to review all at once.

probably best to move the logic up into the controller and have storage be generic crud interface.

duglin:
LGTM

PR #83, opened 2016-09-19
Remove some CF-ism

duglin:
I only focused on the JSON/messages and not on the descriptive text. That's
a much bigger PR which we can do if we agree on this change first.

Signed-off-by: Doug Davis dug@us.ibm.com

### DIFF
--- a/specs/sb/access-control.md
+++ b/specs/sb/access-control.md
@@ -282,9 +282,10 @@ $ cf curl /v2/service_plan_visibilities -X 'POST' -d '{"service_plan_guid":"1113
     },
     "entity": {
         "service_plan_guid": "1113aa0-124e-4af2-1526-6bfacf61b111",
-        "organization_guid": "aaaa1234-da91-4f12-8ffa-b51d0336aaaa",
         "service_plan_url": "/v2/service_plans/1113aa0-124e-4af2-1526-6bfacf61b111",
-        "organization_url": "/v2/organizations/c54bf317-d791-4d12-89f0-b56d0936cfdc"
+        "context": {
+            "dept": "abc123"
+		}
     }
 }
 </pre>
--- a/specs/sb/api.md
+++ b/specs/sb/api.md
@@ -573,9 +573,9 @@ existing system.
 </thead>
 <tbody>
 <tr>
-  <td>organization_guid*</td>
-  <td>string</td>
-  <td>The Cloud Controller GUID of the organization under which the service is to be provisioned. Although most brokers will not use this field, it could be helpful in determining data placement or applying custom business rules.</td>
+  <td>context*</td>
+  <td>object</td>
+  <td>Platform specific contextual information under which the service is to be provisioned. Although most brokers will not use this field, it could be helpful in determining data placement or applying custom business rules.</td>
 </tr>
 <tr>
   <td>plan_id*</td>
@@ -586,11 +586,6 @@ existing system.
   <td>service_id*</td>
   <td>string</td>
   <td>The ID of the service within the catalog above.</td>
-</tr>
-<tr>
-  <td>space_guid*</td>
-  <td>string</td>
-  <td>Similar to organization_guid, but for the space.</td>
 </tr>
 <tr>
   <td>parameters</td>
@@ -609,11 +604,12 @@ existing system.
 
 <pre class="terminal">
 {
-  "organization_guid": "org-guid-here",
-  "plan_id":           "plan-guid-here",
-  "service_id":        "service-guid-here",
-  "space_guid":        "space-guid-here",
-  "parameters":        {
+  "context": {
+    "some_field": "some-contextual-data"
+  },
+  "plan_id":      "plan-guid-here",
+  "service_id":   "service-guid-here",
+  "parameters":   {
     "parameter1": 1,
     "parameter2": "value"
   }
@@ -623,11 +619,12 @@ existing system.
 ##### cURL #####
 <pre class="terminal">
 $ curl http://username:password@broker-url/v2/service_instances/:instance_id -d '{
-  "organization_guid": "org-guid-here",
-  "plan_id":           "plan-guid-here",
-  "service_id":        "service-guid-here",
-  "space_guid":        "space-guid-here",
-  "parameters":        {
+  "context": {
+    "some_field": "some-contextual-data"
+  },
+  "plan_id":      "plan-guid-here",
+  "service_id":   "service-guid-here",
+  "parameters":   {
     "parameter1": 1,
     "parameter2": "value"
   }
@@ -766,14 +763,9 @@ To enable this functionality, a broker declares support for each service by incl
   <td>ID of the service for the instance.</td>
 </tr>
 <tr>
-  <td>previous_values.organization_id</td>
-  <td>string</td>
-  <td>ID of the organization containing the instance.</td>
-</tr>
-<tr>
-  <td>previous_values.space_id</td>
-  <td>string</td>
-  <td>ID of the space containing the instance..</td>
+  <td>previous_values.context</td>
+  <td>object</td>
+  <td>Contextual data under which the instance is created.</td>
 </tr>
 <tr>
   <td>accepts_incomplete</td>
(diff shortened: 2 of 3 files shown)


specs/sb/access-control.md
```diff
@@ -266,12 +266,12 @@ In this example, the GUID of my-org is c54bf317-d791-4d12-89f0-b56d0936cfdc.
 
 To make a private plan available to a specific organization, run:
 
-`cf curl /v2/service_plan_visibilities -X POST -d '{"service_plan_guid":"SERVICE_PLAN_GUID","organization_guid":"ORG_GUID"}'`
+`cf curl /v2/service_plan_visibilities -X POST -d '{"service_plan_guid":"SERVICE_PLAN_GUID","context_guid":"CONTEXT_GUID"}'`
```
shalako: This is a cloud controller endpoint used for access control, and is unrelated to the service broker.

specs/sb/access-control.md
```diff
@@ -266,12 +266,12 @@ In this example, the GUID of my-org is c54bf317-d791-4d12-89f0-b56d0936cfdc.
...
 
 Example:
 
 <pre class="terminal">
-$ cf curl /v2/service_plan_visibilities -X 'POST' -d '{"service_plan_guid":"1113aa0-124e-4af2-1526-6bfacf61b111","organization_guid":"aaaa1234-da91-4f12-8ffa-b51d0336aaaa"}'
+$ cf curl /v2/service_plan_visibilities -X 'POST' -d '{"service_plan_guid":"1113aa0-124e-4af2-1526-6bfacf61b111","context_guid":"aaaa1234-da91-4f12-8ffa-b51d0336aaaa"}'
```
shalako: Same here.

specs/sb/api.md
```diff
@@ -573,9 +573,9 @@ existing system.
 </thead>
 <tbody>
 <tr>
-  <td>organization_guid*</td>
+  <td>context_guid*</td>
```
shalako: This may need to be a JSON object or array, as some brokers may need both the org and space. This enables billing reconciliation when a broker sends the CF operator a bill, even when the org and/or space have since been deleted (these objects are not soft-deleted) in CF.

specs/sb/access-control.md
duglin: doi - got too zealous with the global search and replace.  Fixed

specs/sb/access-control.md
duglin: fixed

specs/sb/api.md
duglin: I'd really prefer to put the burden on the platform to manage this. And, in fact, I would think it would be a bit of a security issue if the only source of truth for the space was the SB because then what if it lied? Wouldn't it be safer for the platform to always just send a GUID (which in some cases could be a SHA/HASH of the info) so that the SB couldn't muck with the info at all?

shalako: Interesting. I see what you mean. I'll talk to the folks I know are using this.

FWIW, we wouldn't be able to remove the org and space fields from the spec until v3 of the broker API, but we could introduce a replacement field and have CF send it as you have suggested, then start having consumers of the org and space fields move to the new one.

pmorie: > FWIW, we wouldn't be able to remove the org and space fields from the spec until v3 of the broker API, but we could introduce a replacement field and have CF send it as you have suggested, then start having consumers of the org and space fields move to the new one.

Stupid question: are we not discussing v3 of the API in this space?

specs/sb/api.md
```diff
@@ -573,9 +573,9 @@ existing system.
...
   <td>string</td>
-  <td>The Cloud Controller GUID of the organization under which the service is to be provisioned. Although most brokers will not use this field, it could be helpful in determining data placement or applying custom business rules.</td>
+  <td>The Cloud Controller GUID of the context under which the service is to be provisioned. Although most brokers will not use this field, it could be helpful in determining data placement or applying custom business rules.</td>
```
pmorie: I think it might be helpful to provide an example here, both of how this would be used within CF and another PaaS/orchestrator.

shalako:
Not necessarily. 

Where additive changes can be made in backward compatible ways, we will stay on the v2 line. Only when there is no other option but to make a backward incompatible change would we introduce v3.

pmorie:
@shalako that makes sense -- do you have minor versions of the API, or do you just re-release v2 and ensure that it's backward compatible?

shalako:
We have minor versions that are backward compatible in the same major series.

specs/sb/api.md
duglin: @shalako can you provide me with a good/real example that I can put there? I might be able to come up with one that's totally made-up but I'd prefer to include one that's based on a real implementation.

Would a SB use this during the billing cycle to tell a platform how much to charge the owner of each "context" ?

robertgmoss:
LGTM

duglin:
Updated PR to make 'context' an object instead of a GUID.

shalako:
Use cases for brokers using org and space:
- CF operator also operates a billing system that needs identifiers for the billable entity (org) and to group line items (space), even when these resources have been deleted from CF
- Service providers (broker authors) want to send email notifications to users of a service instance provisioned through a CF marketplace. As user membership of a space containing a service instance is transient (no user owns a service instance; they are owned by the space), the CF Notifications service will take an email sent to an alias for the space (e.g. `space-guid@cf.example.com`) and forward it to the addresses of the current members of the space.

We'll explore the first point a bit further and update use cases in this thread.

duglin:
Service Brokers want to limit the number of instances (or bindings) based on the context - e.g. just one instance per space - so we need some info in the message sent to the brokers to know which context the request is scoped to.

Other context information we pass in there includes the CF instance that the broker is connected to.  I realize there are other potential ways to do that (unique creds per env, unique path per env), but those don't scale from a mgmt perspective.

angarg12:
In last week's call some colleagues expressed their concern about allowing a field with arbitrary values in the create instance call.

Do we have a tangible proposal on how to avoid this or should we accept it as a necessary evil?

duglin:
I view the ability to put extra data in the 'context' as an extensibility point - and usually those are safe/good to have in spec.  I'd like to better understand what the concerns are with allowing this data.

shalako:
Recap of use cases...

Billing integration: as this integration is custom to Pivotal, brokers have to be designed specifically to do something special with org+space

Email notifications: for this use case org+space are provided as a convenience to service providers, enabling them to communicate with end users. Requires broker authors to implement  CF-specific logic.

Limit service instances to one per space: unless all platforms send the same field, brokers will have to have platform specific logic to fulfill this use case (space for CF, ? for K8)

Based on the above, we'd have to imagine brokers having to have specific support for platform features. Maybe its ok to have a set of core functionality all brokers support, and then extras that are platform specific? I want to avoid fostering brokers that can't be added to one platform or another. These platform specific features should be optional.

> Other context information we pass in there includes the CF instance that the broker is connected to. I realize there are other potential ways to do that (unique creds per env, unique path per env), but those don't scale from a mgmt perspective.

Doug, could you elaborate on this? Thanks!

shalako:
Feedback from the CF community, including many use cases in which org and space are leveraged: http://cf-dev.70369.x6.nabble.com/cf-dev-Do-your-service-brokers-leverage-the-Org-and-Space-fields-in-the-broker-API-td6056.html

angarg12:
So what is the general feeling on moving org and space id inside the context object?

duglin:
I'd like to suggest the following:
1 - after the spec rewrite is done, this PR is rebased on that
2 - the spec only mentions org/space as top-level deprecated/backwards-compat properties that are leveraging extensibility points
3 - we define a "CF profile" that defines org/space properties under "context" that CF impls MUST use
4 - we define a "K8s profile" that defines a "namespace" property under "context" that K8s impls MUST use.

angarg12:
@duglin where would these profiles be defined?

If provider specific details are described, I would like to see them defined outside the main spec body in an accessory doc.

duglin:
@angarg12 outside the main spec but still in the same git repo works for me.

duglin:
I haven't updated the spec.md doc since that's going thru some major changes. However, I did add the start of a `context` profile doc per our previous discussions.  I didn't touch the kubeneretes side of it yet, but took a stab at the CF stuff.

duglin:
See issue: https://github.com/openservicebrokerapi/servicebroker/issues/115

duglin:
Closing for now - per the new process. See https://github.com/openservicebrokerapi/servicebroker/issues/115

PR #87, opened 2016-09-20
Fix a panic or two

duglin:
when some maps were not initialized properly


duglin:
LGTM

PR #124, opened 2017-01-18
replacing old spec with latest; CF content removed

shalako:
This PR includes the minimum API spec docs with CF content removed. It also includes a release notes file to provide some history.

We can add more content to this doc later as we in the CF community revisit the remaining files in https://github.com/cloudfoundry/docs-services, as some of that content could be generic across platforms.

In order to maintain a seamless UX for the CF docs, we'll be pulling content from these files as partials. For this reason the files must be named with an underscore and have the suffixes `.html.md.erb` for the time being.

We do have the flexibility to replace html in `_api.html.md.erb` with markdown, as the CF docs apps should support it.

### DIFF
--- a/_api.md
+++ b/_api.md
@@ -69,20 +69,9 @@ The following sections describe catalog requests and responses in the Service Br
 
 ### Response ###
 
-<table border="1" class="nice">
-<thead>
-<tr>
-  <th>Status Code</th>
-  <th>Description</th>
-</tr>
-</thead>
-<tbody>
-<tr>
-  <td>200 OK</td>
-  <td>The expected response body is below.</td>
-</tr>
-</tbody>
-</table>
+| Status Code  | Description  |
+|---|---|
+| 200 OK  | The expected response body is below. |
 
 #### Body - Schema of Service Objects ####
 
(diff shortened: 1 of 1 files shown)


duglin:
@shalako what's the best way to view this file via github?

shalako:
I'm not sure why it isn't rendered.

Shannon Coen
Product Manager, Cloud Foundry
Pivotal, Inc.

duglin:
I think its the extension - changing it to .md seems to work for me.

duglin:
@shalako is it possible to rename this file so that it can display nicely in github but then rename it to its current name/suffix during the CF build process?

shalako:
The CF docs team is currently investigating how we can have the file this
working group manages be rendered by Github, while pulling it in as a
partial into the CF docs.

Shannon Coen
Product Manager, Cloud Foundry
Pivotal, Inc.

shalako:
CF docs can pull from files named with only `.md` extensions. Updated the PR. You should see it rendering correctly now. The files will need to remain named with an underscore prefix for now.

shalako:
PR has been updated; html tables replaced with markdown.

pmorie:
@shalako Could we do one of the following:

1.  do a `git mv` in this branch so that github can see that the spec.md file is being renamed and render diffs
2.  rename the file in a separate PR so that this branch will render

shalako:
Hi Paul, I can look into this `git mv` thing.

You can see the page render by clicking on View for `_api.md`, which will take you to my fork: https://github.com/shalako/servicebroker/blob/a2d276e7869bf29b7e7b08d3f4781754f129e374/_api.md. Could you clarify the problem?

shalako:
I haven't been able to figure out how to apply the `git mv` after I've already removed the spec.md file. I tried starting from a new clone of this repo, doing the `git mv`, then applying a saved copy of my forked content on top. But my fork rejected the push. So, I don't know how to move forward without closing the PR, reforking, then pushing to my fork again.

Is the old content that valuable? It could be argued that nothing matters before the content coming with this PR.

shalako:
@duglin In the minutes from today's call Alex says you had a question about the name of the spec doc. Could you clarify?

shalako:
Will resubmit the PR

duglin:
@shalako I was just wondering if the name of the spec could be something a bit more meaningful to this group and not set based on CF's build pipeline's needs.  But I see you opened a new PR for this so will look over there to see the latest.

PR #131, opened 2017-01-27
latest spec from CF master

shalako:
Replaces #124 

Includes latest API spec from CF. When this PR is merged, the CF docs will be updated to reference this repo for the pages: http://docs.cloudfoundry.org/services/api.html and http://docs.cloudfoundry.org/services/release-notes.html. At that point this repo will be the source of truth for the broker API spec, and future versions can be released from here.


angarg12:
LGTM

duglin:
LGTM

Funny side note: I did a quick search to see if CF was mentioned in there at all and its not, as expected. But, neither is "cloud"  - not sure what that means  :-)

bmelville:
LGTM

duglin:
3 LGTMs are good enough, merging....

PR #135, opened 2017-02-08
Fix release notes link

shalako:
Because the release notes file name has an underscore (as it is embedded in the CF docs as a partial), a link to the release notes in the spec doesn't work both in this repo and in the CF docs. 

Moved the link out of the spec, and added it to the README. In the CF docs we'll add a link to the release notes by other means.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 ![Open Service Broker API](https://www.openservicebrokerapi.org/wp-content/uploads/2016/12/osbapi_logo_concept3_wtm.png)
 
 [The Open Service Broker API specification](_spec.md)
+
 [Release Notes for past versions](_release-notes.md)
 
 ## Communications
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,8 +1,7 @@
 ![Open Service Broker API](https://www.openservicebrokerapi.org/wp-content/uploads/2016/12/osbapi_logo_concept3_wtm.png)
 
-This repo contains the Open Service Broker API specification.
-
-The specification can be found [here](_spec.md).
+[The Open Service Broker API specification](_spec.md)
+[Release Notes for past versions](_release-notes.md)
```
angarg12: Needs a line break so that it is displayed in two separate lines.

angarg12:
Both links are displayed in the same line. Add a line break and we can merge it.

shalako:
Fixed.

angarg12:
LGTM

Issue #144, opened 2017-03-08
Metadata spec appears to be missing from the spec


pmorie:
Likely removed inadvertently in #131

pmorie:
@shalako can we get this added back in?

shalako:
Are you referring to this: http://docs.cloudfoundry.org/services/catalog-metadata.html

vaikas-google:
The question was more about treatment of it, should we assume in the OSB that there are some expected fields, or should the entire object be treated as an opaque blob. We discussed this during the meeting today and it is a blob, little more details in the agenda doc.

duglin:
I think this can be closed now due to #235 being merged.

Reopen if I'm mistaken.

PR #149, opened 2017-03-14
Address #144 by clarifying that the metadata is opaque blob.

vaikas-google:
Also fix the 'bullets' section of examples to conform to what CF docs have for consistency.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -122,7 +122,7 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
 | tags  |  array-of-strings | Tags provide a flexible mechanism to expose a classification, attribute, or base technology of a service, enabling equivalent services to be swapped out without changes to dependent logic in applications, buildpacks, or other services. E.g. mysql, relational, redis, key-value, caching, messaging, amqp.  |
 |  requires | array-of-strings  | A list of permissions that the user would have to give the service, if they provision it. The only permissions currently supported are <code>syslog\_drain</code>, <code>route\_forwarding</code> and <code>volume\_mount</code>.  |
 | bindable*  | boolean  | Specifies whether instances of the service can be bound to applications. This specifies the default for all plans of this service. Plans can override this field (see [Plan Object](#PObject)).  |
-| metadata  |  object | A list of metadata for a service offering.  |
+| metadata  |  JSON object | An opaque object of metadata for a service offering. Controller treats this as a blob. Note that there are conventions in existing brokers and controllers for fields that aid in the display of catalog data. |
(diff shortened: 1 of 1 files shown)


pmorie:
LGTM

angarg12:
LGTM

spec.md
```diff
@@ -197,13 +197,11 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
...
             }
          ],
-        "bullets": [{
-            "content": "Shared fake server"
-          }, {
-            "content": "5 TB storage"
-          }, {
-            "content": "40 concurrent connections"
-        }]
+        "bullets": [
```
duglin: why did you change this?  The old version was valid.

vaikas-google: I tried to explain in the PR description, but, while this is a blob and yes, this is technically valid, it is:
 1. silly the way it's written
 2. does not comform to the existing Cloud Foundry convention, where bullets is an array of strings, not a map, so unless there's a strong reason for it to diverge, seemed like they should at least be consistent.
https://docs.cloudfoundry.org/services/catalog-metadata.html

Their examples are:
              "metadata":{
                  "bullets":[
                     "20 GB of messages",
                     "20 connections"
                  ],

Which makes more sense than duplicating 'content' over and over for no reason. It's confusing.

duglin: ah ok - I missed the connection - thanks

duglin:
LGTM

avade:
Could we at least mention the existing CF community standards an FYI (https://docs.cloudfoundry.org/services/catalog-metadata.html) in the spec? 

Otherwise it seems fine.

avade:
Could we at least mention the existing CF community standards an FYI (https://docs.cloudfoundry.org/services/catalog-metadata.html) in the spec? 

Otherwise it seems fine.

vaikas-google:
I did mention that there are existing community standards, are you asking for a link to be included?

pmorie:
I think a link to somewhere where the existing community standards are document would be very useful.

vaikas-google:
Ok, added a link, ptal.

avade:
lgtm

angarg12:
It seems like pullapprove didn't pick up the LGTM, maybe we need to close and reopen the PR.

duglin:
LGTM
agai

duglin:
@pmorie need your LGTM again

pmorie:
LGTM

vaikas-google:
@avade Looks like you LGTMd it, but it didn't result in a pullapprove, is there something else that you need to do to convert that into pullapprove?

PR #157, opened 2017-03-29
fixes header issue

abbyachau:


spec.md
```diff
@@ -22,11 +21,9 @@
   - [Broker Errors](#broker-errors)
   - [Orphans](#orphans)
 
-<div id="changes"/> 
-## Changes ##
+## <div id="changes"/> Changes
```
duglin: doesn't md add the anchors automatically for us?

abbyachau: @duglin This pr attempts to fix the styling for all the headers in the current spec.md doc. Thanks, pls let me know if you want any further clarification.

duglin: my point was that I don't think you need to manually add the "div" here.   In fact, this this "div" in there it gets an anchor named "-changes" which might throw off tooling that assumes normal github md processing.

angarg12: I agree with @duglin, we should use the md anchors instead of divs.

abbyachau: @duglin and @angarg12 thanks both. changes made per your feedback - this last commit includes the removal of the divs and minor changes to make the markdown anchors work properly. if you spot anything else, pls don't hesitate to get in touch. thanks again.

angarg12:
Thanks a lot, unless I missed something it LGTM

vaikas-google:
LGTM

jmrodri:
LGTM what's required to merge this? the doc is annoying to read without these changes.

duglin:
LGTM

@mattmcneeney in your PR I think you've added some div's - you may want to remove those to align with this change.

mattmcneeney:
Thanks for the heads up @duglin. Have removed added divs in the branches I've been working on.

pmorie:
LGTM

PR #162, opened 2017-04-06
Brokers must validate parameters passed to provision, update and bind

mattmcneeney:
With work underway regarding service brokers declaring a supported schema for the parameters field ( #59 ), I think it makes sense that brokers ensure that any parameters provided via the provision, update and bind calls are valid.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -356,7 +356,7 @@ The `:instance_id` of a service instance is provided by the platform. This ID wi
 |---|---|---|
 | service_id*  | string  | The ID of the service (from the catalog). Must be globally unique.  |
 | plan_id*  | string  | The ID of the plan (from the catalog) for which the service instance has been requested. Must be unique to a service.  |
-|  parameters |  JSON object | Configuration options for the service instance. Controller treats this as a blob. Note that there are [conventions](https://docs.cloudfoundry.org/services/catalog-metadata.html) in existing brokers and controllers for fields that aid in the display of catalog data. Brokers must validate these parameters to ensure that they are valid. |
+|  parameters |  JSON object | Configuration options for the service instance. Controller treats this as a blob. Note that there are [conventions](https://docs.cloudfoundry.org/services/catalog-metadata.html) in existing brokers and controllers for fields that aid in the display of catalog data. Brokers should ensure that the client has provided valid configuration parameters and values for the operation. |
 |  accepts_incomplete | boolean  | A value of true indicates that the marketplace and its clients support asynchronous broker operations. If this parameter is not included in the request, and the broker can only provision an instance of the requested plan asynchronously, the broker should reject the request with a 422 as described below.  |
 | organization_guid*  | string  | The platform GUID for the organization under which the service is to be provisioned. Although most brokers will not use this field, it may be helpful for executing operations on a user's behalf.  |
 |  space_guid* |  string |  The identifier for the project space within the platform organization. Although most brokers will not use this field, it may be helpful for executing operations on a user's behalf. |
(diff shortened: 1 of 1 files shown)


mattmcneeney:
I've fixed my membership regarding the cfdreddbot message. Please can an owner close and reopen this PR, thanks!

spec.md
```diff
@@ -570,7 +570,7 @@ the resource it creates.
 | plan_id*  | string  | ID of the plan from the catalog.  |
 | app_guid  | string  | Deprecated in favor of <code>bind\_resource.app\_guid</code>. GUID of an application associated with the binding to be created.  |
 | bind_resource  | JSON object  | A JSON object that contains data for platform resources associated with the binding to be created. Current valid values include <code>app\_guid</code> for [credentials](#types-of-binding) and <code>route</code> for [route services](#route_services).  |
-| parameters | JSON object  |  Configuration options for the service binding. An opaque object, controller treats this as a blob.  |   |
+| parameters | JSON object  |  Configuration options for the service binding. An opaque object, controller treats this as a blob. Brokers must validate these parameters to ensure that they are valid. |   |
```
avade: "Brokers should ensure that the client has provided valid configuration parameters and values for the operation"

avade: I think this is clearer advice.

mattmcneeney: I agree, thanks. I've updated the branch!

avade:
LGTM

angarg12:
LGTM

duglin:
Even though we're not using RFC2119 keywords (yet), I'm glad you used "should" instead of "must". I was a little uncomfortable with the idea that an implementation couldn't have really smart error/typo recovery if they wanted.

So, what should they do if they do find something wrong? Do we want to RECOMMEND that they generate an error and reject the request? W/o guidance someone could interpret it as valid to just ignore the bad parameters and keep processing.  I'd suggest we generate an error and reject the request.

Also, should this be moved to a more global spot in the spec and apply to all fields?

vaikas-google:
One would expect that any parameters (not 'parameters' but any input) passed in should be validated by the broker and handled appropriately? Isn't the point we're trying to make here that even though there's a schema that you can specify for the parameters you expect, ultimately the broker is responsible for validating the configuration parameters because to the controller, it's opaque. Even if the controller could validate it, a broker should always validate its inputs.
I think we might want to give guidance on what the error message returned should look like, for example, we might look at the k8s api conventions and use that as a guidance on how to help the developers to convey meaningful error messages back to the controller.
https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#validation

mattmcneeney:
I agree that we should give guidance on returning meaningful error messages across the board. Should we handle that in a separate issue since that applies to the whole spec rather than just input parameters?

mattmcneeney:
Just raising attention to this PR again. Discussion above suggests we need more guidance on returning meaningful error message across the API, but is this a useful standalone addition for now?

avade:
> Discussion above suggests we need more guidance on returning meaningful error message across the API, but is this a useful standalone addition for now?

Yes, I think that adds value as a standalone addition. The input in the config parameters is not crafted by the platform, so I feel more prone to error.

Can you change `should` to `SHOULD` to meet RFC2119 keywords spec.

mattmcneeney:
All `SHOULD`s are now uppercase.

angarg12:
Is this ready for another round of lgtm?

mattmcneeney:
I hope so!

duglin:
LGTM

mattmcneeney:
Please can we get two more on this (mainly so I don't have to fix more merge conflicts!) 😄

avade:
LGTM

avade:
@fmui, @pmorie, @shalako, @vaikas - needs one more LGTM.

vaikas-google:
LGTM

Issue #137, opened 2017-02-16
Support async binding operations


bmelville:
Binding to an instance may encapsulate a long running workflow, similar to creating an instance. This might include provisioning new user accounts in a service, setting up firewalls, etc. It seems like the Bind API would benefit from the same last_operation model that Provision uses, and I'd like to get feedback from others who may have more context around this.

Latest proposal:  https://github.com/mattmcneeney/servicebroker/blob/async-bindings/spec.md#binding

ccemeraldeyes:
+1. I have already run into a scenario where I needed async binding (Google CloudSQL).

angarg12:
We also had the situation where bindings need to return information, potentially asynchronously.

This raises the question of whether it would be better to implement the last_operation function, or a fully fledged GET method on bindings.

Would a GET operation cover your use cases?

RamakrishnanArun:
Would that be adding additional information in the response to the GET /last_operation?

avade:
I would advocate that we follow the current async pattern of using last
operation.

Adding a new endpoint to get bindings would also mean the bindings are
stored in a retrievable way, which is likely a security issue.

avade:
Unless there is any further discussion I would suggest that @bmelville creates the proposed spec changes on a branch. This can then be shared with the cf-dev mailing list.

I would support this feature. 

Any concerns @shalako?

avade:
@vaikas-google will you be picking up drafting the spec changes in a branch?

vaikas-google:
I think so. Though for the next few days I'm not able to get to this due to
other commitments still being on my plate, but I'll tackle this as soon as
I get a chance.

avade:
It is also interesting the spec doesn't include timeout information.

Currently, Cloud Foundry times out binding requests after 60 seconds by default. This is configurable by the operator on a global basis, meaning it applies to all brokers.

https://github.com/cloudfoundry/cloud_controller_ng/blob/master/bosh/jobs/cloud_controller_ng/spec#L570

@pmorie does service catalog have any request timeout logic that you are enforcing?

mattmcneeney:
I've created a [branch](https://github.com/avade/servicebroker/tree/async-binding) that adds async binding and unbinding to the spec: [View proposed spec](https://github.com/avade/servicebroker/blob/async-binding/spec.md) (or [changes](https://github.com/avade/servicebroker/commit/497ecdaa7307dc05b0f4448c2d51bbe1dbe5a684)).

[...]

Issue #159, opened 2017-04-04
Fetching current input parameters values (for a service or a binding)


mattmcneeney:
Following on from [Issue 59 (Service broker can declare supported schema for parameters field)](https://github.com/openservicebrokerapi/servicebroker/issues/59), we are expecting that platforms will want to be able to fetch the current values of any service or binding input parameters.

How could this be integrated into the service broker API?

---

**Potential idea**

Add two new endpoints:
- `GET /v2/service_instances/:instance_id` to fetch input parameters for a service.
- `GET /v2/service_instances/:instance_id/service_bindings/:binding_id` to fetch input parameters for a binding.

Concerns:
- Security
- Service brokers having to maintain some kind of state

---

### Proposal Doc 
(edit by @avade)

[Proposal: Fetching Configuration Parameters](https://docs.google.com/document/d/13EviTJXm9-P-SBqeGexyEy_wclKPCr1S3ZGVDpaC0Tw/edit?usp=sharing)

angarg12:
This seems to come back to the bigger question of whether OSB should provide a full REST API. Also I think that the common feeling around here is not to require brokers to keep state.

If we have specific use cases where clients require to retrieve the value of parameters it would make for a much compelling argument in favour of this requirement, instead of guessing a possible future need.

In any case it is probably better to raise this question in the weekly meeting.

mattmcneeney:
Of course, I've added it to the agenda for today's call. There are some use cases within Cloud Foundry whereby app developers would want to retrieve the configuration of their service instances and bindings. It will be interesting to hear if anyone else has run into this problem.

gberche-orange:
@mattmcneeney 

> Following on from Issue 59 (Service broker can declare supported schema for parameters field), we are expecting that platforms will want to be able to fetch the current values of any service or binding input parameters.
> How could this be integrated into the service broker API?

SB receive service instance and service binding parameters as in the ``parameters`` request field of the https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#provisioning request and would perform validation of received params against the schema declared in the catalog. Platform may optionally also validate the parameters field as they flow them to brokers.

I'm not sure to understand why platforms would need to fetch the current values of parameters of any previously provisioned service instance or binding. Can you please detail the rationale behind this ?

gberche-orange:
Use case mentionned during the WG call is that client UIs could display previously submitted parameters fields as to improve “update service instance” UX.

[...]

PR #164, opened 2017-04-07
Support async binding operations

mattmcneeney:
Binding to an instance may encapsulate a long running workflow, similar to provisioning. Discussions in issue #137 have resulted in:
* Changing the **Binding** endpoint to support asynchronous operations.
* Adding a new `last_operation` endpoint for bindings (`GET /v2/service_instances/:instance_id/service_bindings/:binding_id/last_operation`)

Feedback welcome as always.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -13,8 +13,9 @@
   - [Synchronous and Asynchronous Operations](#synchronous-and-asynchronous-operations)
     - [Synchronous Operations](#synchronous-operations)
     - [Asynchronous Operations](#asynchronous-operations)
-  - [Polling Last Operation](#polling-last-operation)
-    - [Polling Interval and Duration](#polling-interval-and-duration)
+  - [Polling Last Service Instance Operation](#polling-last-service-instance-operation)
+  - [Polling Last Service Binding Operation](#polling-last-service-binding-operation)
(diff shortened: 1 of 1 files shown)


mattmcneeney:
Issues this PR does not address:
1. How long should service brokers maintain the result of a binding operation? If the `last_operation` binding endpoint is called multiple times, should the result (ie. credentials) always be returned?
1. Would it make more sense to use a new endpoint to fetch information on bindings (see [proposed spec)](https://github.com/openservicebrokerapi/servicebroker/compare/master...avade:async-binding-with-additional-endpoint) once the `last_operation` says the operation has completed?

I would propose that we review this change to the spec first as this brings binding in line with provision, and then we can discuss adding additional endpoints to fetch per-service-instance or per-binding information (eg. credentials, health, etc).

duglin:
Can you rebase and squash?

mattmcneeney:
@duglin done, but be aware the PR contains a few "tidy-ups" of various double spaces and unnecessary `#`s. I can go through and split these out if everyone would like (although that would be _painful_).

mattmcneeney:
Closed as we will continue down this path on the back of #159 where `GET` endpoints are being added for service instances and bindings.

PR #165, opened 2017-04-07
Service broker can declare supported schema for parameters field

mattmcneeney:
See issue #59.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -228,9 +228,49 @@ how platforms might expose these values to their users.
 | metadata | JSON object | An opaque object of metadata for a service plan. Controller treats this as a blob. Note that there are [conventions](profile.md#service-metadata) in existing brokers and controllers for fields that aid in the display of catalog data. |
 | free | boolean | When false, service instances of this plan have a cost. The default is true. |
 | bindable | boolean | Specifies whether service instances of the service plan can be bound to applications. This field is OPTIONAL. If specified, this takes precedence over the `bindable` attribute of the service. If not specified, the default is derived from the service. |
+| [schemas](#schema-object) | object | Schema definitions for service instances and bindings for the plan. |
 
 
 \* Fields with an asterisk are REQUIRED.
+
+##### Schema Object
+
+| Response field | Type | Description |
+| --- | --- | --- |
+| [service_instance](#service-instances-object) | object | The schema definitions for creating and updating a service instance. |
+| [service_binding](#service-bindings-object) | object | The schema definition for creating a service binding. Used only if the service plan is bindable. |
+
+
+##### Service Instances Object
+
+| Response field | Type | Description |
+| --- | --- | --- |
+| [create](#input-parameters-object) | object | The schema definition for creating a service instance. |
+| [update](#input-parameters-object) | object | The schema definition for updating a service instance. |
+
+
+##### Service Bindings Object
+
+| Response field | Type | Description |
+| --- | --- | --- |
+| [create](#input-parameters-object) | object | The schema definition for creating a service binding. |
+
+
+##### Input Parameters Object
+
+| Response field | Type | Description |
+| --- | --- | --- |
+| parameters | JSON schema object | The schema definition for the input parameters. Each input parameter is expressed as a property within a JSON object. |
+
+The following rules apply if `parameters` is included anywhere in the catalog:
+* Platforms MUST support at least
+[JSON Schema draft v4](http://json-schema.org/).
+* Platforms SHOULD be prepared to support later versions of JSON schema also.
+* The `$schema` key MUST be present in the schema declaring the version of JSON
+schema being used.
+* Schemas MUST NOT contain any external references.
+* Schemas MUST NOT be larger than 64kB.
+
 
 ```
 {
@@ -284,6 +324,48 @@ how platforms might expose these values to their users.
           "5 TB storage",
           "40 concurrent connections"
         ]
+      },
+      "schemas": {
+        "service_instance": {
+          "create": {
+            "parameters": {
+              "$schema": "http://json-schema.org/draft-04/schema#",
+              "type": "object",
+              "properties": {
+                "billing-account": {
+                  "description": "Billing account number used to charge use of shared fake server.",
+                  "type": "string"
+                }
+              }
+            }
+          },
+          "update": {
+            "parameters": {
+              "$schema": "http://json-schema.org/draft-04/schema#",
+              "type": "object",
+              "properties": {
+                "billing-account": {
+                  "description": "Billing account number used to charge use of shared fake server.",
+                  "type": "string"
+                }
+              }
+            }
+          }
+        },
+        "service_binding": {
+          "create": {
+            "parameters": {
+              "$schema": "http://json-schema.org/draft-04/schema#",
+              "type": "object",
+              "properties": {
+                "billing-account": {
+                  "description": "Billing account number used to charge use of shared fake server.",
+                  "type": "string"
+                }
+              }
+            }
+          }
+        }
       }
     }, {
       "name": "fake-plan-2",
@@ -457,7 +539,7 @@ broker will use it to correlate the resource it creates.
 | context | object | Platform specific contextual information under which the service instance is to be provisioned. Although most brokers will not use this field, it could be helpful in determining data placement or applying custom business rules. `context` will replace `organization_guid` and `space_guid` in future versions of the specification - in the interim both SHOULD be used to ensure interoperability with old and new implementations. |
 | organization_guid* | string | Deprecated in favor of `context`. The platform GUID for the organization under which the service instance is to be provisioned. Although most brokers will not use this field, it might be helpful for executing operations on a user's behalf. MUST be a non-empty string. |
 | space_guid* | string | Deprecated in favor of `context`. The identifier for the project space within the platform organization. Although most brokers will not use this field, it might be helpful for executing operations on a user's behalf. MUST be a non-empty string. |
-| parameters | JSON object | Configuration options for the service instance. Controller treats this as a blob. Brokers SHOULD ensure that the client has provided valid configuration parameters and values for the operation. |
+| parameters | JSON object | Configuration options for the service instance. Brokers SHOULD ensure that the client has provided valid configuration parameters and values for the operation. |
 
 \* Fields with an asterisk are REQUIRED.
 
@@ -557,7 +639,7 @@ It MUST be the ID a previously provisioned service instance.
 | context | object | Contextual data under which the service instance is created. |
 | service_id* | string | The ID of the service (from the catalog). MUST be globally unique. MUST be a non-empty string. |
 | plan_id | string | The ID of the plan (from the catalog) for which the service instance has been requested. MUST be unique within the service. If present, MUST be a non-empty string. If this field is not present in the request message, then the broker MUST NOT change the plan of the instance as a result of this request. |
-| parameters | JSON object | Configuration options for the service instance. An opaque object, controller treats this as a blob. Brokers SHOULD ensure that the client has provided valid configuration parameters and values for the operation. If this field is not present in the request message, then the broker MUST NOT change the parameters of the instance as a result of this request. |
+| parameters | JSON object | Configuration options for the service instance. Brokers SHOULD ensure that the client has provided valid configuration parameters and values for the operation. If this field is not present in the request message, then the broker MUST NOT change the parameters of the instance as a result of this request. |
 | previous_values | object | Information about the service instance prior to the update. |
 | previous_values.service_id | string | Deprecated; determined to be unnecessary as the value is immutable. ID of the service for the service instance. If present, MUST be a non-empty string. |
 | previous_values.plan_id | string | ID of the plan prior to the update. If present, MUST be a non-empty string. |
@@ -717,7 +799,8 @@ it to correlate the resource it creates.
 | plan_id* | string | ID of the plan from the catalog. MUST be a non-empty string. |
 | app_guid | string | Deprecated in favor of `bind_resource.app_guid`. GUID of an application associated with the binding to be created. If present, MUST be a non-empty string. |
 | bind_resource | JSON object | A JSON object that contains data for platform resources associated with the binding to be created. See [Bind Resource Object](#bind-resource-object) for more information. |
-| parameters | JSON object | Configuration options for the service binding. An opaque object, controller treats this as a blob. Brokers SHOULD ensure that the client has provided valid configuration parameters and values for the operation. |
+| parameters | JSON object | Configuration options for the service binding. Brokers SHOULD ensure that the client has provided valid configuration parameters and values for the operation. |
+
 
 ##### Bind Resource Object
 


spec.md
```diff
@@ -143,9 +143,40 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
...
 | name*  | string  | The CLI-friendly name of the plan. Must be unique within the service. All lowercase, no spaces.  |
 | description*  | string  | A short description of the plan.  |
-|  metadata | JSON object  | An opaque object of metadata for a service plan. Controller treats this as a blob. Note that there are (conventions)[https://docs.cloudfoundry.org/services/catalog-metadata.html] in existing brokers and controllers for fields that aid in the display of catalog data. |
+|  metadata | object  | A list of metadata for a service plan.  |
 | free  | boolean  | When false, instances of this plan have a cost. The default is true  |
 |  bindable | boolean  |  Specifies whether instances of the service plan can be bound to applications. This field is optional. If specified, this takes precedence over the <tt>bindable</tt> attribute of the service. If not specified, the default is derived from the service. |
+| [schemas](#SObject) | object | Schema definitions for service instances and bindings for the plan. |
+
+
+##### Schema Object <a name="SObject"></a> #####
```
pmorie: Markdown automatically creates anchors for you - do you still need this?

mattmcneeney: Nope; I was just being consistent with the current spec (I know there is an outstanding PR that addresses this). I'll fix in this branch!

duglin:
rebase needed (and squash too)

mattmcneeney:
@duglin how's this?

spec.md
```diff
@@ -196,12 +228,56 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
...
+      "schemas": {
+        "service_instances": {
+          "create": {
+            "parameters": {
+              "$schema": "http://json-schema.org/draft-04/schema#",
+              "type": "object",
+              "properties": {
+                "billing-account": {
+                  "description": "Billing account number used to charge use of shared fake server.",
+                  "type": "String"
```
Samze: "String" should be lowercase "string"

spec.md
```diff
@@ -196,12 +228,56 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
...
+        },
+        "service_bindings": {
+          "create": {
+            "parameters": {
+              "$schema": "http://json-schema.org/draft-04/schema#",
+              "type": "object",
+              "properties": {
+                "billing-account": {
+                  "description": "Billing account number used to charge use of shared fake server.",
+                  "type": "String"
```
Samze: "String" should be lowercase "string"

spec.md
```diff
@@ -196,12 +228,56 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
...
+            }
+          },
+          "update": {
+            "parameters": {
+              "$schema": "http://json-schema.org/draft-04/schema#",
+              "type": "object",
+              "properties": {
+                "billing-account": {
+                  "description": "Billing account number used to charge use of shared fake server.",
+                  "type": "String"
```
Samze: "String" should be lowercase "string"

spec.md
```diff
@@ -122,13 +122,13 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
...
 |  requires | array-of-strings  | A list of permissions that the user would have to give the service, if they provision it. The only permissions currently supported are <code>syslog\_drain</code>, <code>route\_forwarding</code> and <code>volume\_mount</code>.  |
-| bindable*  | boolean  | Specifies whether instances of the service can be bound to applications. This specifies the default for all plans of this service. Plans can override this field (see [Plan Object](#PObject)).  |
-| metadata  |  JSON object | An opaque object of metadata for a service offering. Controller treats this as a blob. Note that there are (conventions)[https://docs.cloudfoundry.org/services/catalog-metadata.html] in existing brokers and controllers for fields that aid in the display of catalog data. |
-| [dashboard_client](#DObject)  |  object |  Contains the data necessary to activate the Dashboard SSO feature for this service |
+|  bindable | boolean  |  Specifies whether instances of the service plan can be bound to applications. This field is OPTIONAL. If specified, this takes precedence over the `bindable` attribute of the service. If not specified, the default is derived from the service. |
```
duglin: Are we allowed to make this optional? Doesn't that break backwards compat?

duglin: `... of the service.` (3rd sentence) - did you mean "plan" instead of "service" here?

spec.md
```diff
@@ -137,16 +137,47 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
...
-##### Plan Object <a name="PObject"></a> #####
+##### Plan Object
 
 |  Response field | Type  | Description  |
 |---|---|---|
 | id*  | string  | An identifier used to correlate this plan in future requests to the broker. This MUST be globally unique within a platform marketplace. Using a GUID is RECOMMENDED.  |
 | name*  | string  | The CLI-friendly name of the plan. MUST be unique within the service. All lowercase, no spaces.  |
 | description*  | string  | A short description of the plan.  |
-|  metadata | JSON object  | An opaque object of metadata for a service plan. Controller treats this as a blob. Note that there are (conventions)[https://docs.cloudfoundry.org/services/catalog-metadata.html] in existing brokers and controllers for fields that aid in the display of catalog data. |
+|  metadata | object  | A list of metadata for a service plan.  |
```
duglin: Why drop the note?  Perhaps we should move that note into the spec?

spec.md
```diff
@@ -568,7 +644,7 @@ the resource it creates.
 | plan_id*  | string  | ID of the plan from the catalog.  |
 | app_guid  | string  | Deprecated in favor of <code>bind\_resource.app\_guid</code>. GUID of an application associated with the binding to be created.  |
 | bind_resource  | JSON object  | A JSON object that contains data for platform resources associated with the binding to be created. Current valid values include <code>app\_guid</code> for [credentials](#types-of-binding) and <code>route</code> for [route services](#route_services).  |
-| parameters | JSON object  |  Configuration options for the service binding. An opaque object, controller treats this as a blob.  |   |
+| parameters | JSON object  |  Configuration options for the service binding. |   |
```
duglin: I think there's an extra ` |` at the end - was there before too.

mattmcneeney:
Yeah I've actually fixed that in another PR that should go in before this
one! (Async bindings)

spec.md
```diff
@@ -137,16 +137,47 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
...
+|  metadata | JSON object  | An opaque object of metadata for a service plan. Controller treats this as a blob. Note that there are [conventions](https://docs.cloudfoundry.org/services/catalog-metadata.html) in existing brokers and controllers for fields that aid in the display of catalog data. |
 | free  | boolean  | When false, instances of this plan have a cost. The default is true  |
-|  bindable | boolean  |  Specifies whether instances of the service plan can be bound to applications. This field is OPTIONAL. If specified, this takes precedence over the <tt>bindable</tt> attribute of the service. If not specified, the default is derived from the service. |
+|  bindable | boolean  |  Specifies whether instances of the service plan can be bound to applications. This field is optional. If specified, this takes precedence over the `bindable` attribute of the service. If not specified, the default is derived from the service. |
```
duglin: s/optional/OPTIONAL/

spec.md
mattmcneeney: Not sure where this came from. Rebase mess up I think. Sorting now...

spec.md
mattmcneeney: I've removed this. It shouldn't have been added in this PR.

spec.md
```diff
@@ -137,16 +137,47 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
...
+|  metadata | JSON object  | An opaque object of metadata for a service plan. Controller treats this as a blob. Note that there are [conventions](https://docs.cloudfoundry.org/services/catalog-metadata.html) in existing brokers and controllers for fields that aid in the display of catalog data. |
 | free  | boolean  | When false, instances of this plan have a cost. The default is true  |
-|  bindable | boolean  |  Specifies whether instances of the service plan can be bound to applications. This field is OPTIONAL. If specified, this takes precedence over the <tt>bindable</tt> attribute of the service. If not specified, the default is derived from the service. |
+| bindable | boolean | Specifies whether instances of the service plan can be bound to applications. This field is OPTIONAL. If specified, this takes precedence over the `bindable` attribute of the service. If not specified, the default is derived from the service. |
```
mattmcneeney: This is left alone now @duglin

spec.md
```diff
@@ -568,7 +644,7 @@ the resource it creates.
 | plan_id*  | string  | ID of the plan from the catalog.  |
 | app_guid  | string  | Deprecated in favor of <code>bind\_resource.app\_guid</code>. GUID of an application associated with the binding to be created.  |
 | bind_resource  | JSON object  | A JSON object that contains data for platform resources associated with the binding to be created. Current valid values include <code>app\_guid</code> for [credentials](#types-of-binding) and <code>route</code> for [route services](#route_services).  |
-| parameters | JSON object  |  Configuration options for the service binding. An opaque object, controller treats this as a blob. Brokers SHOULD ensure that the client has provided valid configuration parameters and values for the operation. |   |
+| parameters | JSON object | Configuration options for the service binding. Brokers SHOULD ensure that the client has provided valid configuration parameters and values for the operation. |
```
pmorie: Is this an editing/format change or related to this PR?

mattmcneeney:
Totally agree @pmorie. I'll remove formatting changes from this now, and then we can hopefully merge #192 and then rebase this one with the fixed formatting.

mattmcneeney:
This PR has been updated following last weeks' conversations. But we will need to add more to this surrounding the "meta schema" work.

angarg12:
Will you push the changes to this PR, or would you rather merge this one and start a new conversation in a follow up?

mattmcneeney:
@angarg12 I think we can continue in this PR, it'll just need some more changes before we review

mattmcneeney:
OK, so @Samze and @ablease have been doing some validation around the new schemas proposal and have some interesting feedback for us:
* **Resource-level schemas** are:
  * Good for UIs, as they are able to show all config parameters for a service or binding, including those that cannot be changed
  * But makes validation very difficult
    * Validation libraries don't support meta-schems or the additional `creatable` and `updateble` flags
    * How do we deal with required on create but not required on update
* **Method-leve schemas** (original proposal) are:
  * OK for UIs, but they have to be clever when prepopulating forms and can't easily show other resource information (unless we have a separate schema for that)
  * Good for validation libraries (you just have to pass in the create or update schema with the parameters and any library will give you a valid/invalid response)

We know that there will be few UIs to many service brokers, so it's important to keep service brokers easy to implement. Resource-level schemas do not do this, as it makes validation tricky and means broker authors have to understand how to apply the meta-schema to their resource schemas. Method-level schemas are easy to implement, may have duplication (but we don't think this is an important issue), and make building UIs a bit harder, but there are workarounds that don't require super complex logic.

So, in conclusion, we think we should continue with the proposal that was in place before the face-to-face last week. We can talk about this on today's call but wanted to give you all the heads up in case you had time to think about this before the call. Thanks!

pmorie:
@jwforres, what are your thoughts about the feedback here: https://github.com/openservicebrokerapi/servicebroker/pull/165#issuecomment-303367440

jwforres:
From our perspective we have the same validation concerns as a broker as far as client-side validation of JSON schema.  If we try to combine create vs update into the same schema and distinguish between what is required then we are overcomplicating things. There are already great client-side JSON schema form builders and validators, we shouldn't unnecessarily re-invent things in that space. For that reason alone I would agree that sticking with method-level schema makes the most sense.

@mattmcneeney can you clarify what UI benefits you are thinking of that you get with resource-level schema instead.  I could maybe see wanting to show parameters that were set during create as read-only fields during an update, which would require knowing which fields were there during create.

mattmcneeney:
@jwforres There were two benefits that we could think of adding schemes at the resource level. The first is what you have mentioned; being able to 'easily' build a form that showed parameters that were set at create time but weren't updatable. The second was that other resource level fields could be added to the schema (i.e. dashboard URL).

However, our investigations showed that combining the schemes didn't make building forms easier for UI teams and made validation much more difficult. Since we want many brokers to add parameter validation (and platforms too) we believe we need to make life easy for broker authors, which we think is method attached schemes.

mattmcneeney:
@vaikas-google did you hear of any further feedback on this?

spec.md
```diff
@@ -147,10 +147,19 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
...
+| [schemas](#schemas-object) | object | Schema definitions for service instances and bindings for the plan. |
 
 
 \* Fields with an asterisk are REQUIRED.
 
+##### Schemas Object
+
+| Response field | Type | Description |
+| --- | --- | --- |
+| service-instance | JSON schema object | The schema definition for a service instance. Schema definitions MUST be valid [JSON Schema draft v4](http://json-schema.org/). Broker authors SHOULD explicitly mention the JSON schema draft 4 version, as this will prevent breaking changes when in the future the OSB spec starts supporting future versions of JSON schema with low readability impact. Input parameters that can be used to create a service instance can be configured using a top-level `parameters` object. A service that does not support any input parameters for service instances SHOULD provide the following schema: `{ "$schema": "http://json-schema.org/draft-04/schema#", "additionalProperties": false }`. |
```
pmorie: The field names used here and in the example below are inconsistent.  The table uses the names `service-instance` and `service-binding`, while the examples use `service_instances` and `service_bindings`.  The field names used in the example are closer to the naming conventions used elsewhere in the spec.  Additionally, my understanding of this proposal was that we would have separate schemas for create and update of instances.  Did I understand incorrectly?

mattmcneeney: Hi @pmorie 

The code on the branch right now is the outcome of what we discussed in the working group face-to-face (ie. **resource-attached schemas**). I should probably have put it on a different branch as `method-attached-schemas` isn't the clearest of branch names!

I didn't want to change the branch back to the original proposal before the changes made at the face-to-face until everyone had had a chance to look at our feedback and compare.

The one amendment I _do_ think we should make to the original proposal is to change `service_instances` to `service_instance`, and `service_bindings` to `service_binding`. So the `schemas` object would look like:
```json
"schemas": {
  "service_instance": {
    "create": {},
    "update": {}
  },
  "service_binding": {
    "create": {}
  }
}
```
To me that makes more sense, as we are using schemas to describe how to create/update a singular service instance or binding. (But if people disagree, I don't really mind.)

Let me update this branch now with the **method-attached schemas** to reflect how we think the spec change should look.

mattmcneeney:
Alright @pmorie - branch updated. Hopefully this is closer to what you were expecting to see?

For anyone wanting to see the alternative that we discussed in the face to face, you can find the resource level schemas proposal on [this branch](https://github.com/mattmcneeney/servicebroker/tree/resource-attached-schemas).

spec.md
```diff
@@ -147,10 +147,40 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
...
+| [service_instance](#service-instances-object) | object | The schema definitions for all actions for a service instance. |
+| [service_binding](#service-bindings-object) | object | The schema definitions for all actions for a service binding. Used only if the service is bindable. |
+
+
+##### Service Instances Object
+
+| Response field | Type | Description |
+| --- | --- | --- |
+| [create](#input-parameters-object) | object | The schema definitions for the input parameters. |
+| [update](#input-parameters-object) | object | The schema definitions for the input parameters. |
```
pmorie: Since `update` is optional, if update is not supplied, does it imply that the same parameters are accepted for update as for create?  We should probably add some specificity around this.

spec.md
```diff
@@ -147,10 +147,40 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
...
+| Response field | Type | Description |
+| --- | --- | --- |
+| [create](#input-parameters-object) | object | The schema definitions for the input parameters. |
+
+
+##### Input Parameters Object
+
+| Response field | Type | Description |
+| --- | --- | --- |
+| parameters | JSON schema object | The schema definitions for the input parameters. Schema definitions MUST be valid [JSON Schema draft v4](http://json-schema.org/). It is recommended for broker authors explicitly mention the JSON schema draft 4 version, as this will prevent breaking changes when in the future the OSB spec starts supporting future versions of JSON schema with low readability impact. Each input parameter is expressed as a property within a JSON object. A service supporting no input parameters SHOULD provide the following schema: `{ "$schema": "http://json-schema.org/draft-04/schema#", "additionalProperties": false }`. |
```
pmorie: Is the platform required to validate the JSON schema?

spec.md
mattmcneeney: I don't think that is implied; to me this was clear enough without specifically calling it out. Happy to add something if needed, just not sure where it should go?

spec.md
```diff
@@ -147,10 +147,40 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
...
+| [schemas](#schema-object) | object | Schema definitions for service instances and bindings for the plan. |
 
 
 \* Fields with an asterisk are REQUIRED.
 
+##### Schema Object
+
+| Response field | Type | Description |
+| --- | --- | --- |
+| [service_instance](#service-instances-object) | object | The schema definitions for all actions for a service instance. |
```
pmorie: We should make the name of the anchor (`service-instances-object`) agree with the name of the field - which would be `Service Instance Object`.

spec.md
mattmcneeney: It's up to the platforms. In CF world, the cloud controller _probably will_ perform validation, as will client tools that are using this (web UIs and CLIs). There don't seem to many times in the spec where we specify what platforms should/must/may do.

spec.md
mattmcneeney: So the field itself would be called `service-instance-object`?

spec.md
pmorie: I think this is actually the root question: does the absense of a create/update schema imply that no parameters are accepted, or just indicate that they are not documented?

mattmcneeney: It implies they aren't documented, otherwise we'd be breaking backwards compatibility. No schema = we don't know what will be accepted = today's world.

If no parameters are allowed, a schema like this should be exposed by service brokers that have implemented this:
```json
{
    "type": "object",
    "additionalProperties": false
}
```

spec.md
```diff
@@ -202,6 +232,48 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
           "5 TB storage",
           "40 concurrent connections"
         ]
+      },
+      "schemas": {
+        "service_instances": {
```
jim-minter: If it's now called service_instance, then this needs to be changed

spec.md
```diff
@@ -202,6 +232,48 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
...
+              "properties": {
+                "billing-account": {
+                  "description": "Billing account number used to charge use of shared fake server.",
+                  "type": "string"
+                }
+              }
+            }
+          }
+        },
+        "service_bindings": {
```
jim-minter: If it's now called service_binding, then this needs to be changed

spec.md
mattmcneeney: Thanks @jim-minter good catch - changing now

spec.md
mattmcneeney: Thanks @jim-minter good catch - changing now

spec.md
```diff
@@ -147,10 +147,40 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
 | metadata | JSON object | An opaque object of metadata for a service plan. Controller treats this as a blob. Note that there are [conventions](https://docs.cloudfoundry.org/services/catalog-metadata.html) in existing brokers and controllers for fields that aid in the display of catalog data. |
 | free | boolean | When false, instances of this plan have a cost. The default is true. |
 | bindable | boolean | Specifies whether instances of the service plan can be bound to applications. This field is OPTIONAL. If specified, this takes precedence over the `bindable` attribute of the service. If not specified, the default is derived from the service. |
+| [schemas](#schema-object) | object | Schema definitions for service instances and bindings for the plan. |
```
jmrodri: Is schemas supposed to be an array? or a single object?

mattmcneeney: @jmrodri it's supposed to be an object. Plans can only have one schema for each action.

spec.md
```diff
@@ -202,6 +232,48 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
...
+            }
+          }
+        },
+        "service_binding": {
+          "create": {
+            "parameters": {
+              "$schema": "http://json-schema.org/draft-04/schema#",
+              "type": "object",
+              "properties": {
+                "billing-account": {
```
duglin: This repetition of properties is something that will strike people as odd and I'd prefer if we dealt with it head-on. In particular, I think we should add a sentence someplace that makes it clear that brokers need to duplicate the properties on ALL methods and not try to take short-cuts like only putting them on "create" and assume people will carry that info over to the other methods.  Without this some people might take undocumented short-cuts and that will lead to interop issues.

Alternatively, we could provide a mechanism to optimize this and avoid the duplication but I suspect it might get messy to do so.

spec.md
jmrodri: @mattmcneeney thanks

spec.md
jmrodri: @mattmcneeney @pmorie are we limited to `create` and `update`? Or is this just for now with possibility of adding `delete` in the future?

duglin: I don't think we want to head down the path of DELETE having a body - while spec-legal, some httpd servers don't support it.

spec.md
mattmcneeney: With the way it looks in the spec, I don't agree that developers would get confused about where schemas should be provided. We haven't heard any feedback from service teams in the CF community that the proposal is confusing in this regard.

We could change the wording in the tables if you think that would make things clearer though? Any suggestions?

mattmcneeney:
I'd like to bring this up on today's call. We've discussed the proposal in depth with both service broker authors and client tooling authors (Web UIs and CLIs) in the CF community and all parties are happy with the proposal as it stands.

I want to make sure other communities have discussed this and will be happy to implement this before/when it becomes part of the official spec.

duglin:
travis issue is real

spec.md
shalako: This appears to be resolved.

spec.md
shalako: This appears to be resolved.

spec.md
```diff
@@ -187,10 +187,40 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
...
 
 
 \* Fields with an asterisk are REQUIRED.
 
+##### Schema Object
+
+| Response field | Type | Description |
+| --- | --- | --- |
+| [service_instance](#service-instances-object) | object | The schema definitions for all actions for a service instance. |
+| [service_binding](#service-bindings-object) | object | The schema definitions for all actions for a service binding. Used only if the service is bindable. |
```
shalako: Suggestion for description:

"The schema definitions for a service binding. Used only if the service is bindable."

spec.md
shalako: I suggested updated to the descriptions for the service instance and service binding schema objects that should help clarify this.

spec.md
mattmcneeney: Thanks @shalako - update inbound

duglin:
See: https://github.com/kubernetes-incubator/service-catalog/issues/937
There's an issue w.r.t. ordering of parameters and whether a platform must retain the order of them or not. Ideally we should probably put something in the schema that specifies the order since by default these are just unordered maps.

mattmcneeney:
@duglin Are we able to handle ordering in a different issue? JSON schema does not support ordering today.

mattmcneeney:
Updated the branch to contain the following sentence by the `parameters` definition: `Note that UIs MAY not abide by the ordering of input parameters within the schema definition.`

mattmcneeney:
I have an update on this proposed feature.

Support for service instance schemas has now been merged into the Cloud Controller in Cloud Foundry (see [this PR](https://github.com/cloudfoundry/cloud_controller_ng/pull/865)) and should be in an official CF release soon. We have also implemented the spec change outlined in this PR for the Spring Cloud Services service broker.

The implementation of schemas for service instances in both the platform and a service broker has raised no concerns. We are in the process of adding support to this for service bindings, and have a high confidence that no issues will be found. I'd therefore propose that we begin the review process for this PR and look to get it merged into the specification.

Let me know if you have any queries or concerns.

cc @Samze @avade @zrob

mattmcneeney:
The implementation for service bindings has also been [merged](https://github.com/cloudfoundry/cloud_controller_ng/pull/866). No issues were found during the implementation.

I believe this is now ready for reviewing!

shalako:
LGTM

mattmcneeney:
JSON UI schema issue that discusses a UI-focussed extension that we could potentially use as a meta schema: https://github.com/json-schema-org/json-schema-spec/issues/67

jmrodri:
LGTM we added this to our ansible service broker.

spec.md
```diff
@@ -210,10 +210,40 @@ how platforms might expose these values to their users.
...
+| --- | --- | --- |
+| [create](#input-parameters-object) | object | The schema definition for creating a service binding. |
+
+
+##### Input Parameters Object
+
+| Response field | Type | Description |
+| --- | --- | --- |
+| parameters | JSON schema object | The schema definition for the input parameters. Schema definitions MUST be valid [JSON Schema draft v4](http://json-schema.org/). It is RECOMMENDED that broker authors explicitly mention the JSON schema draft 4 version, as this will prevent breaking changes when in the future this spec starts supporting future versions of JSON schema with low readability impact. Each input parameter is expressed as a property within a JSON object. Note that UIs MAY not abide by the ordering of input parameters within the schema definition. A service supporting no input parameters SHOULD provide the following schema: `{ "$schema": "http://json-schema.org/draft-04/schema#", "type": "object", "additionalProperties": false }`. |
+
```
skaegi: draft 6 is the current version that is being pushed at http://json-schema.org and we might want to in particular figure out what it would mean if I used "$schema": "http://json-schema.org/draft-06/schema#" instead of v4.

We might also consider starting our support at v6 instead of v4

spec.md
skaegi: One mechanism available in JSON Schema to eliminate repetitions is $ref.

skaegi: Thinking about this more we might want to flatten JSON References after incorporating the definitions into the catalog.

spec.md
```diff
@@ -242,7 +242,7 @@ how platforms might expose these values to their users.
...
+| parameters | JSON schema object | The schema definition for the input parameters. Platforms MUST support at least [JSON Schema draft v4](http://json-schema.org/), but SHOULD be prepared to support multiple versions. Schema definitions MUST be [JSON Schema draft v4](http://json-schema.org/) or a later version, and it is RECOMMENDED that broker authors explicitly mention the JSON schema version they are using. Each input parameter is expressed as a property within a JSON object. Note that UIs MAY not abide by the ordering of input parameters within the schema definition. A service supporting no input parameters SHOULD provide the following schema: `{ "$schema": "http://json-schema.org/draft-04/schema#", "type": "object", "additionalProperties": false }`. |
```
duglin: If no version is specified we should state what the default is or explicitly say that its undefined and up to the platform.

skaegi: I'm fine with how you talk about version support for the Platform. I think we might in the future want to update to call out v6 but looks good for now.

As per Doug's comment add something like... "If no schema version is provided by the broker author, platforms MUST minimally interpret definitions using [JSON Schema draft v4] and if unsuccessful MAY try using later JSON Schema versions."

I would altogether remove the last two statements.
1. We never could depend on ordering in JSON Schema or the more general ordering of JSON properties to mean anything and so feel it is unnecessary. 
2. A statement like "additionalProperties": false is ill advised for backwards compatibility of services that might in the future add properties. It of course makes sense for a service that will never accept parameters but I again would not call it out explicitly and leave it to the author to write their schemas as they see fit.

mattmcneeney:
Alright, if everyone is happy, I think we can move this into review!
@duglin @shalako @pmorie @angarg12 @fmui @vaikas-google @avade

profile.md
```diff
@@ -228,6 +228,25 @@ For example:
...
+### JSON Schema Restrictions
+
+Platforms MAY enforce restrictions on the JSON schemas service brokers can use
+to describe the input parameters they accept (see [spec](spec.md#schema-object).
+Each section below will describe any restrictions a Platform places on input
+parameter schemas.
+
+#### Cloud Foundry
+
+The following restrictions are enforced within a Cloud Foundry deployment:
```
duglin: hmm, these are requirements on brokers right? So, are we saying that a broker that wants to be used by both Kube and CF may need to have different schema?  I guess I'm asking whether these CF requirements are really CF specific or if they should be true for all platforms/brokers?

profile.md
```diff
@@ -228,6 +228,25 @@ For example:
...
+
+- The `parameters` object MUST have a top-level field as follows:
+  ```
+  "type": "object"
+  ```
+- The `$schema` field MUST be included at the top level of any provided schema.
```
duglin: Didn't we talk about $schema being optional but then letting the platform pick a default? Or am I mixing up conversations?

spec.md
```diff
@@ -228,10 +228,40 @@ how platforms might expose these values to their users.
...
+| Response field | Type | Description |
+| --- | --- | --- |
+| [create](#input-parameters-object) | object | The schema definition for creating a service binding. |
+
+
+##### Input Parameters Object
+
+| Response field | Type | Description |
+| --- | --- | --- |
+| parameters | JSON schema object | The schema definition for the input parameters. Platforms MUST support at least [JSON Schema draft v4](http://json-schema.org/), but SHOULD be prepared to support multiple versions. If no schema version is provided by the broker author using `$schema`, platforms MUST minimally interpret definitions using [JSON Schema draft v4](http://json-schema.org/), and if unsuccessful, MAY try using later JSON Schema versions. Each input parameter is expressed as a property within a JSON object. |
```
duglin: I wonder if this statement about schema versions should be more globally stated and not hidden with this one field.

spec.md
```diff
@@ -557,7 +629,7 @@ It MUST be the ID a previously provisioned service instance.
 | context | object | Contextual data under which the service instance is created. |
 | service_id* | string | The ID of the service (from the catalog). MUST be globally unique. MUST be a non-empty string. |
 | plan_id | string | The ID of the plan (from the catalog) for which the service instance has been requested. MUST be unique within the service. If present, MUST be a non-empty string. If this field is not present in the request message, then the broker MUST NOT change the plan of the instance as a result of this request. |
-| parameters | JSON object | Configuration options for the service instance. An opaque object, controller treats this as a blob. Brokers SHOULD ensure that the client has provided valid configuration parameters and values for the operation. If this field is not present in the request message, then the broker MUST NOT change the parameters of the instance as a result of this request. |
+| parameters | JSON object | Configuration options for the service instance. Brokers SHOULD ensure that the client has provided valid configuration parameters and values for the operation. |
```
duglin: any reason you dropped the MUST NOT sentence?

profile.md
mattmcneeney: We did, but in CF we really want to know the exact version so we can tell our validation library which version to validate against. I put this in profile.md to save having this requirement in the core spec, though I expect most broker authors to add the `$schema` key normally anyway.

spec.md
mattmcneeney: Maybe once we start using schemas for more than just parameters?

profile.md
mattmcneeney: I was hoping they wouldn't be clashing requirements - just additive things like fields that should be expressed (rather than "Broker authors must not do ...").

profile.md
mattmcneeney: Obviously happy to move both of these to the core spec if everyone agrees with them

spec.md
mattmcneeney: Maybe, but it was so longer ago I can't remember the reason, so I'll put it back in!

profile.md
mattmcneeney: @duglin do you have any further thoughts on this?

avade:
Ok, it seems like everyone has had a chance to review and give feedback. 

I think this is now in a state that we can merge it into the spec and move forward with the 2.13 release.

LGTM

profile.md
duglin: I'd prefer for it to be consistent between platforms otherwise we'll have an interop issue, and if we do that then we might as well just put it in the core spec, no?

skaegi: I agree that we want to be consistent here and think we should consider making appropriate changes to the OSB spec...

After reflecting, the feedback I gave earlier around handling multiple spec versions by inference was ultimately not good. Leaving it up to the platform to infer the JSON Schema version is just not of sufficient value to special case in the spec; it only leads to unexpected behaviour.  So... IF an author would like their parameters to be validated according to JSON Schema the contents of "parameters" MUST be a valid JSON Schema document (with a valid "$schema" property value that the platform understands) or else be empty.

Beyond stating that the document must be a valid JSON Schema document I would not stipulate "type": "object" on the schema as this requirement will vary with the schema version. e.g. in v4 only "type": "object" is permitted but in v6 the meta-scheme permits "type": "boolean, object", and who knows what future versions might say here. I would leave this detail up to the schema spec and not in OSB.

profile.md
```diff
@@ -228,6 +228,25 @@ For example:
...
+- No external references are allowed in schemas.
+- Schemas MUST be no larger than 64kB.
```
skaegi: Is there some special reason why there is a 64kB restriction?

mattmcneeney:
We limit to 64kB for security and performance reasons (we measured how long
it took to register a broker with many services with many plans with many
schemes). In reality, we don't think schemas will close to 64kB in size.

I agree with your comments Simon. In my opinion, the important points the
spec needs to detail are:
1. Platforms may want to support multiple versions of schema to support as
many brokers as possible.
2. Platforms need to know what version of JSON schema a broker is using to
correctly validate it (and possibly build a UI using it).
3. Platforms may want to offer different restrictions on schemas.

I think we agree on 1 and 2, in which case I believe the PR as it stands
(updated today) is sufficient. Brokers must have the $schema field and use
at least v4 (the minimum platforms may support).

The restrictions are interesting. I agree about type object, so we'd be
happy to remove this as it is required in Schema v4 anyway. If the group
agrees on the size limit and no external references (again for security
reasons in our case) then I'd be happy to move these into the core spec and
have no platform profiles section for schemas.

What do you think?

skaegi:
No external references is reasonable. I'm less comfortable with the size cap but admit that a 64kB schema seems unlikely at this point - most I have used in our platform are less than 1kB. 

Could we say that in a less restrictive way? If a schema is larger than 64kB a platform MAY refuse to register the broker?

mattmcneeney:
Sure, I don't mind that, but if we want to make it easy for people to
create platform agnostic brokers I think we should definite very clear
rules that all platforms abide by. We can always change or remove
restrictions in the future if a need arises.

angarg12:
This is a big feature, so instead of trying to overengineer it beforehand I would prefer to get it out there and see how different brokers use it. If further requirements/restrictions arise in the future we can address them later.

So I'm in favor of moving the restrictions out of the profiles to the spec.

spec.md
```diff
@@ -228,10 +228,40 @@ how platforms might expose these values to their users.
...
+| Response field | Type | Description |
+| --- | --- | --- |
+| [create](#input-parameters-object) | object | The schema definition for creating a service binding. |
+
+
+##### Input Parameters Object
+
+| Response field | Type | Description |
+| --- | --- | --- |
+| parameters | JSON schema object | The schema definition for the input parameters. Platforms MUST support at least [JSON Schema draft v4](http://json-schema.org/), but SHOULD be prepared to support multiple versions. Broker auhtors MUST provide the version of JSON Schema they are using. Schemas MUST NOT contain any external references and MUST be no larger than 64kB. Each input parameter is expressed as a property within a JSON object. |
```
skaegi: typo -- auhtors instead of authors

mattmcneeney: Thanks - fixed!

mattmcneeney:
Looking for some LGTMs today! (Thanks for the feedback so far)
@angarg12 @duglin @pmorie @skaegi @shalako @vaikas-google @fmui @avade

angarg12:
LGTM

spec.md
```diff
@@ -228,10 +228,40 @@ how platforms might expose these values to their users.
 | metadata | JSON object | An opaque object of metadata for a service plan. Controller treats this as a blob. Note that there are [conventions](profile.md#service-metadata) in existing brokers and controllers for fields that aid in the display of catalog data. |
 | free | boolean | When false, service instances of this plan have a cost. The default is true. |
 | bindable | boolean | Specifies whether service instances of the service plan can be bound to applications. This field is OPTIONAL. If specified, this takes precedence over the `bindable` attribute of the service. If not specified, the default is derived from the service. |
+| [schemas](#schema-object) | object | Schema definitions for service instances and bindings for the plan. |
```
duglin: all of these `object`s, should they be `JSON object` to be consistent with what I above? .e.g `metadata`

spec.md
```diff
@@ -228,10 +228,40 @@ how platforms might expose these values to their users.
...
+| Response field | Type | Description |
+| --- | --- | --- |
+| [create](#input-parameters-object) | object | The schema definition for creating a service binding. |
+
+
+##### Input Parameters Object
+
+| Response field | Type | Description |
+| --- | --- | --- |
+| parameters | JSON schema object | The schema definition for the input parameters. Platforms MUST support at least [JSON Schema draft v4](http://json-schema.org/), but SHOULD be prepared to support multiple versions. Broker authors MUST provide the version of JSON Schema they are using. Schemas MUST NOT contain any external references and MUST be no larger than 64kB. Each input parameter is expressed as a property within a JSON object. |
```
duglin: I'm ok with the requirements of this field. I'm just wondering whether it would read better if everything after the first sentence were moved to a non-tabular format. Meaning, put it in a paragraph/section right after this table so that we have more space to be more explicit. For example:

+++
While the presence of schema in the Plan is OPTIONAL, if it is present then the following applies:
- Platforms MUST support at least [JSON Schema draft v4](http://json-schema.org).
- Platforms SHOULD be prepared to support other versions of JSON Schema well.
- The `$schema` keyword MUST be present in the schema with the version of the JSON Schema being used.
- Schemas MUST NOT contain any external references (e.g. use the `$ref` keyword).
- Schemas MUST NOT be larger than 64kB.
- Each input parameter MUST be expressed as a child of the `properties` field in the schema.

+++

Just allows us to be a little more verbose and show the exact json entities we're talking about.

duglin: Did you want to mention that `type` is required? Or is that obvious?

spec.md
```diff
@@ -284,6 +314,48 @@ how platforms might expose these values to their users.
...
+      },
+      "schemas": {
+        "service_instance": {
+          "create": {
+            "parameters": {
+              "$schema": "http://json-schema.org/draft-04/schema#",
+              "type": "object",
+              "properties": {
+                "billing-account": {
+                  "description": "Billing account number used to charge use of shared fake server.",
```
duglin: Do we want (or need) to place any requirements on the set of fields used to describe the properties?  I'm guessing no but wanted to ask to be sure.

spec.md
mattmcneeney: yes it should be - thanks!

mattmcneeney: Wait - I'm not actually sure it should be. Elsewhere in the spec when the object definition is defined in the spec, we just use `object`, and use `JSON object` where the structure is not defined.

duglin: hmm I think you're right.  That distinction was totally lost on me and I wonder whether its one worth keeping - after all, `object` is always a `JSON object`, right?

duglin: But that can be a different PR, if we want to change it

spec.md
mattmcneeney: We're leaving that up to the spec, since this changes in v6 schema.
Moving the text out of the table - check back in 2 mins!

spec.md
mattmcneeney: Yeah, I don't think there are any cases now in the spec where object != JSON object, but we can handle that elsewhere :)

duglin:
LGTM

angarg12:
LGTM

skaegi:
LGTM

avade:
LGTM

vaikas-google:
LGTM

mattmcneeney:
Woohoo! Thanks all!

PR #166, opened 2017-04-11
Use RFC2119 keywords - first of many

duglin:
This is the first of many PRs to migrate the spec to use RFC2119 keywords
to tighten up the language so its clear when things are required vs just
optional/guidance.

In this pass:
- add a title header
- added a "Notational Conventions"  section - to bring in RFC2119
- move "Changes" section down one section since I think "API Overview"
  is more important. We should really add an "Abstract" section at some
  point too.
- for each RFC2119 keyword already in the spec I either made it all CAPS or
  replaced it with a more appropate word. Note that sometimes we were
  using the word w/o meaning for it to be normative, like in examples.
  Sometimes I did swap things like "should" for "MUST" when it seemed
  like that was the real intent. But please do look over each one to make
  sure I didn't go too far.
- Related to the previous one, I tried to avoid tighening the language
  in the spec in most cases because I was hoping to limit this to more of
  a syntax change PR and we can leave PRs that might impact semantics for
  follow-on PRs.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1,8 +1,11 @@
-# Table of Contents
+# Open Service Broker API
+
+## Table of Contents
+  - [API Overview](#api-overview)
+  - [Notational Conventions](#notational-conventions)
   - [Changes](#changes)
     - [Change Policy](#change-policy)
     - [Changes Since v2.10](#changes-since-v210)
-  - [API Overview](#api-overview)
   - [API Version Header](#api-version-header)
   - [Authentication](#authentication)
   - [Catalog Management](#catalog-management)
@@ -21,22 +24,7 @@
   - [Broker Errors](#broker-errors)
   - [Orphans](#orphans)
 
-## Changes
-
-### Change Policy 
-
-* Existing endpoints and fields will not be removed or renamed.
-* New optional endpoints, or new HTTP methods for existing endpoints, may be
-added to enable support for new features.
-* New fields may be added to existing request/response messages.
-These fields must be optional and should be ignored by clients and servers
-that do not understand them.
-
-## Changes Since v2.10
-
-* Add <tt>bindable</tt> field to [Plan Object](#PObject) to allow services to have both bindable and non-bindable plans.
-
-## API Overview 
+## API Overview
 
 The Service Broker API defines an HTTP interface between the services marketplace of a platform and service brokers.
 
@@ -44,34 +32,55 @@ The service broker is the component of the service that implements the Service B
 
 In general, provisioning reserves a resource on a service; we call this reserved resource a service instance. What a service instance represents can vary by service. Examples include a single database on a multi-tenant server, a dedicated cluster, or an account on a web application.
 
-What a binding represents may also vary by service. In general creation of a binding either generates credentials necessary for accessing the resource or provides the service instance with information for a configuration change.
+What a binding represents MAY also vary by service. In general creation of a binding either generates credentials necessary for accessing the resource or provides the service instance with information for a configuration change.
 
-A platform marketplace may expose services from one or many service brokers, and an individual service broker may support one or many platform marketplaces using different URL prefixes and credentials.
+A platform marketplace MAY expose services from one or many service brokers, and an individual service broker MAY support one or many platform marketplaces using different URL prefixes and credentials.
+
+## Notational Conventions
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL" in this document are to
+be interpreted as described in [RFC 2119]( https://tools.ietf.org/html/rfc2119).
+
+## Changes
+
+### Change Policy
+
+* Existing endpoints and fields MUST NOT be removed or renamed.
+* New OPTIONAL endpoints, or new HTTP methods for existing endpoints, MAY be
+added to enable support for new features.
+* New fields MAY be added to existing request/response messages.
+These fields MUST be OPTIONAL and SHOULD be ignored by clients and servers
+that do not understand them.
+
+### Changes Since v2.10
+
+* Add <tt>bindable</tt> field to [Plan Object](#PObject) to allow services to have both bindable and non-bindable plans.
 
 ## API Version Header
 
-Requests from the platform to the service broker must contain a header that declares the version number of the Service Broker API that the marketplace will use:
+Requests from the platform to the service broker MUST contain a header that declares the version number of the Service Broker API that the marketplace will use:
 
 `X-Broker-Api-Version: 2.11`
 
 The version numbers are in the format `MAJOR.MINOR`, using semantic versioning such that 2.10 comes before 2.11.
 
-This header allows brokers to reject requests from marketplaces for versions they do not support. While minor API revisions will always be additive, it is possible that brokers depend on a feature from a newer version of the API that is supported by the platform. In this scenario the broker may reject the request with `412 Precondition Failed` and provide a message that informs the operator of the required API version.
+This header allows brokers to reject requests from marketplaces for versions they do not support. While minor API revisions will always be additive, it is possible that brokers depend on a feature from a newer version of the API that is supported by the platform. In this scenario the broker MAY reject the request with `412 Precondition Failed` and provide a message that informs the operator of the API version that is to be used instead.
 
 ## Authentication
 
-The marketplace must authenticate with the service broker using HTTP
-basic authentication (the `Authorization:` header) on every request. The broker is responsible for validating the username and password and returning a `401 Unauthorized` message if credentials are invalid. It is recommended that brokers support secure communication from platform marketplaces over TLS.
+The marketplace MUST authenticate with the service broker using HTTP
+basic authentication (the `Authorization:` header) on every request. The broker is responsible for validating the username and password and returning a `401 Unauthorized` message if credentials are invalid. It is RECOMMENDED that brokers support secure communication from platform marketplaces over TLS.
 
-## Catalog Management 
+## Catalog Management
 
-The first endpoint that a broker must implement is the service catalog.
+The first endpoint that a platform will interact with on the broker is the service catalog.
 
 The platform marketplace fetches this endpoint from all brokers in order to present an aggregated user-facing catalog.
 
 Warnings for broker authors:
 
-- Be cautious removing services and plans from their catalogs, as platform marketplaces may have provisioned service instances of these plans. Consider your deprecation strategy.
+- Be cautious removing services and plans from their catalogs, as platform marketplaces might have provisioned service instances of these plans. Consider your deprecation strategy.
 - Do not change the ids of services and plans. This action is likely to be evaluated by a platform marketplace as a removal of one plan and addition of another. See above warning about removal of plans.
 
 The following sections describe catalog requests and responses in the Service Broker API.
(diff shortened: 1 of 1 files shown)


angarg12:
Good for a first pass, LGTM

duglin:
I don't think this PR changed that aspect of the spec, but I do agree we should reformat it - I'd like to do that in a follow-on PR though. This one is just about adding the RFC2119 keywords.

duglin:
any chance of some reviews on this?

spec.md
```diff
@@ -282,9 +291,9 @@ The request provides these query string parameters as useful hints for brokers.
 |---|---|---|
 | service_id  |  string | ID of the service from the catalog.  |
 | plan_id  | string  | ID of the plan from the catalog.  |
-| operation  |  string | A broker-provided identifier for the operation. When a value for <code>operation</code> is included with asynchronous responses for [Provision](#provisioning), [Update](#updating_service_instance), and [Deprovision](#deprovisioning) requests, the broker client should provide the same value using this query parameter as a URL-encoded string.  |
+| operation  |  string | A broker-provided identifier for the operation. When a value for <code>operation</code> is included with asynchronous responses for [Provision](#provisioning), [Update](#updating_service_instance), and [Deprovision](#deprovisioning) requests, the broker client SHOULD provide the same value using this query parameter as a URL-encoded string.  |
```
pmorie: I appreciate that we are adding these clarifications now.

We hit an issue this week with a broker we are developing at RH with this.  Is this really a SHOULD or is it a MUST?  If it is a SHOULD, I'm not sure what the point is of having this is at all.

spec.md
```diff
@@ -296,22 +305,22 @@ $ curl http://username:password@broker-url/v2/service_instances/:instance_id/las
 | Status Code  |  Description |
 |---|---|
 | 200 OK |  The expected response body is below. |
-|  410 GONE | Appropriate only for asynchronous delete operations. The platform should consider this response a success and remove the resource from its database. The expected response body is <code>{}</code>. Returning this while the platform is polling for create or update operations should be interpreted as an invalid response and the platform should continue polling.  |
+|  410 GONE | Appropriate only for asynchronous delete operations. The platform can consider this response a success and remove the resource from its database. The expected response body is <code>{}</code>. Returning this while the platform is polling for create or update operations SHOULD be interpreted as an invalid response and the platform SHOULD continue polling.  |
```
pmorie: `can consider` seems a little ambiguous, given that the sentence has technical statements about what the platform should do.  Is this a SHOULD or a MUST or out of scope?

spec.md
```diff
@@ -320,32 +329,32 @@ For success responses, the following fields are valid.
 }
 </pre>
 
-### Polling Interval and Duration    
+### Polling Interval and Duration
 
-The frequency and maximum duration of polling may vary by platform client. If a platform has a max polling duration and this limit is reached, the platform will cease polling and the operation state will be considered `failed`.
+The frequency and maximum duration of polling MAY vary by platform client. If a platform has a max polling duration and this limit is reached, the platform will cease polling and the operation state will be considered `failed`.
```
pmorie: I think "will be considered `failed`" could use some consideration - I think 'will' here is probably equivalent to a MUST.

spec.md
```diff
@@ -320,32 +329,32 @@ For success responses, the following fields are valid.
...
+When the broker receives a provision request from the platform, it MUST take whatever action is necessary to create a new resource. What provisioning represents varies by service and plan, although there are several common use cases. For a MySQL service, provisioning could result in an empty dedicated database server running on its own VM or an empty schema on a shared database server. For non-data services, provisioning could just mean an account on an multi-tenant SaaS application.
 
 ### Request ###
 
 ##### Route #####
 `PUT /v2/service_instances/:instance_id`
 
-The `:instance_id` of a service instance is provided by the platform. This ID will be used for future requests (bind and deprovision), so the broker must use it to correlate the resource it creates.
+The `:instance_id` of a service instance is provided by the platform. This ID will be used for future requests (bind and deprovision), so the broker will use it to correlate the resource it creates.
```
pmorie: I think 'will' here is a MUST.  Brokers have to maintain state of what the instance id means.

spec.md
```diff
@@ -379,26 +388,26 @@ $ curl http://username:password@broker-url/v2/service_instances/:instance_id -d
 | Status Code  | Description  |
 |---|---|
 | 201 Created  | Service instance has been provisioned. The expected response body is below.  |
-| 200 OK  |  May be returned if the service instance already exists and the requested parameters are identical to the existing service instance. The expected response body is below. |
+| 200 OK  |  MAY be returned if the service instance already exists and the requested parameters are identical to the existing service instance. The expected response body is below. |
```
pmorie: I think this should be a MUST.

spec.md
```diff
@@ -502,44 +511,44 @@ For success responses, a broker may return the following field. Others will be i
...
+Some platforms MAY support proxying of application requests to service instances. In this case the platform needs to know where to send application requests; to facilitate this, the broker MUST return a `route_service_url` in the response for a request to create a binding. Not all services of this type expect to receive requests proxied by the platform; some services will have been configured out-of-band to intermediate requests to applications. In this case, the broker will not return `route_service_url` in response to the create binding request. By sending `bind-resource` as described above, the platform enables dynamic configuration of a service instance already in the application request path for the route, requiring no change in the platform routing tier.
```
pmorie: > MUST return a `route_service_url`

I think this is a MAY.

spec.md
```diff
@@ -502,44 +511,44 @@ For success responses, a broker may return the following field. Others will be i
...
+There are a class of services that provide network storage to applications via volume mounts in the application container. A service broker MUST return data needed for this configuration with `volume_mount` in response to the request to create a binding.
 
-The `requires` field in the [Catalog](#catalog-mgmt) endpoint enables a platform marketplace to validate a response for create binding that includes a `volume_mounts`. Platform marketplaces should consider a broker's response invalid if it includes a `volume_mounts` and `"requires":["volume_mount"]` is not present in the [Catalog](#catalog-mgmt) endpoint.
+The `requires` field in the [Catalog](#catalog-mgmt) endpoint enables a platform marketplace to validate a response for create binding that includes a `volume_mounts`. Platform marketplaces can consider a broker's response invalid if it includes a `volume_mounts` and `"requires":["volume_mount"]` is not present in the [Catalog](#catalog-mgmt) endpoint.
```
pmorie: We should make this clarification to the other special service types:

Platform marketplaces can consider a broker's response invalid if it includes a `volume_mounts` and `"requires":["volume_mount"]` is not present in the [Catalog](#catalog-mgmt) endpoint.

spec.md
```diff
@@ -589,24 +598,24 @@ $ curl http://username:password@broker-url/v2/service_instances/:instance_id/ser
...
-| 200 OK  |  May be returned if the binding already exists and the requested parameters are identical to the existing binding. The expected response body is below. |
-| 409 Conflict  | Should be returned if the requested binding already exists. The expected response body is <code>{}</code>, though the description field can be used to return a user-facing error message, as described in [Broker Errors](#broker-errors).  |
-| 422 Unprocessable Entity  | Should be returned if the broker requires that <code>app_guid</code> be included in the request body. The expected response body is: <code>{ "error": "RequiresApp", "description": "This service supports generation of credentials through binding an application only." }</code>  |
+| 200 OK  |  MUST be returned if the binding already exists and the requested parameters are identical to the existing binding. The expected response body is below. |
+| 409 Conflict  | MUST be returned if the requested binding already exists. The expected response body is <code>{}</code>, though the description field can be used to return a user-facing error message, as described in [Broker Errors](#broker-errors).  |
```
pmorie: >  can be used

MAY be used?

spec.md
```diff
@@ -589,24 +598,24 @@ $ curl http://username:password@broker-url/v2/service_instances/:instance_id/ser
...
-| volume\_mounts  | array-of-objects  | An array of configuration for mounting volumes. <code>"requires":["volume_mount"]</code> must be declared in the [Catalog](#catalog-mgmt) endpoint or the platform should consider the response invalid.  |
+| credentials  | object  |  A free-form hash of credentials that can be used by applications or users to access the service. |
+| syslog\_drain_url  | string  | A URL to which logs MUST be streamed. <code>"requires":["syslog\_drain"]</code> MUST be declared in the [Catalog](#catalog-mgmt) endpoint or the platform can consider the response invalid.  |
+| route\_service_url  | string  |  A URL to which the platform MUST proxy requests for the address sent with <code>bind\_resource.route</code> in the request body. <code>"requires":["route\_forwarding"]</code> MUST be declared in the [Catalog](#catalog-mgmt) endpoint or the platform can consider the response invalid. |
```
pmorie: >  can consider the response invalid.

Is there really a use-case for not specifying the route forwarding requirement but providing a `route_service_url`, or is this a MUST?

spec.md
duglin: I agree that this should be a MUST but I'd prefer to tighten the language in a follow-on PR because I'd like to get this boilerplate PR in there first - and then leave the bike-shedding over each RFC2119 keyword swap for more focused PRs.

spec.md
duglin: Yep - I think in this case MUST is more appropriate than "can consider" - will change.

spec.md
duglin: yup - I think the intent here was for those two "will"s to be MUSTs.

spec.md
duglin: I think I swapped this from a "must" to a "will" because how the correlation is done is an impl detail. Clearly the intent is for people to use the ID, that's why it's there, but if someone uses some other mechanism behind the scenes does that mean they're violating the spec?  For example, what if each client is only allowed one service - then the ID becomes meaningless and they may not use it. If someone looks at an impl's code and notices that, but everything works as expected, are they non-compliant?   The other reason I thought MUST may not be right is that I wasn't sure whether this is testable w/o looking at the code.  But I'd like to hear what others think.

spec.md
duglin: I want to agree but I'm wondering if a 202 is a valid response as well in the cases where an async request is being replayed.  Sending back a 200 could impl the resource is ready, but in fact it might still be under construction.  Perhaps we could revisit this in a follow-on PR. I think it may just need a bit more language around it.

spec.md
duglin: I made this a MUST to indicate that **if** this feature is supported then a "route_service_url" MUST be returned.
Perhaps a rewording like this:
```
Some platforms MAY support proxying of application requests to service instances. If a broker supports this
feature then it MUST return a `route_service_url` in the response for a request to create a binding. Not all
services of this type expect to receive requests proxied by the platform; some services will have been
configured out-of-band to intermediate requests to applications. In this case, the broker MUST NOT 
return `route-service_url` in the response to create binding requests.  By sending... (same as before)
```

spec.md
duglin: opened https://github.com/openservicebrokerapi/servicebroker/issues/169 to track it in a follow-on PR

spec.md
duglin: changed

spec.md
duglin: I think you're correct, but swapping this for a MUST here feels awkward to me - and this isn't the only spot. I'm not sure what saying "MUST consider the response invalid" means. Does it mean they MUST reject the msg? Can they do some kind of error recovery?  I would prefer to talk about what MUST or MUST NOT be in the message and then add a section that talks about what the platform is to do when it gets a non-compliant message in general (assuming that covers all of the cases).  But we can do that in a follow-on PR.

But, in the meantime... I'll make this a MUST since that's the  biggest change I'd like to make this PR on this topic and I think its a step in the right direction.

duglin:
Pushed new version to address @pmorie's comments.

spec.md
duglin: Didn't add this change yet - wanted to get feedback first.

spec.md
jmrodri: +1 to being a MUST.

duglin:
Any chance of getting some final LGTMs on this one?

angarg12:
I repeat my LGTM

avade:
LGTM

spec.md
vaikas-google: Well, there's two things here at play so let's make sure we figure out which one we're talking about:
 1. Broker SHOULD/MUST return a last_operation in the response to async provisioning request
 2. Platform SHOULD/MUST provide the same value

@pmorie  you and I were discussing #1 because it would simplify writing the state management in the controller. So just wanted to make sure we're talking about the right 'this' :)

spec.md
vaikas-google: +1 on tightening up the language, my reading is that a 202 should be returned to the async request if one is already ongoing.

vaikas-google:
LGTM

duglin:
LGTM

vaikas-google:
LGTM

angarg12:
LGTM

avade:
LGTM

avade:
Leaving for @duglin to merge when ready.

PR #168, opened 2017-04-12
Add anchors to fix ToC

bentarnoff:
This PR should make the linked ToC at the top of http://docs.cloudfoundry.org/services/api.html work. Also, I wasn't sure how deep you wanted the ToC to go, but you can add additional anchor links to any headers that you want reflected in the ToC.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -21,9 +21,9 @@
   - [Broker Errors](#broker-errors)
   - [Orphans](#orphans)
 
-## Changes
+## <a id='changes'></a>Changes
 
-### Change Policy 
+### <a id='change-policy'></a>Change Policy 
 
 * Existing endpoints and fields will not be removed or renamed.
 * New optional endpoints, or new HTTP methods for existing endpoints, may be
@@ -32,11 +32,11 @@ added to enable support for new features.
 These fields must be optional and should be ignored by clients and servers
 that do not understand them.
 
-## Changes Since v2.10
+## <a id='since-v2.10'></a>Changes Since v2.10
 
 * Add <tt>bindable</tt> field to [Plan Object](#PObject) to allow services to have both bindable and non-bindable plans.
 
-## API Overview 
+## <a id='api-overview'></a>API Overview 
 
 The Service Broker API defines an HTTP interface between the services marketplace of a platform and service brokers.
 
@@ -48,7 +48,7 @@ What a binding represents may also vary by service. In general creation of a bin
 
 A platform marketplace may expose services from one or many service brokers, and an individual service broker may support one or many platform marketplaces using different URL prefixes and credentials.
 
-## API Version Header
+## <a id='version-header'></A>API Version Header
 
 Requests from the platform to the service broker must contain a header that declares the version number of the Service Broker API that the marketplace will use:
 
@@ -58,12 +58,12 @@ The version numbers are in the format `MAJOR.MINOR`, using semantic versioning s
 
 This header allows brokers to reject requests from marketplaces for versions they do not support. While minor API revisions will always be additive, it is possible that brokers depend on a feature from a newer version of the API that is supported by the platform. In this scenario the broker may reject the request with `412 Precondition Failed` and provide a message that informs the operator of the required API version.
 
-## Authentication
+## <a id='authentication'></a>Authentication
 
 The marketplace must authenticate with the service broker using HTTP
 basic authentication (the `Authorization:` header) on every request. The broker is responsible for validating the username and password and returning a `401 Unauthorized` message if credentials are invalid. It is recommended that brokers support secure communication from platform marketplaces over TLS.
 
-## Catalog Management 
+## <a id='catalog-mgmt'></a>Catalog Management 
 
 The first endpoint that a broker must implement is the service catalog.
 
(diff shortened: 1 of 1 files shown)


duglin:
what is this fixing? We should use the normal anchors produced by the MD processing, but the TOC seems to work just fine for me.

duglin:
ok I think I see what's going on. The TOC on the CF version of the spec isn't acting right, modifying our spec.md file isn't the right fix, the fix should be on the CF building/tweaking of their version of the file.

bentarnoff:
Our file only consists of a partial that calls your `spec.md`, so there isn't much we can change on our end. https://github.com/cloudfoundry/docs-services/blob/master/api.html.md.erb

duglin:
Can you elaborate on what's wrong?  It looks like your anchors match what github produces so why is this needed?

MHBauer:
I know very little html, but the generated CF docs page doesn't have the identifiers for the headers of each section. So there is a link in the TOC, but where it goes doesn't "exist" as far as html is concerned.

Github renders the markdown with the identifier. Is there a way to edit the rendering code to add identifying elements to all of the headers based on the header text?

MHBauer:
example, View source of each page
GH TOC
```
<li><a href="#changes">Changes</a>
```
GH Header
```
<h2><a id="user-content-changes" class="anchor" href="#changes" aria-hidden="true"><svg aria-hidden="true" class="octicon octicon-link" height="16" version="1.1" viewBox="0 0 16 16" width="16"><path fill-rule="evenodd" d="M4 9h1v1H4c-1.5 0-3-1.69-3-3.5S2.55 3 4 3h4c1.45 0 3 1.69 3 3.5 0 1.41-.91 2.72-2 3.25V8.59c.58-.45 1-1.27 1-2.09C10 5.22 8.98 4 8 4H4c-.98 0-2 1.22-2 2.5S3 9 4 9zm9-3h-1v1h1c1 0 2 1.22 2 2.5S13.98 12 13 12H9c-.98 0-2-1.22-2-2.5 0-.83.42-1.64 1-2.09V6.25c-1.09.53-2 1.84-2 3.25C6 11.31 7.55 13 9 13h4c1.45 0 3-1.69 3-3.5S14.5 6 13 6z"></path></svg></a><a id="user-content-changes"></a>Changes</h2>
```

CF TOC
```
<li><a href="#changes">Changes</a>
```
CF Header
```
<h2>Changes</h2>
```

I don't think manually checking in the source changes is a good idea unless we have a tool to do it automatically. Maintenance burden to do it by hand.

bentarnoff:
Yeah, the problem is that https://docs.cloudfoundry.org/services/api.html doesn't have the page ToC that the other topics have (for instance, look at http://docs.cloudfoundry.org/services/managing-service-brokers.html). The tool we use to publish our documentation, Bookbinder, generates the ToC from `<a id='foo'></a>` tags, which is why we added them in this PR to try to make the ToC work correctly. Unfortunately we don't presently have the engineering resources to modify Bookbinder, so we're not able to generate the ToC any other way, as far as I know.

Looping in @shalako, who originated this request.

duglin:
If I provide you with a tool that took an md file as input and output another md file with the "a href" tags inserted, could you use that as input into your process instead of the default spec.md ?

bentarnoff:
@duglin Sure. Maybe the tool could output the file with the `<a id=''>` tags to the `servicebroker` repo, and then we could consume it as a partial instead of `spec.md`?

MHBauer:
Ah, another instance of "github doesn't generate a TOC" bites again.

isaacs/github#215

duglin:
@bentarnoff here's a bash script that will read a md file from stdin and insert the anchors for each header:
```
#!/bin/bash

IFS=''
while read -r line ; do
    if echo "$line" | grep "^[[:space:]]*#\+[[:space:]]\+[^#]\+$" > /dev/null 2>&1 ; then
        # get everything before the heading text
        lead=$(echo "${line}" | sed "s/\([[:space:]]*##*[[:space:]]\+\).*/\1/")
        # get the heading text
        rest=$(echo "${line}" | sed "s/\([[:space:]]*##*[[:space:]]\+\)\(.*\)/\2/")
        # convert the heading into a markdown anchor
        # - remove trailing spaces
        # - convert upper to lower case
        # - remove all punct
        # - convert all spaces to -'s
        ref=$(echo -n "${rest}" | \
                sed "s/[[:space:]]*$//g" | \
                tr "[:upper:]" "[:lower:]" | \
                sed "s/[[:punct:]]*//g" | \
                tr "[:space:]" "-")
        # now add the anchor to the line and print it
        echo "${lead} <a id='${ref}'></a> ${rest}"
    else
        # not a header so just print it
        echo "${line}"
    fi
done
```

bentarnoff:
@duglin Thanks for this. How is that script going to be used? Since we consume `spec.md` as a partial, and we don't own `spec.md`, you'd have to run the script on the file on your end.

duglin:
@bentarnoff I'm not sure what you mean by "partial". To use it just send the spec thru it via stdin/out - e.g.
```
mdAddAnchors < spec.md > _spec.md
```
Since you guys, I think, want the spec to start with an underscope anyway, just putting this line in your CI process might do the trick.

avade:
@bentarnoff what Doug is suggesting is that in the concourse pipeline you run this script against our spec.md before pushing the files.

I have setup a call to discuss.

angarg12:
Can we confirm if this issue has been solved, and close this PR?

avade:
We will try to resolve the CF docs issue in `2.12` of the spec, but for now I have created a branch at `f12b5e3` that will be considered the `stable` for `2.11`. CF docs will point at that branch.

https://github.com/openservicebrokerapi/servicebroker/commits/v2.11

avade:
Closing PR and will follow up with CF docs folks for 2.12.

PR #178, opened 2017-05-03
Clean-up some text around our status codes

duglin:
Follow-on to this comment: https://github.com/openservicebrokerapi/servicebroker/pull/166/files#r113097103

Changes:
- Add MUSTs/MAYs on each status code in the tables
- Add a 404 as the more appropriate response after a DELETE, not 410
- Allow for a 200 in spots where it was missing
- Clarify things, like when a conflict happens
- lots of typographical changes (spaces, wrong casing, etc...) - nits

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -99,7 +99,7 @@ $ curl -H "X-Broker-API-Version: 2.11" http://username:password@broker-url/v2/ca
 
 | Status Code | Description |
 | --- | --- |
-| 200 OK | The expected response body is below. |
+| 200 OK | MUST be returned upon successful processing of this request. The expected response body is below. |
 
 #### Body - Schema of Service Objects
 
@@ -124,7 +124,7 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
 | requires | array-of-strings | A list of permissions that the user would have to give the service, if they provision it. The only permissions currently supported are `syslog_drain`, `route_forwarding` and `volume_mount`. |
 | bindable* | boolean | Specifies whether instances of the service can be bound to applications. This specifies the default for all plans of this service. Plans can override this field (see [Plan Object](#plan-object)). |
 | metadata | JSON object | An opaque object of metadata for a service offering. Controller treats this as a blob. Note that there are [conventions](https://docs.cloudfoundry.org/services/catalog-metadata.html) in existing brokers and controllers for fields that aid in the display of catalog data. |
-| [dashboard_client](#dashboard-client-object) | object | Contains the data necessary to activate the Dashboard SSO feature for this service |
+| [dashboard_client](#dashboard-client-object) | object | Contains the data necessary to activate the Dashboard SSO feature for this service. |
 | plan_updateable | boolean | Whether the service supports upgrade/downgrade for some plans. Please note that the misspelling of the attribute `plan_updatable` to `plan_updateable` was done by mistake. We have opted to keep that misspelling instead of fixing it and thus breaking backward compatibility. |
 | [plans*](#plan-object) | array-of-objects | A list of plans for this service, schema is defined below. MUST contain at least one plan. |
 
@@ -304,7 +304,7 @@ $ curl http://username:password@broker-url/v2/service_instances/:instance_id/las
 
 | Status Code | Description |
 | --- | --- |
-| 200 OK | The expected response body is below. |
+| 200 OK | MUST be returned upon successful processing of this request. The expected response body is below. |
 | 410 Gone | Appropriate only for asynchronous delete operations. The platform MUST consider this response a success and remove the resource from its database. The expected response body is `{}`. Returning this while the platform is polling for create or update operations SHOULD be interpreted as an invalid response and the platform SHOULD continue polling. |
 
 Responses with any other status code SHOULD be interpreted as an error or invalid response. The platform SHOULD continue polling until the broker returns a valid response or the [maximum polling duration](#polling-interval-and-duration) is reached. Brokers MAY use the `description` field to expose user-facing error messages about the operation state; for more info see [Broker Errors](#broker-errors).
@@ -388,13 +388,12 @@ $ curl http://username:password@broker-url/v2/service_instances/:instance_id?acc
 ```
 
 ### Response
-
 | Status Code | Description |
-| --- | --- |
-| 201 Created | Service instance has been provisioned. The expected response body is below. |
-| 200 OK | MAY be returned if the service instance already exists and the requested parameters are identical to the existing service instance. The expected response body is below. |
-| 202 Accepted | Service instance provisioning is in progress. This triggers the platform marketplace to poll the [Last Operation](#polling-last-operation) endpoint for operation status. |
-| 409 Conflict | MUST be returned if a service instance with the same id already exists but with different attributes. The expected response body is `{}`. |
+|---|---|
+| 200 OK | MUST be returned if the service instance already exists, is fully provisioned, and the requested parameters are identical to the existing service instance. The expected response body is below. |
+| 201 Created | MUST be returned if the service instance was provisioned as a result of this request. The expected response body is below. |
+| 202 Accepted | MUST be returned if the service instance provisioning is in progress. This triggers the platform marketplace to poll the [Service Instance Last Operation Endpoint](#polling-last-operation) for operation status. Note that a re-sent `PUT` request MUST return a `202 Accepted`, not a `200 OK`, if the service instance is not yet fully provisioned. |
+| 409 Conflict | MUST be returned if a service instance with the same id already exists but with different attributes. The expected response body is `{}`, though the description field MAY be used to return a user-facing error message, as described in [Broker Errors](#broker-errors). |
 | 422 Unprocessable Entity | MUST be returned if the broker only supports asynchronous provisioning for the requested plan and the request did not include `?accepts_incomplete=true`. The expected response body is: `{ "error": "AsyncRequired", "description": "This service plan requires client support for asynchronous service operations." }`, as described below. |
 
 Responses with any other status code will be interpreted as a failure. Brokers can include a user-facing message in the `description` field; for details see [Broker Errors](#broker-errors).
@@ -493,8 +492,8 @@ $ curl http://username:password@broker-url/v2/service_instances/:instance_id?acc
 
 | Status Code | Description |
 | --- | --- |
-| 200 OK | The requested changes have been applied. The expected response body is `{}`. |
-| 202 Accepted | Service instance update is in progress. This triggers the platform marketplace to poll the [Last Operation](#polling-last-operation) endpoint for operation status. |
+| 200 OK | MUST be returned if the request's changes have been applied. The expected response body is `{}`. |
+| 202 Accepted | MUST be returned if the service instance update is in progress. This triggers the platform marketplace to poll the [Last Operation](#polling-last-operation) for operation status. Note that a re-sent `PATCH` request MUST return a `202 Accepted`, not a `200 OK`, if the requested update has not yet completed. |
 | 422 Unprocessable entity | MUST be returned if the requested change is not supported or if the request cannot currently be fulfilled due to the state of the instance (e.g. instance utilization is over the quota of the requested plan). Brokers SHOULD include a user-facing message in the body; for details see [Broker Errors](#broker-errors). Additionally, a `422 Unprocessable Entity` can also be returned if the broker only supports asynchronous update for the requested plan and the request did not include `?accepts_incomplete=true`; in this case the expected response body is: `{ "error": "AsyncRequired", "description": "This service plan requires client support for asynchronous service operations." }` |
 
 Responses with any other status code will be interpreted as a failure. Brokers can include a user-facing message in the `description` field; for details see [Broker Errors](#broker-errors).
@@ -604,9 +603,9 @@ $ curl http://username:password@broker-url/v2/service_instances/:instance_id/ser
 
 | Status Code | Description |
 | --- | --- |
-| 201 Created | Binding has been created. The expected response body is below. |
 | 200 OK | MUST be returned if the binding already exists and the requested parameters are identical to the existing binding. The expected response body is below. |
-| 409 Conflict | MUST be returned if the requested binding already exists. The expected response body is `{}`, though the description field MAY be used to return a user-facing error message, as described in [Broker Errors](#broker-errors). |
+| 201 Created | MUST be returned if the binding was created as a result of this request. The expected response body is below. |
+| 409 Conflict | MUST be returned if a service binding with the same id, for the same service instance, already exists but with different parameters. The expected response body is `{}`, though the description field MAY be used to return a user-facing error message, as described in [Broker Errors](#broker-errors). |
 | 422 Unprocessable Entity | MUST be returned if the broker requires that `app_guid` be included in the request body. The expected response body is: `{ "error": "RequiresApp", "description": "This service supports generation of credentials through binding an application only." }` |
 
 Responses with any other status code will be interpreted as a failure and an unbind request will be sent to the broker to prevent an orphan being created on the broker. Brokers can include a user-facing message in the `description` field; for details see [Broker Errors](#broker-errors).
@@ -671,7 +670,7 @@ $ curl 'http://username:password@broker-url/v2/service_instances/:instance_id/
 
 | Status Code | Description |
 | --- | --- |
-| 200 OK | Binding was deleted. The expected response body is `{}`. |
+| 200 OK | MUST be returned if the binding was deleted as a result of this request. The expected response body is `{}`. |
 | 410 Gone | MUST be returned if the binding does not exist. The expected response body is `{}`. |
 
 Responses with any other status code will be interpreted as a failure and the binding will remain in the marketplace database. Brokers can include a user-facing message in the `description` field; for details see [Broker Errors](#broker-errors).
@@ -718,8 +717,8 @@ $ curl 'http://username:password@broker-url/v2/service_instances/:instance_id?ac
 
 | Status Code | Description |
 | --- | --- |
-| 200 OK | Service instance was deleted. The expected response body is `{}`. |
-| 202 Accepted | Service instance deletion is in progress. This triggers the marketplace to poll the [Last Operation](#polling-last-operation) endpoint for operation status. |
+| 200 OK | MUST be returned if the service instance was deleted as a result of this request. The expected response body is `{}`. |
+| 202 Accepted | MUST be returned if the service instance deletion is in progress. This triggers the marketplace to poll the [Service Instance Last Operation Endpoint](#polling-last-operation) for operation status. Note that a re-sent `DELETE` request MUST return a `202 Accepted`, not a `200 OK`, if the delete request has not completed yet. |
 | 410 Gone | MUST be returned if the service instance does not exist. The expected response body is `{}`. |
 | 422 Unprocessable Entity | MUST be returned if the broker only supports asynchronous deprovisioning for the requested plan and the request did not include `?accepts_incomplete=true`. The expected response body is: `{ "error": "AsyncRequired", "description": "This service plan requires client support for asynchronous service operations." }`, as described below. |
 


spec.md
```diff
@@ -345,19 +346,19 @@ When the broker receives a provision request from the platform, it MUST take wha
 The `:instance_id` of a service instance is provided by the platform. This ID will be used for future requests (bind and deprovision), so the broker will use it to correlate the resource it creates.
 
 #### Parameters ####
-| Parameter name  | Type  |  Description |
+| Parameter name | Type | Description |
 |---|---|---|
-|  accepts_incomplete | boolean  | A value of true indicates that the marketplace and its clients support asynchronous broker operations. If this parameter is not included in the request, and the broker can only provision an instance of the requested plan asynchronously, the broker should reject the request with a 422 as described below.  |
+| accepts_incomplete | boolean | A value of true indicates that the marketplace and its clients support asynchronous broker operations. If this parameter is not included in the request, and the broker can only provision an instance of the requested plan asynchronously, the broker should reject the request with a 422 as described below. |
```
mattmcneeney: Should this be "the broker should reject the request with a `422 Unprocessable Entity` as described below" for consistency?

spec.md
```diff
@@ -390,13 +391,13 @@ $ curl http://username:password@broker-url/v2/service_instances/:instance_id?acc
...
-| 422 Unprocessable Entity  | MUST be returned if the broker only supports asynchronous provisioning for the requested plan and the request did not include <code>?accepts_incomplete=true</code>. The expected response body is: <code>{ "error": "AsyncRequired", "description": "This service plan requires client support for asynchronous service operations." }</code>, as described below.  |
+| 200 OK | MUST be returned if the service instance already exists, is fully provisioned, and the requested parameters are identical to the existing service instance. The expected response body is below. |
+| 201 Created | MUST be returned if the service instance was provisioned as a result of this request. The expected response body is below. |
+| 202 Accepted | MUST be returned if the service instance provisioning is in progress. This triggers the platform marketplace to poll the [Service Instance Last Operation Endpoint](#polling) for operation status. Note that a resent `PUT` request MUST return a 202, not a 200, if the service instance is not yet fully provisioned. |
```
mattmcneeney: Same comment as above.

spec.md
```diff
@@ -493,11 +494,11 @@ $ curl http://username:password@broker-url/v2/service_instances/:instance_id?acc
...
+| 200 OK | MUST be returned if the request's changes have been applied. The expected response body is <code>{}</code>. |
+| 202 Accepted | MUST be returned if the service instance update is in progress. This triggers the platform marketplace to poll the [Service Instance Last Operation Endpoint](#polling) for operation status. Note that a resent `PATCH` request MUST return a 202, not a 200, if the requested update has not yet completed. |
+| 422 Unprocessable entity | MUST be returned if the requested change is not supported or if the request cannot currently be fulfilled due to the state of the instance (e.g. instance utilization is over the quota of the requested plan). Broker SHOULD include a user-facing message in the body; for details see [Broker Errors](#broker-errors). Additionally, a 422 can also be returned if the broker only supports asynchronous update for the requested plan and the request did not include <code>?accepts_incomplete=true</code>; in this case the expected response body is: <code>{ "error": "AsyncRequired", "description": "This service plan requires client support for asynchronous service operations." }</code> |
```
mattmcneeney: Looks like there are lots of places. Maybe it's OK to ignore!

spec.md
```diff
@@ -604,12 +605,12 @@ $ curl http://username:password@broker-url/v2/service_instances/:instance_id/ser
...
-| 409 Conflict  | MUST be returned if the requested binding already exists. The expected response body is <code>{}</code>, though the description field MAY be used to return a user-facing error message, as described in [Broker Errors](#broker-errors).  |
-| 422 Unprocessable Entity  | MUST be returned if the broker requires that <code>app_guid</code> be included in the request body. The expected response body is: <code>{ "error": "RequiresApp", "description": "This service supports generation of credentials through binding an application only." }</code>  |
+| 200 OK | MUST be returned if the binding already exists and the requested parameters are identical to the existing binding. The expected response body is below. |
+| 201 Created | MUST be returned if the binding was created as a result of this request. The expected response body is below. |
+| 409 Conflict | MUST be returned if a service binding with the same id, for the same service instance, already exists but with different parameters. The expected response body is <code>{}</code>, though the description field MAY be used to return a user-facing error message, as described in [Broker Errors](#broker-errors). |
```
mattmcneeney: Regarding the `409 Conflict`, should we say "but with different parameters"? I'd imagine that the conflict error should be returned if you try to create a binding for the same service instance with an existing binding ID regardless of parameters?

spec.md
duglin: probably. As you mentioned below there are quite a few spots where we just use the number and not the text, but let me see if I can find them all.

spec.md
duglin: done

spec.md
duglin: done

spec.md
duglin: My assumption is that all of the APIs in the spec are meant to be idempotent, in which case returning an error on the 2nd try violates that. Also, if the first response message is lost, which is why they're resending the request, if we return a 409 how will the client recover?  I think that goes back to the idempotent nature of things.  This also relates to another discussion we're having about how long, or if at all, a broker needs to hold on to the credentials for a binding. I know some people have said that brokers might be stateless but that seems a bit odd in the case of credentials since they need to hold on to them so they can verify that apps are using the right ones, right? So, if they have to hang on to them to so auth checking then is it really a burden to ask them to use them to replay a binding response message?

But, if my assumption is wrong about them being idempotent then we may need to look at all of the APIs to make sure we have a clear recovery plan for each when messages are lost.

avade: Interested to hear what @shalako thinks on this point.

I think the spec was intending to give flexibility to brokers that may not be able to replay requests for some reason.

mattmcneeney: @duglin I spoke to a few of the CF service teams about holding onto credentials. It seems that most brokers are stateless and whatever downstream component they talk to give them credentials when they create a new binding (synchronously) which they immediately return to the platform. There isn't currently a way for brokers to ask whatever is downstream to give them the credentials again for a particular binding. I'm not saying this is the right way, but I think that's what today's world looks like.

spec.md
```diff
@@ -287,11 +287,11 @@ Valid values for `state` are `in progress`, `succeeded`, and `failed`. The platf
...
 
-|  Query-String Field | Type  | Description  |
+| Query-String Field | Type | Description |
 |---|---|---|
-| service_id  |  string | ID of the service from the catalog.  |
-| plan_id  | string  | ID of the plan from the catalog.  |
-| operation  |  string | A broker-provided identifier for the operation. When a value for <code>operation</code> is included with asynchronous responses for [Provision](#provisioning), [Update](#updating_service_instance), and [Deprovision](#deprovisioning) requests, the broker client SHOULD provide the same value using this query parameter as a URL-encoded string.  |
+| service_id | string | ID of the service from the catalog. |
+| plan_id | string | ID of the plan from the catalog. |
+| operation | string | A broker-provided identifier for the operation. When a value for <code>operation</code> is included with asynchronous responses for [Provision](#provisioning), [Update](#updating_service_instance), and [Deprovision](#deprovisioning) requests, the broker client SHOULD provide the same value using this query parameter as a URL-encoded string. |
```
vaikas-google: SHOULD for operation (the broker client SHOULD) is being changed to MUST in another PR, I assume it's going to get changed here as a merge?

duglin: yup - doing the rebase now and its picking up the MUST

duglin:
rebased

LGTM

angarg12:
LGTM

avade:
LGTM

angarg12:
Looks like it needs rebasing.

duglin:
rebased

angarg12:
lgtm

duglin:
ping @avade @vaikas-google @pmorie for some LGTMs

avade:
LGTM

duglin:
rebased

angarg12:
Looks like your script is already giving results :)

`tools/../spec.md: Can't find section '#route_services' in tools/../spec.md`
`tools/../spec.md - 443: Use 'SHOULD'`

duglin:
nice! :-)
fixed - I messed up on the rebase.  Would be so much easier with shorter lines.

angarg12:
LGTM

mattmcneeney:
@duglin we can cut all of the lines to 80 characters and add a travis check in the #192 PR if you'd like.

duglin:
I would love that :-)  I'll add comments there (later today) of other syntax changes that we should make, or that I've made in other PRs (that I can pull out).

Question is... do we merge #192 before or after the other current PRs?  :-)  While it might cause a ton of rebases (especially for me), it would make it easier to review those other PRs if #192 went in first.

mattmcneeney:
@duglin Whilst #192 would mean other PRs need a rebase, the rebases should typically be fairly small as long as the PR is only touching a small bit of the spec. Some fixes that people (ie. me) had bundled into other PRs when they spotted them have been pulled into #192 so it should ensure the other PRs are cleaner too.

I'm probably biased (I've been rebasing the PR on each change which isn't the best job) but I think it's sensible enough for merge it soon. But happy with whatever everyone else would like.

duglin:
rebase

duglin:
ready for review

spec.md
```diff
@@ -304,8 +304,10 @@ $ curl http://username:password@broker-url/v2/service_instances/:instance_id/las
 
 | Status Code | Description |
 | --- | --- |
-| 200 OK | The expected response body is below. |
+| 200 OK | MUST be returned upon successful processing of this request. The expected response body is below. |
+| 404 Not Found | MUST be returned if the service instance does not exist. |
 | 410 Gone | Appropriate only for asynchronous delete operations. The platform MUST consider this response a success and remove the resource from its database. The expected response body is `{}`. Returning this while the platform is polling for create or update operations SHOULD be interpreted as an invalid response and the platform SHOULD continue polling. |
+| 410 Gone | MAY be returned if the broker knows about the service instance but it had been previously deleted. Appropriate only for asynchronous delete operations. The platform MUST consider this response a success and remove the resource from its database. The expected response body is `{}`. Returning this while the platform is polling for create or update operations SHOULD be interpreted as an invalid response and the platform SHOULD continue polling. While a `410 Gone` is allowed, the RECOMMENDED response is `404 Not Found`. |
```
angarg12: 410 Gone is duplicated.

duglin: fixed!

angarg12:
LGTM

spec.md
```diff
@@ -304,8 +304,9 @@ $ curl http://username:password@broker-url/v2/service_instances/:instance_id/las
 
 | Status Code | Description |
 | --- | --- |
-| 200 OK | The expected response body is below. |
-| 410 Gone | Appropriate only for asynchronous delete operations. The platform MUST consider this response a success and remove the resource from its database. The expected response body is `{}`. Returning this while the platform is polling for create or update operations SHOULD be interpreted as an invalid response and the platform SHOULD continue polling. |
+| 200 OK | MUST be returned upon successful processing of this request. The expected response body is below. |
+| 404 Not Found | MUST be returned if the service instance does not exist. |
+| 410 Gone | MAY be returned if the broker knows about the service instance but it had been previously deleted. Appropriate only for asynchronous delete operations. The platform MUST consider this response a success and remove the resource from its database. The expected response body is `{}`. Returning this while the platform is polling for create or update operations SHOULD be interpreted as an invalid response and the platform SHOULD continue polling. While a `410 Gone` is allowed, the RECOMMENDED response is `404 Not Found`. |
```
avade: @mattmcneeney / @zrob does CF already conform to this flow?

avade:
LGTM

duglin:
rebased again

vaikas-google:
LGTM

avade:
LGTM

angarg12:
LGTM

spec.md
shalako: For async delete operations, brokers must return a 410. For async create and update, brokers must return a 200. If the resource does not exist, brokers must return a 410.

A 404 is problematic because it may be temporarily returned by a intermediating component (e.g. the CF router). We would not want a platform to remove a resource from its database as a result of a 404 response, since we cannot guarantee that a 404 was returned by the broker itself.

spec.md
```diff
@@ -447,10 +447,10 @@ Not all permutations of plan changes are expected to be supported. For example,
...
 | parameters | JSON object | Configuration options for the service instance. An opaque object, controller treats this as a blob. Brokers SHOULD ensure that the client has provided valid configuration parameters and values for the operation. |
 | previous_values | object | Information about the instance prior to the update. |
-| previous\_values.service_id | string | ID of the service for the instance. |
-| previous\_values.plan_id | string | ID of the plan prior to the update. |
-| previous\_values.organization_id | string | ID of the organization specified for the instance. |
-| previous\_values.space_id | string | ID of the space specified for the instance. |
+| previous_values.service_id | string | ID of the service for the instance. |
+| previous_values.plan_id | string | ID of the plan prior to the update. |
+| previous_values.organization_id | string | ID of the organization specified for the instance. |
+| previous_values.space_id | string | ID of the space specified for the instance. |
```
shalako: I believe the escaped underscores were required to prevent the CF docs from styling characters between them as italic.

duglin: The escape char as been removed in other spots (I think by Matt) so this is just following that pattern. Notice like 449 doesn't have it. Or is there something special about these lines that makes it required here?  @mattmcneeney ?

spec.md
```diff
@@ -671,8 +671,9 @@ $ curl 'http://username:password@broker-url/v2/service_instances/:instance_id/
 
 | Status Code | Description |
 | --- | --- |
-| 200 OK | Binding was deleted. The expected response body is `{}`. |
-| 410 Gone | MUST be returned if the binding does not exist. The expected response body is `{}`. |
+| 200 OK | MUST be returned if the binding was deleted as a result of this request. The expected response body is `{}`. |
+| 404 Not Found | MUST be returned if the service binding does not exist. |
+| 410 Gone | MAY be returned if the binding is known to the service broker but had been previously deleted. The expected response body is `{}`. While a `410 Gone` is allowed, the RECOMMENDED response is `404 Not Found`. |
```
shalako: Since a 404 may be temporarily returned by an intermediating component, we wouldn't want the platform to execute a state change based on it. Only 200 and 410 should be supported responses.

spec.md
```diff
@@ -718,9 +719,10 @@ $ curl 'http://username:password@broker-url/v2/service_instances/:instance_id?ac
 
 | Status Code | Description |
 | --- | --- |
-| 200 OK | Service instance was deleted. The expected response body is `{}`. |
-| 202 Accepted | Service instance deletion is in progress. This triggers the marketplace to poll the [Last Operation](#polling-last-operation) endpoint for operation status. |
-| 410 Gone | MUST be returned if the service instance does not exist. The expected response body is `{}`. |
+| 200 OK | MUST be returned if the service instance was deleted as a result of this request. The expected response body is `{}`. |
+| 202 Accepted | MUST be returned if the service instance deletion is in progress. This triggers the marketplace to poll the [Service Instance Last Operation Endpoint](#polling-last-operation) for operation status. Note that a resent `DELETE` request MUST return a `202 Accepted`, not a `200 OK`, if the delete request has not completed yet. |
+| 404 Not Found | MUST be returned if the service instance does not exist. |
+| 410 Gone | MAY be returned if the service instance is known to the broker but had been previously deleted. The expected response body is `{}`. While a `410 Gone` is allowed, the RECOMMENDED response is `404 Not Found`. |
```
shalako: Same issues with 404. We should not introduce this as a supported response without further discussion.

spec.md
duglin: this feels like we're letting an impl detail bleed into the spec and, sort of, reinventing http response codes in the process.  Can we discuss this on the next call?

spec.md
shalako: If it was remove elsewhere, it may be breaking CF docs. Members of this group who are involved with the CF community must consider themselves custodians of the CF broker docs, and consider impact when making nitpick changes.

That said, it is possible that @mattmcneeney has discussed the topic with the CF docs team and confirmed that removal of the escape char no longer causes a change in styling.

shalako:
> @duglin we can cut all of the lines to 80 characters and add a travis check in the #192 PR if you'd like.

Please don't cut all lines. Soft wrap is your friend.

mattmcneeney:
I haven't spoken to the docs team, so maybe I shouldn't have removed the
underscores. I'll check this and re-escape them if necessary.

spec.md
```diff
@@ -387,16 +388,15 @@ $ curl http://username:password@broker-url/v2/service_instances/:instance_id?acc
 }' -X PUT -H "X-Broker-API-Version: 2.11" -H "Content-Type: application/json"
 ```
 
-### Response
```
zrob: was removing this heading accidental?

spec.md
zrob: agree with @shalako that this is a change to the spec, not appropriate for a PR titled clean-up text

perhaps a separate issue should be raised around what status codes should be supported.  as is, this change potentially invalidates existing broker and platform implementations

spec.md
```diff
@@ -493,8 +493,8 @@ $ curl http://username:password@broker-url/v2/service_instances/:instance_id?acc
 
 | Status Code | Description |
 | --- | --- |
-| 200 OK | The requested changes have been applied. The expected response body is `{}`. |
-| 202 Accepted | Service instance update is in progress. This triggers the platform marketplace to poll the [Last Operation](#polling-last-operation) endpoint for operation status. |
+| 200 OK | MUST be returned if the request's changes have been applied. The expected response body is `{}`. |
+| 202 Accepted | MUST be returned if the service instance update is in progress. This triggers the platform marketplace to poll the [Last Operation](#polling-last-operation) for operation status. Note that a resent `PATCH` request MUST return a `202 Accepted`, not a `200 OK`, if the requested update has not yet completed. |
```
zrob: resent -> re-sent

spec.md
```diff
@@ -387,16 +388,15 @@ $ curl http://username:password@broker-url/v2/service_instances/:instance_id?acc
...
 
-### Response
-
 | Status Code | Description |
-| --- | --- |
-| 201 Created | Service instance has been provisioned. The expected response body is below. |
-| 200 OK | MAY be returned if the service instance already exists and the requested parameters are identical to the existing service instance. The expected response body is below. |
-| 202 Accepted | Service instance provisioning is in progress. This triggers the platform marketplace to poll the [Last Operation](#polling-last-operation) endpoint for operation status. |
+|---|---|
+| 200 OK | MUST be returned if the service instance already exists, is fully provisioned, and the requested parameters are identical to the existing service instance. The expected response body is below. |
```
zrob: A change from MAY to MUST feels like a spec change rather than a text fix.  It looks like the intent is that 201 can be returned in the case the service instance already existed.  Are we concerned with enforcing a change here?

spec.md
duglin: geez, not sure how I did this - thanks

spec.md
duglin: I'll move this to a new issue since I think more discussions are needed.

spec.md
duglin: Since PUT is supposed to be idempotent, I was trying to distinguish between "already there", "completed as a result of this request" and "doing it asynchronously" as the platform may want to know what happened on this particular request.  IMO, the old wording implies what the new text says - just not as clearly.

spec.md
duglin: done - and in the other spots too

spec.md
duglin: reverting for now and we can discuss later - see previous comment on this

spec.md
duglin: reverting for now

spec.md
pmorie: I agree with @duglin on this matter.  MAY is also significantly less useful (ie, not useful at all) compared to MUST here.

spec.md
```diff
@@ -387,15 +387,15 @@ $ curl http://username:password@broker-url/v2/service_instances/:instance_id?acc
...
+| 201 Created | MUST be returned if the service instance was provisioned as a result of this request. The expected response body is below. |
+| 202 Accepted | MUST be returned if the service instance provisioning is in progress. This triggers the platform marketplace to poll the [Service Instance Last Operation Endpoint](#polling-last-operation) for operation status. Note that a re-sent `PUT` request MUST return a `202 Accepted`, not a `200 OK`, if the service instance is not yet fully provisioned. |
 | 409 Conflict | MUST be returned if a service instance with the same id already exists but with different attributes. The expected response body is `{}`. |
 | 422 Unprocessable Entity | MUST be returned if the broker only supports asynchronous provisioning for the requested plan and the request did not include `?accepts_incomplete=true`. The expected response body is: `{ "error": "AsyncRequired", "description": "This service plan requires client support for asynchronous service operations." }`, as described below. |
 
+
```
pmorie: spurious newline?

spec.md
```diff
@@ -387,15 +387,15 @@ $ curl http://username:password@broker-url/v2/service_instances/:instance_id?acc
...
-| 202 Accepted | Service instance provisioning is in progress. This triggers the platform marketplace to poll the [Last Operation](#polling-last-operation) endpoint for operation status. |
+|---|---|
+| 200 OK | MUST be returned if the service instance already exists, is fully provisioned, and the requested parameters are identical to the existing service instance. The expected response body is below. |
+| 201 Created | MUST be returned if the service instance was provisioned as a result of this request. The expected response body is below. |
+| 202 Accepted | MUST be returned if the service instance provisioning is in progress. This triggers the platform marketplace to poll the [Service Instance Last Operation Endpoint](#polling-last-operation) for operation status. Note that a re-sent `PUT` request MUST return a `202 Accepted`, not a `200 OK`, if the service instance is not yet fully provisioned. |
 | 409 Conflict | MUST be returned if a service instance with the same id already exists but with different attributes. The expected response body is `{}`. |
```
pmorie: This should have the same guidance as we have elsewhere for conflicts - that user-facing errors may be supplied

spec.md
avade: I agree that MUST clarifies things here, because what else would you return, as there is only one conditional that can trigger this case.

spec.md
duglin: removed

spec.md
duglin: done

shalako:
LGTM

avade:
LGTM

vaikas-google:
LGTM

zrob:
LGTM

Issue #169, opened 2017-04-25
Add clarity around special services


duglin:
See: https://github.com/openservicebrokerapi/servicebroker/pull/166/files#r113097798

---
We should make this clarification to the other special service types:

Platform marketplaces can consider a broker's response invalid if it includes a volume_mounts and "requires":["volume_mount"] is not present in the Catalog endpoint.
---

PR #179, opened 2017-05-03
Clarify when route_service_url is required

duglin:
Clarify when route_service_url is required
    
See: https://github.com/openservicebrokerapi/servicebroker/pull/166/files#r113097378
    
Closes: #169
    
Add some clarity around when a broker can return extra data in the bindings
    
I decided to reword this to be a requirement on the broker. Talking about
the platform didn't make any sense to me since we can't really say anything
normative about it from the POV.
    
While in there I also `s/broker client/platform/g` for consistency.

Signed-off-by: Doug Davis <dug@us.ibm.com>

PR #180, opened 2017-05-03
Add some clarity around when a broker can return extra data in the bindings

duglin:
Closes: #169

I decided to reword this to be a requirement on the broker. Talking about
the platform didn't make any sense to me since we can't really say anything
normative about it from the POV.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -534,9 +534,16 @@ Credentials are a set of information used by an application or a user to utilize
 
 #### Log Drain ####
 
-There are a class of service offerings that provide aggregation, indexing, and analysis of log data. To utilize these services an application that generates logs needs information for the location to which it will stream logs. If a broker represents one of these services, it MAY return a `syslog_drain_url` in the response for a request to create a service binding, to which logs MUST be streamed.
+There are a class of service offerings that provide aggregation, indexing, and
+analysis of log data. To utilize these services an application that generates
+logs needs information for the location to which it will stream logs. A
+create binding response from one of these services MUST include a
+`syslog_drain_url`. Applications MUST use this URL when sending logs to the
+service.
 
-The `requires` field in the [Catalog](#catalog-mgmt) endpoint enables a platform marketplace to validate a response for create binding that includes a `syslog_drain_url`. Platform marketplaces can consider a broker's response invalid if it includes a `syslog_drain_url` and `"requires":["syslog_drain"]` is not present in the [Catalog](#catalog-mgmt) endpoint.
+Brokers MUST NOT include a `syslog_drain_url` in a create binding response
+if it did not include a `"requires":["syslog_drain"]` property in its
+[Catalog](#catalog-mgmt) endpoint.
 
 #### Route Services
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -546,11 +553,19 @@ Some platforms MAY support proxying of application requests to service instances
 
 The `requires` field in the [Catalog](#catalog-mgmt) endpoint enables a platform marketplace to validate requests to create bindings. A platform MAY reject requests to create bindings when a broker has declared `"requires":["route_forwarding"]` for a service in the catalog endpoint.
 
+Brokers MUST NOT include a `route_service_url` in a create binding response
+if it did not include a `"requires":["route_forwarding"]` property in its
+[Catalog](#catalog-mgmt) endpoint.
+
```
angarg12: Doesn't this overlap with #179 ?

spec.md
```diff
@@ -534,9 +534,16 @@ Credentials are a set of information used by an application or a user to utilize
 
 #### Log Drain ####
 
-There are a class of service offerings that provide aggregation, indexing, and analysis of log data. To utilize these services an application that generates logs needs information for the location to which it will stream logs. If a broker represents one of these services, it MAY return a `syslog_drain_url` in the response for a request to create a service binding, to which logs MUST be streamed.
+There are a class of service offerings that provide aggregation, indexing, and
+analysis of log data. To utilize these services an application that generates
+logs needs information for the location to which it will stream logs. A
+create binding response from one of these services MUST include a
+`syslog_drain_url`. Applications MUST use this URL when sending logs to the
+service.
```
gberche-orange: > Applications MUST use this URL when sending logs to the service.

In the case of CF, it is the platform that sends logs to the service, not directly the application, see http://docs.cloudfoundry.org/devguide/deploy-apps/streaming-logs.html The wording might therefore be confusing to CF community.

spec.md
duglin: yea, I think I got myself confused when I tried to keep these two (very related) PR separate. I'm going to merge them into one.

duglin:
Moving this one's work to #179

spec.md
duglin: The old text to me was a bit ambiguous as to who is sending the logs, I'll reword it so that we don't imply who is doing the sending.  I assume its ok for the app to send logs there too, right?

gberche-orange: >  I assume its ok for the app to send logs there too, right?

That's a platform choice whether require hosted apps to embed syslog libraries and have networking access to reach syslog drain endpoint. CF takes this burden off the apps and expects apps to be writing to stdout/stderr. 

Up to now the OSB was not normative about describing the usage of the syslog drain url:
- who is writing the logs (app vs platform)
- the expected format the log, see https://github.com/cloudfoundry/loggregator/issues/134
- the expected content/meta-data from the log (e.g. stdout vs stderr, timestamp)

Clarifying this would certainly help making syslog drain services more portable across platforms, and might deserve its own issue.

PR #190, opened 2017-05-09
Fix typo in bind_resource

pmorie:
... it was mispelled `bind-resource` in one place.


angarg12:
LGTM

avade:
LGTM

duglin:
LGTM

We really need to fix it so the spec keeps like down to 80 chars, it'll be so much easier to spot diffs after that. :-)  Plus, if we do end up using swagger/yaml then it'll look nicer too.

avade:
@zrob @shalako @fmui @vaikas-google we need one more lgtm here

zrob:
LGTM

PR #194, opened 2017-05-10
Add clarity around app_guid

duglin:
Just explain why it might not be present in the message

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -564,8 +564,24 @@ the resource it creates.
 | service_id* | string | ID of the service from the catalog. MUST be a non-empty string. |
 | plan_id* | string | ID of the plan from the catalog. MUST be a non-empty string. |
 | app_guid | string | Deprecated in favor of `bind_resource.app_guid`. GUID of an application associated with the binding to be created. If present, MUST be a non-empty string. |
-| bind_resource | JSON object | A JSON object that contains data for platform resources associated with the binding to be created. Current valid values include `app_guid` for [credentials](#types-of-binding) and `route` for [route services](#route-services). |
+| bind_resource | JSON object | A JSON object that contains data for platform resources associated with the binding to be created. See [Bind Resource Object](#bind-resource-object) for more information. |
 | parameters | JSON object | Configuration options for the service binding. An opaque object, controller treats this as a blob. Brokers SHOULD ensure that the client has provided valid configuration parameters and values for the operation. |
+
+##### Bind Resource Object
+
+The `bind_resource` object contains platform specific information
+related to the context in which the service will be used.
+In some cases the platform might not be able to provide this information
+at the time of the binding request, therefore the `bind_resource` and
+its fields are OPTIONAL.
+
+Below are some common fields that MAY be used. Platforms MAY choose to
+add additional ones as needed.
+
+| Request Field | Type | Description |
+| --- | --- | --- |
+| app_guid | string | GUID of an application associated with the binding. For [credentials](#types-of-binding) bindings. |
+| route | string | URL of the application to be intermediated. For [route services](#route-services) bindings. |
 
 \* Fields with an asterisk are REQUIRED.
 


avade:
LGTM

angarg12:
LGTM

duglin:
rebased

spec.md
```diff
@@ -567,9 +567,22 @@ the resource it creates.
...
 | app_guid  | string  | Deprecated in favor of <code>bind\_resource.app\_guid</code>. GUID of an application associated with the binding to be created.  |
-| bind_resource  | JSON object  | A JSON object that contains data for platform resources associated with the binding to be created. Current valid values include <code>app\_guid</code> for [credentials](#types-of-binding) and <code>route</code> for [route services](#route-services).  |
+| bind_resource  | JSON object  | A JSON object that contains data for platform resources associated with the binding to be created. See [Bind Resource Object](#bind-resource-object) for more information. |
 | parameters | JSON object  |  Configuration options for the service binding. An opaque object, controller treats this as a blob. Brokers SHOULD ensure that the client has provided valid configuration parameters and values for the operation. |   |
 
+##### Bind Resource Object
+
+| Request Field | Type | Description |
+| --- | --- | --- |
+| app_guid | string | GUID of an application associated with the binding. For [credentials](#types-of-bindings) bindings. |
```
pmorie: typo: credentials bindings

spec.md
```diff
@@ -567,9 +567,22 @@ the resource it creates.
...
+| route | string | URL of the application to be intermediated. For [route services](#route-services) bindings. |
+
+Note: the `app_guid` is OPTIONAL because based on the platform's use of the
```
pmorie: Suggestion: change this verbiage:

> because based on the platform's use of the service binding there might not be an application known at this time.

to 'because the consuming application may not be known when the binding is created'.

pmorie:
Thinking a little further about this, i wonder if this is another area where need something similar to `context`

spec.md
duglin: what's the typo?  my eyes must be playing tricks on me  :-)

spec.md
duglin: done

duglin:
re: context - I've been thinking of something similar except I was thinking as it more like we just need to open up the `bind_resource` struct to allow for platform specific data to appear in there - thus turning it into a kind of "context"-like object.

angarg12:
This and some other PR's need a rebase, so I will wait for it for a new round of LGTM!

duglin:
rebased

angarg12:
LGTM

avade:
lgtm

mattmcneeney:
Needs another L G T M after rebase.
Previous approvers: @angarg12 @avade 
More needed from: @pmorie @shalako @vaikas-google

spec.md
```diff
@@ -564,9 +564,25 @@ the resource it creates.
 | service_id* | string | ID of the service from the catalog. MUST be a non-empty string. |
 | plan_id* | string | ID of the plan from the catalog. MUST be a non-empty string. |
 | app_guid | string | Deprecated in favor of `bind_resource.app_guid`. GUID of an application associated with the binding to be created. If present, MUST be a non-empty string. |
-| bind_resource | JSON object | A JSON object that contains data for platform resources associated with the binding to be created. Current valid values include `app_guid` for [credentials](#types-of-binding) and `route` for [route services](#route-services). |
+| bind_resource | JSON object | A JSON object that contains data for platform resources associated with the binding to be created. See [Bind Resource Object](#bind-resource-object) for more information. |
 | parameters | JSON object | Configuration options for the service binding. An opaque object, controller treats this as a blob. Brokers SHOULD ensure that the client has provided valid configuration parameters and values for the operation. |
 
+##### Bind Resource Object
+
+The `bind_resource` object contains platform specific information
```
pmorie: I sincerely appreciate that this is wrapped at 80 chars

spec.md
```diff
@@ -564,9 +564,25 @@ the resource it creates.
...
+In some cases the platform might not be able to provide this information
+at the time of the binding request, therefore the `bind_resource` and
+its fields are OPTIONAL.
+
+Below are some common fields that MAY be used. Platforms MAY choose to
+add additional ones as needed.
+
+| Request Field | Type | Description |
+| --- | --- | --- |
+| app_guid | string | GUID of an application associated with the binding. For [credentials](#types-of-binding) bindings. |
```
pmorie: I had been under the impression that app_guid is required for credentials bindings and that route is required for route_service bindings -- am I mistaken?

avade: app_guid is optional, some bindings may not have an associated app, in CF we call these service keys.

avade:
LGTM

spec.md
duglin: :-)

spec.md
duglin: correct for app_guid, but I think @pmorie is right that "route" is required for route_service bindings, right?  I'll add text to make that clearer.

spec.md
shalako: :(

duglin: @shalako keeping things at 80 cols makes diffs (and in particular github diffs) a lot nicer. It doesn't do well with long lines and it increases the chances of conflicts if two people happen t touch different parts of the same really long line.

shalako:
LGTM

spec.md
shalako: I prefer formatting of documentation source to be as close to the user-consumed product as possible. As I wouldn't publish documentation with sentences arbitrarily wrapped like a poem, and which can't expand with browser width, and I can't assume when new lines in source are honored or joined, I prefer them joined in source.

angarg12:
LGTM

spec.md
duglin: I think it depends on which view of the file is for what user. For example, I think the generated html (e.g. CF docs) and markdown (via github/browser) are for end users. And I agree those should not have 80 col wrapping. Which they don't.  However, I don't expect an end user to ever look at these files in their 'raw' format except when they go into a more of an "expert mode". Kind of like when a user chooses to look at the source (html) of a web page. Most never will. This means that the raw version should be in the form that's best for the user of that form - meaning us.  And long lines, IMO, make it harder on us to do our job since it causes more conflicts and git's diff facility doesn't handle it very well.

duglin: For example, its really hard to see the changes here:  
https://github.com/openservicebrokerapi/servicebroker/pull/215/files#diff-958e7270f96f5407d7d980f500805b1bL497

image

shalako: This seems like a github bug.

vaikas-google:
LGTM

spec.md
duglin: Actually, I think #179 will add the needed clarity for this - if not let me know and I'll open a new PR

mattmcneeney:
@duglin The spec is now a bit inconsistent with some 80 char breaks and some long lines. I agree with @shalako that limiting every line could make rendered documentation ugly, but regardless of which we choose, we should still tidy up the spec at some point for consistency.

PR #197, opened 2017-05-10
Make sure strings can't be empty when appropriate

duglin:
I decided that while in most cases an empty string can be treated the
same as the property being absent, when it came to IDs I thought it was
best to make it clear that if an optional ID is present then it MUST
have a value. Otherwise someone could write code to deal with an empty
string ID and that might have different semantics from no ID at all.
E.g. construct a string of the form xxx?id=  when there is a value
(even if empty), but if its missing it might construct it as just "xxx".
And having that difference would be an interop problem.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -117,9 +117,9 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
 
 | Response field | Type | Description |
 | --- | --- | --- |
-| name* | string | A CLI-friendly name of the service. All lowercase, no spaces. This MUST be globally unique within a platform marketplace. |
-| id* | string | An identifier used to correlate this service in future requests to the broker. This MUST be globally unique within a platform marketplace. Using a GUID is RECOMMENDED. |
-| description* | string | A short description of the service. |
+| name* | string | A CLI-friendly name of the service. All lowercase, no spaces. This MUST be globally unique within a platform marketplace. MUST be a non-empty string. |
+| id* | string | An identifier used to correlate this service in future requests to the broker. This MUST be globally unique within a platform marketplace. MUST be a non-empty string. Using a GUID is RECOMMENDED. |
+| description* | string | A short description of the service. MUST be a non-empty string. |
 | tags | array-of-strings | Tags provide a flexible mechanism to expose a classification, attribute, or base technology of a service, enabling equivalent services to be swapped out without changes to dependent logic in applications, buildpacks, or other services. E.g. mysql, relational, redis, key-value, caching, messaging, amqp. |
 | requires | array-of-strings | A list of permissions that the user would have to give the service, if they provision it. The only permissions currently supported are `syslog_drain`, `route_forwarding` and `volume_mount`. |
 | bindable* | boolean | Specifies whether instances of the service can be bound to applications. This specifies the default for all plans of this service. Plans can override this field (see [Plan Object](#plan-object)). |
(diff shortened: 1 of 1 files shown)


angarg12:
Can we make sure that the JSON schema also considers this cases?

duglin:
rebased

angarg12:
LGTM

duglin:
rebased

angarg12:
LGTM

spec.md
```diff
@@ -132,18 +132,18 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
...
 
 ##### Plan Object
 
 | Response field | Type | Description |
 | --- | --- | --- |
-| id* | string | An identifier used to correlate this plan in future requests to the broker. This MUST be globally unique within a platform marketplace. Using a GUID is RECOMMENDED. |
-| name* | string | The CLI-friendly name of the plan. MUST be unique within the service. All lowercase, no spaces. |
-| description* | string | A short description of the plan. |
+| id* | string | An identifier used to correlate this plan in future requests to the broker. This MUST be globally unique within a platform marketplace. MUST be a non-empty string. Using a GUID is RECOMMENDED. |
+| name* | string | The CLI-friendly name of the plan. MUST be unique within the service. MUST be a non-emptry string containing only lowercase alphanumeric characters, no spaces. |
```
shalako: empty is spelled incorrectly

I would suggest `All lowercase, no spaces. MUST be a non-empty string.`

The reason is that that the snakecase guidelines for name is meant to provide a better CLI UX, but CF doesn't currently validate it other than it is a string. Hyphens and underscores should be ok. I wouldn't want to say that this field must only be alphanumeric. However, if we're going to open that can of worms, it should be addressed for a number of other display name fields (e.g. service name). Instead, I'd recommend leaving the guidance as "All lowercase, no spaces."

spec.md
```diff
@@ -350,12 +350,10 @@ The `:instance_id` of a service instance is provided by the platform. This ID wi
 | accepts_incomplete | boolean | A value of true indicates that the marketplace and its clients support asynchronous broker operations. If this parameter is not included in the request, and the broker can only provision an instance of the requested plan asynchronously, the broker MUST reject the request with a `422 Unprocessable Entity` as described below. |
 
 ##### Body
-| Request field | Type | Description |
-| --- | --- | --- |
-| service_id* | string | The ID of the service (from the catalog). MUST be globally unique. |
-| plan_id* | string | The ID of the plan (from the catalog) for which the service instance has been requested. MUST be unique to a service. |
+| service_id* | string | The ID of the service (from the catalog). MUST be globally unique. MUST be a non-empty string. |
```
shalako: Have you removed the table header row?

spec.md
```diff
@@ -443,14 +441,14 @@ Not all permutations of plan changes are expected to be supported. For example,
 
 | Request Field | Type | Description |
 | --- | --- | --- |
-| service_id* | string | The ID of the service (from the catalog). MUST be globally unique. |
-| plan_id | string | The ID of the plan (from the catalog) for which the service instance has been requested. MUST be unique to a service. |
+| service_id* | string | The ID of the service (from the catalog). MUST be globally unique. MUST be a non-emptry string. |
```
shalako: empty is spelled incorrectly again. I recommend a spellcheck

spec.md
duglin: updated.  The wishy-washy-ness of "All lowercase, no spaces" not being normative makes me twitch a little :-)  but we can deal with that later.  For now at least this makes us more consistent with other spots in the spec.

spec.md
duglin: oops good catch - fixed

spec.md
duglin: fixed

avade:
lgtm

vaikas-google:
LGTM

angarg12:
LGTM

shalako:
LGTM

PR #198, opened 2017-05-11
Clarify default value of plan_updateable

pmorie:
It is specified in text later in the document, but was not specified in the table of response body fields.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -125,7 +125,7 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
 | bindable* | boolean | Specifies whether instances of the service can be bound to applications. This specifies the default for all plans of this service. Plans can override this field (see [Plan Object](#plan-object)). |
 | metadata | JSON object | An opaque object of metadata for a service offering. Controller treats this as a blob. Note that there are [conventions](https://docs.cloudfoundry.org/services/catalog-metadata.html) in existing brokers and controllers for fields that aid in the display of catalog data. |
 | [dashboard_client](#dashboard-client-object) | object | Contains the data necessary to activate the Dashboard SSO feature for this service. |
-| plan_updateable | boolean | Whether the service supports upgrade/downgrade for some plans. Please note that the misspelling of the attribute `plan_updatable` to `plan_updateable` was done by mistake. We have opted to keep that misspelling instead of fixing it and thus breaking backward compatibility. |
+| plan_updateable | boolean | Whether the service supports upgrade/downgrade for some plans. Please note that the misspelling of the attribute `plan_updatable` to `plan_updateable` was done by mistake. We have opted to keep that misspelling instead of fixing it and thus breaking backward compatibility.  Defaults to false. |
 | [plans*](#plan-object) | array-of-objects | A list of plans for this service, schema is defined below. MUST contain at least one plan. |
 
 ##### Dashboard Client Object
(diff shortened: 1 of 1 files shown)


angarg12:
LGTM

spec.md
```diff
@@ -125,7 +125,7 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
...
 | metadata  |  JSON object | An opaque object of metadata for a service offering. Controller treats this as a blob. Note that there are (conventions)[https://docs.cloudfoundry.org/services/catalog-metadata.html] in existing brokers and controllers for fields that aid in the display of catalog data. |
 | [dashboard_client](#DObject)  |  object |  Contains the data necessary to activate the Dashboard SSO feature for this service |
-| plan\_updateable  | boolean  |  Whether the service supports upgrade/downgrade for some plans. Please note that the misspelling of the attribute <code>plan\_updatable</code> to <code>plan\_updateable</code> was done by mistake. We have opted to keep that misspelling instead of fixing it and thus breaking backward compatibility.  |
+| plan\_updateable  | boolean  |  Whether the service supports upgrade/downgrade for some plans. Please note that the misspelling of the attribute <code>plan\_updatable</code> to <code>plan\_updateable</code> was done by mistake. We have opted to keep that misspelling instead of fixing it and thus breaking backward compatibility.  Defaults to false. |
```
avade: Should we have a defaults column in the table? 

This is really instructions to the platform and an FYI to broker authors that it is safe to leave this unconfigured.

"Platforms MUST default to false" 

@duglin thoughts?

mattmcneeney: There's also a double space before Defaults. And if possible please can we change `<code>plan\_updatable</code>` to `` `plan_updateable` `` (my tidying up PR does this for the rest of the spec).

duglin: No strong opinion on whether we add a "Default" column vs add text.  I think it might depend on whether we think we need to specify a default on most things or just a few.

I've seen "default" language like this w/ and w/o MUSTs in other specs so either seems fine. It might be good to throw a `If not present, ...` in there though.  But I'm ok either way.

duglin:
LGTM

spec.md
pmorie: > "Platforms MUST default to false"

This is very different than saying the field is optional and defaults to false.  If the instruction is for the platform, it means the field is required to be set, to me.

Do we have a mismatch?

pmorie: @mattmcneeney The reference to `plan_updatable` here is where we establish that the field is spelled incorrectly, so I think we should leave this one in :)

pmorie: I'll remove the double space before `Defaults`

angarg12:
Yikes, needs rebase.

shalako:
LGTM

avade:
lgtm

vaikas-google:
Will LGTM after rebase, since it starts the counter again iiuc.

angarg12:
LGTM

duglin:
LGTM

avade:
LGTM

angarg12:
The rebase nightmare strikes once again @pmorie

duglin:
So close!  :-)

pmorie:
heh - will rebase tonight

pmorie:
Heh.  I think we are bound to have rebase issues as long as the entire spec is in a single file, but I'm sure not trying to solve that problem at 1am.

pmorie:
Rebased

angarg12:
LGTM - quickly, before more rebases are needed

avade:
LGTM

pmorie:
LGTM, what a PR.  What a guy

duglin:
LOL I thought we were avoiding self-LGTMs  :-)

duglin:
LGTM

PR #202, opened 2017-05-16
Add master to title

mattmcneeney:
I think we should make it obvious which version people are looking at, especially since Cloud Foundry and other platforms may pull this spec in to form part of their own documentation.

The version 2.11 is already present in the docs, but this change just adds it to the title to make it clearer.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1,4 +1,4 @@
-# Open Service Broker API
+# Open Service Broker API (v2.11)
 
 ## Table of Contents
   - [API Overview](#api-overview)


avade:
Good idea.

LGTM

angarg12:
LGTM :ok_hand:

spec.md
```diff
@@ -1,4 +1,4 @@
-# Open Service Broker API
+# Open Service Broker API (v2.11)
```
pmorie: In master, should we also make a note that the spec may contain changes which are not yet release?

mattmcneeney:
In place of the version number? Yeah makes sense to me. Just need to add
the fact that this needs to change to the release notes.

duglin:
LGTM

spec.md
duglin: change this to "master" or "WIP".

Issue #206, opened 2017-05-17
Can a service have zero plans?


duglin:
If yes then we should say so in the spec and perhaps give a reason for it.
If not then we should mandate that the list can't be empty.

pmorie:
As far as I understand, it cannot, because plan ID is required when provisioning.

mattmcneeney:
If you register a service in CF today with no plans, I think we just hide it from the marketplace. I can't think of any reason for a service broker to either have no services, or a service with no plans, as I don't think we have a use case for which this would be useful for platforms.

I'd be keen to explicitly state that a broker must expose at least one service, and that services must expose at least one plan.

avade:
> I'd be keen to explicitly state that a broker must expose at least one service, and that services must expose at least one plan.

Agreed - want to draft a PR?

mattmcneeney:
Done!

duglin:
Closing since this was addressed in PR #212

PR #212, opened 2017-05-30
Brokers must have at least one service, and services must have at least one plan

mattmcneeney:
https://github.com/openservicebrokerapi/servicebroker/issues/206

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -111,7 +111,7 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
 
 | Response field | Type | Description |
 | --- | --- | --- |
-| services* | array-of-service-objects | Schema of service objects defined below. MUST contain at least one service. |
+| services* | array-of-service-objects | Schema of service objects defined below. MAY be empty. |
 
 ##### Service Objects
 
--- a/spec.md
+++ b/spec.md
@@ -111,7 +111,7 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
 
 | Response field | Type | Description |
 | --- | --- | --- |
-| services* | array-of-service-objects | Schema of service objects defined below. |
+| services* | array-of-service-objects | Schema of service objects defined below. MUST contain at least one service. |
 
 ##### Service Objects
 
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -111,7 +111,7 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
 
 | Response field | Type | Description |
 | --- | --- | --- |
-| services* | array-of-service-objects | Schema of service objects defined below. |
+| services* | array-of-service-objects | Schema of service objects defined below. MUST contain at least one service. |
```
angarg12: Small wording, should we say 'MUST be non-empty' or the like instead?

avade: I think as the type is a list of services, specifying at least one service makes sense.

angarg12:
LGTM

duglin:
hmm, if we require non-empty service lists what does the broker return if it has none? Some 4xx/5xx error?  Feel odd to return an error when its not really an error, its possible a broker could return a non-zero list for other platforms, just not this current platform.  Perhaps empty service list really is ok.

mattmcneeney:
That's true. I hadn't thought of a use case for a broker returning no services, but it makes sense if it was only supporting a subset of the platforms that could call it. I'll update this PR.

mattmcneeney:
However you could make the same argument for plans, right? Is this PR redundant?

duglin:
I think plans are different because a platform can't do anything with a service that doesn't have any plans, so returning the service at that point is pointless.

So perhaps this PR should just be:
1 - a service w/o any plans MUST be excluded
2 - a service list MAY be empty.

mattmcneeney:
I changed this so that services `MAY be empty` and plans `MUST contain at least one plan`. I think that covers those points?

duglin:
Yup it implies it, so unless someone wants it stated explicitly....
LGTM

vaikas-google:
LGTM

avade:
lgtm

angarg12:
LGTM

PR #218, opened 2017-06-05
Fix bad italics in docs

mattmcneeney:
For the four lines below only, bookbinder is incorrectly treating the underscore as an italics marker. This fixes this issue in the CF docs, so if y'all wouldn't mind, it would be great to squeeze this into this release.

@shalako (thanks for the heads up)
@duglin @avade @vaikas-google @pmorie

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -446,10 +446,10 @@ Not all permutations of plan changes are expected to be supported. For example, 
 | plan_id | string | The ID of the plan (from the catalog) for which the service instance has been requested. MUST be unique to a service. If present, MUST be a non-empty string. |
 | parameters | JSON object | Configuration options for the service instance. An opaque object, controller treats this as a blob. Brokers SHOULD ensure that the client has provided valid configuration parameters and values for the operation. |
 | previous_values | object | Information about the instance prior to the update. |
-| previous_values.service_id | string | ID of the service for the instance. If present, MUST be a non-empty string. |
-| previous_values.plan_id | string | ID of the plan prior to the update. If present, MUST be a non-empty string. |
-| previous_values.organization_id | string | ID of the organization specified for the instance. If present, MUST be a non-empty string. |
-| previous_values.space_id | string | ID of the space specified for the instance. If present, MUST be a non-empty string. |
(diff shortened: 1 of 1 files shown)


avade:
Does this mess with other markdown renders?

mattmcneeney:
This works with Bookbinder and Github markdown renderers. If anyone uses something else, please check it doesn't break stuff and then L G T M 😄

avade:
ok, thanks, Matt.

LGTM

duglin:
we really need to find a better solution than trying to support two versions of md in one file. Can we please check to see if the CF docs process can be modified so that it doesn't take the md file directly from github but instead can support reading from a local file? Then we can send the github version thru a pre-processor to do things like add these underscores and not have to worry about this going forward. It seems to me that since Bookbinder appears to be a CF owned app it shouldn't be too hard to make this minor change.  I'll volunteer to help create the converter, if someone from the CF side will modify Bookbinder to read from a local file.

mattmcneeney:
@bentarnoff Do you have any idea how difficult it would be to add support for this? I think this is the only discrepancy between how Github renders Markdown vs Bookbinder. Thanks!

duglin:
In this underscore case, shouldn't BB not pick-up the underscore in this case? it seem to me that an underscore in the middle of a word (no space before or after it) should not be treated as a signal to turn on italics.

avade:
If bookbinder is picking up `_` in full words then there is an issue. `_hello_` should result in italics where `hello_hello` should not.

Currently, there is no active team on bookbinder so we would need to PR to get this fixed.

@mattmcneeney could your team look into it? or we could pair tomorrow for a bit.

mattmcneeney:
Going to take a look at this today (puts developer hat on). Will feedback any findings.

mattmcneeney:
PR submitted to fix this @shalako @avade @bentarnoff
https://github.com/pivotal-cf/bookbinder/pull/116

Going to close this PR since it will no longer be required.

avade:
Awesome!

duglin:
thanks!

PR #222, opened 2017-06-12
Add an Originating Identity http header

### LINKED ISSUE #110: broker requires user identity to perform operations on users behalf
Proposal from @gberche-orange:
https://docs.google.com/document/d/1DoAbJa_YiGIJbOZ_zPzakh7sc4TB9Tmadq41cfSX0dw/edit?usp=sharing. Discussion of that proposal: http://cf-dev.70369.x6.nabble.com/cf-dev-service-broker-user-delegation-beyond-service-dashboard-tt2658.html

Proposal from @fraenkel: https://docs.google.com/document/d/1w4swx_5erf_NmPtfcCW2fn2jbEBfRHf2xLVhWpc1PLM/edit

duglin:
Per my AI from last week's call, I've created a PR for the proposal so its not hidden within the conversation in #110.

Closes #110

### DIFF
--- a/profile.md
+++ b/profile.md
@@ -6,9 +6,14 @@ The [Open Service Broker API specification](spec.md) allows for extensions
 and variations based on the environments in which it is being used; this
 document contains the suggested usage pattern for some of those variants.
 
+While use of this profile is OPTIONAL, an implementation is not compliant
+with this profile if it fails to satisfy one or more of the MUST, SHALL
+or REQUIRED level requirements defined herein.
+
 ## Table of Contents
 
 - [Notations and Terminology](#notations-and-terminology)
+- [Originating Identity Header](#originating-identity-header)
 - [Context Object](#context-object)
   - [Context Object Properties](#context-object-properties)
   - [Cloud Foundry](#cloud-foundry)
@@ -55,6 +60,77 @@ the [Open Service Broker API specification](spec.md) SHALL take precedence.
 
 - *Application*: The software that uses a Service Instance via Service Binding.
 
+## Originating Identity Header
+
+In the [Open Service Broker API specification](spec.md) it defines an
+additional HTTP Header that can be included in messages from the Platform
+to identify the user that requested the action to be taken.
+
+The header consists of two parts: a `platform` string and `value` string,
+where the `value` is a Base64 encoded serialized JSON object.
+Both parts will vary based on the Platform which is being used. The
+following sections define the values to be used based on the Platform
+and which properties are expected to appear in the `value` JSON.
+
+Note that when both the originating identity HTTP Header and the Context
+object appear in the same message the `platform` value MUST be the same
+for both.
+
+### Cloud Foundry
+
+*`platform` Value*: `cloudfoundry`
+
+The following properties MUST appear within the JSON encoded `value`:
+
+| Property | Type | Description |
+| --- | --- | --- |
+| user_id | string | The `user_id` value from the Cloud Foundry JWT token. |
+
+Platforms MAY include additional properties.
+
+For example, a `value` of:
+```
+{
+  "user_id": "683ea748-3092-4ff4-b656-39cacc4d5360"
+}
+```
+would appear in the HTTP Header as:
+```
+X-Broker-API-Originating-Identity: cloudfoundry eyANCiAgInVzZXJfaWQiOiAiNjgzZWE3NDgtMzA5Mi00ZmY0LWI2NTYtMzljYWNjNGQ1MzYwIiwNCiAgInVzZXJfbmFtZSI6ICJqb2VAZXhhbXBsZS5jb20iDQp9
+```
+
+### Kubernetes
+
+*`platform` Value*: `kubernetes`
+
+The following properties MUST appear within the JSON encoded `value`:
+
+| Property | Type | Description |
+| --- | --- | --- |
+| username | string | The `username` property from the Kubenernetes `user.info` object. |
+| uid | string | The `uid` property from the Kubenernetes `user.info` object. |
+| groups | array-of-strings | The `groups` property from the Kubenernetes `user.info` object. |
+| extra | map-of-array-of-strings | The `extra` property from the Kubernetes `user.info` object. |
+
+Platforms MAY include additional properties.
+
+For example, a `value` of:
+```
+{
+  "username": "duke",
+  "uid": "c2dde242-5ce4-11e7-988c-000c2946f14f",
+  "groups": { "admin", "dev" },
+  "extra": {
+    "mydata": { "data1", "data3" }
+  }
+}
+```
+would appear in the HTTP Header as:
+```
+X-Broker-API-Originating-Identity: kubernetes eyANCiAgInVzZXJuYW1lIjogImR1a2UiLA0KICAidWlkIjogImMyZGRlMjQyLTVjZTQtMTFlNy05ODhjLTAwMGMyOTQ2ZjE0ZiIsDQogICJncm91cHMiOiB7ICJhZG1pbiIsICJkZXYiIH0NCn0=
+```
+
+
 ## Context Object
 
 In the [Open Service Broker API specification](spec.md) there are certain
@@ -68,10 +144,6 @@ While the `context` property is defined as an opaque JSON object, in practice,
 it is often useful and necessary for there to be an agreed upon set of
 properties to ensure a common understanding of this data between the
 Platform and the Service Brokers.
-
-While use of this profile is OPTIONAL, an implementation is not compliant
-with this profile if it fails to satisfy one or more of the MUST, SHALL
-or REQUIRED level requirements defined herein.
 
 ### Context Object Properties
 
@@ -92,6 +164,10 @@ The `platform` property MUST be a `string` and serialized as follows:
 
 Each section below will define the `platform` value that MUST be used based
 on the Platform and the set of additional properties that MUST be present.
+
+Note that when both the originating identity HTTP Header and the Context
+object appear in the same message the `platform` value MUST be the same
+for both.
 
 #### Cloud Foundry
 
--- a/spec.md
+++ b/spec.md
@@ -9,6 +9,7 @@
   - [API Version Header](#api-version-header)
   - [Authentication](#authentication)
   - [URL Properties](#url-properties)
+  - [Originating Identity](#originating-identity)
   - [Catalog Management](#catalog-management)
     - [Adding a Broker to the Platform](#adding-a-broker-to-the-platform)
   - [Synchronous and Asynchronous Operations](#synchronous-and-asynchronous-operations)
@@ -148,6 +149,58 @@ If a character outside of the "Unreserved Characters" set is used, then
 it SHOULD be percent-encoded prior to being used as part of the HTTP request,
 per [RFC3986](https://tools.ietf.org/html/rfc3986#section-2.1).
 
+## Originating Identity
+
+Often a service broker will need to know the identity of the user that
+initiated the request from the platform. For example, this might be needed
+for auditing or authorization purposes. In order to facilitate this, the
+platform will need to provide this identification information to the broker
+on each request. Platforms MAY support this feature, and if they do, they
+MUST adhere to the following:
+- For any OSB API request that is the result of an action taken by
+  a platform's user, there MUST be an associated `OriginatingIdentity` header
+  on that HTTP request.
+- Any OSB API request that is not associated with an action from a platform's
+  user, such as the platform querying the marketplace, MAY
+  exclude the header from that HTTP request.
+- If present on a request, the `OriginatingIdentity` header MUST contain
+  the identify information for the platform's user that took the action
+  to cause the request to be sent.
+
+The format of the header MUST be:
+
+```
+X-Broker-API-Originating-Identity: platform value
+```
+
+`platform` MUST be a non-empty string indicating the platform from which
+the request is being sent. The specific value SHOULD match the values
+defined in the [profile](profile.md) document for the `context.platform`
+property. When `context` is sent as part of a message, this value MUST
+be the same as the `context.platform` value.
+
+`value` MUST be a Base64 encoded string. The string MUST be a serialized
+JSON object. The specific properties will be platform specific - see
+the [profile](profile.md) document for more information.
+
+For example:
+```
+X-Broker-API-Originating-Identity: cloudfoundry eyANCiAgInVzZXJfaWQiOiAiNjgzZWE3NDgtMzA5Mi00ZmY0LWI2NTYtMzljYWNjNGQ1MzYwIiwNCiAgInVzZXJfbmFtZSI6ICJqb2VAZXhhbXBsZS5jb20iDQp9
+```
+
+Where the `value`, when decoded, is:
+```
+{
+  "user_id": "683ea748-3092-4ff4-b656-39cacc4d5360"
+}
+```
+
+Note that not all messages sent to a broker are initiated by an end-user
+of the platform. For example, during orphan mitigation or during the
+querying of the broker's catalog, the platform might not have an end-user
+with which to associate the request, therefore in those cases the
+originating identity header would not be included in those messages.
+
 ## Catalog Management
 
 The first endpoint that a platform will interact with on the broker is the
@@ -183,6 +236,16 @@ Broker API.
 
 #### Route
 `GET /v2/catalog`
+
+#### Headers
+
+The following HTTP Headers are defined for this operation:
+
+| Header | Type | Description |
+| --- | --- | --- |
+| X-Broker-API-Version* | string | See [API Version Header](#api-version-header). |
+
+\* Headers with an asterisk are REQUIRED.
 
 #### cURL
 ```
@@ -496,6 +559,16 @@ The request provides these query string parameters as useful hints for brokers.
 
 Note: Although the request query parameters `service_id` and `plan_id` are not mandatory, the platform SHOULD include them on all `last_operation` requests it makes to service brokers.
 
+#### Headers
+
+The following HTTP Headers are defined for this operation:
+
+| Header | Type | Description |
+| --- | --- | --- |
+| X-Broker-API-Version* | string | See [API Version Header](#api-version-header). |
+
+\* Headers with an asterisk are REQUIRED.
+
 #### cURL
 ```
 $ curl http://username:password@broker-url/v2/service_instances/:instance_id/last_operation
@@ -552,6 +625,17 @@ broker will use it to correlate the resource it creates.
 | Parameter name | Type | Description |
 | --- | --- | --- |
 | accepts_incomplete | boolean | A value of true indicates that the marketplace and its clients support asynchronous broker operations. If this parameter is not included in the request, and the broker can only provision a service instance of the requested plan asynchronously, the broker MUST reject the request with a `422 Unprocessable Entity` as described below. |
+
+#### Headers
+
+The following HTTP Headers are defined for this operation:
+
+| Header | Type | Description |
+| --- | --- | --- |
+| X-Broker-API-Version* | string | See [API Version Header](#api-version-header). |
+| X-Broker-API-Originating-Identity | string | See [Originating Identity](#originating-identity). |
+
+\* Headers with an asterisk are REQUIRED.
 
 #### Body
 | Request field | Type | Description |
@@ -653,6 +737,17 @@ It MUST be the ID a previously provisioned service instance.
 | Parameter name | Type | Description |
 | --- | --- | --- |
 | accepts_incomplete | boolean | A value of true indicates that the marketplace and its clients support asynchronous broker operations. If this parameter is not included in the request, and the broker can only provision a service instance of the requested plan asynchronously, the broker SHOULD reject the request with a `422 Unprocessable Entity` as described below. |
+
+#### Headers
+
+The following HTTP Headers are defined for this operation:
+
+| Header | Type | Description |
+| --- | --- | --- |
+| X-Broker-API-Version* | string | See [API Version Header](#api-version-header). |
+| X-Broker-API-Originating-Identity | string | See [Originating Identity](#originating-identity). |
+
+\* Headers with an asterisk are REQUIRED.
 
 #### Body
 
@@ -812,6 +907,17 @@ It MUST be the ID of a previously provisioned service instance.
 `:binding_id` MUST be a globally unique non-empty string.
 This ID will be used for future unbind requests, so the broker will use
 it to correlate the resource it creates.
+
+#### Headers
+
+The following HTTP Headers are defined for this operation:
+
+| Header | Type | Description |
+| --- | --- | --- |
+| X-Broker-API-Version* | string | See [API Version Header](#api-version-header). |
+| X-Broker-API-Originating-Identity | string | See [Originating Identity](#originating-identity). |
+
+\* Headers with an asterisk are REQUIRED.
 
 #### Body
 
@@ -992,6 +1098,17 @@ The request provides these query string parameters as useful hints for brokers.
 
 \* Query parameters with an asterisk are REQUIRED.
 
+#### Headers
+
+The following HTTP Headers are defined for this operation:
+
+| Header | Type | Description |
+| --- | --- | --- |
+| X-Broker-API-Version* | string | See [API Version Header](#api-version-header). |
+| X-Broker-API-Originating-Identity | string | See [Originating Identity](#originating-identity). |
+
+\* Headers with an asterisk are REQUIRED.
+
 #### cURL
 ```
 $ curl 'http://username:password@broker-url/v2/service_instances/:instance_id/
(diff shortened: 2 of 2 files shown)


jim-minter:
(copy from #110) I think there's value in copying the Authorization header more closely as follows:
a) use base64 encoding - less chance of implementers tripping up compared to ;charset; spec makes no comment about what's contained inside
b) prepend an identity scheme in us-ascii followed by space (similar to the 'Basic' or 'Bearer' in the Authorization header); allows some extensibility and allows implementations to recognise a scheme without diving into the base64 blob.

example: X-Broker-API-Originating-Identity: MyJSONUserAndGroupScheme eyJ1c2VyIjoiZm9vIiwiZ3JvdXBzIjpbImJhciIsImJheiJdfQ==

spec.md
```diff
@@ -103,6 +104,26 @@ This header allows brokers to reject requests from marketplaces for versions the
...
 basic authentication (the `Authorization:` header) on every request. The broker is responsible for validating the username and password and returning a `401 Unauthorized` message if credentials are invalid. It is RECOMMENDED that brokers support secure communication from platform marketplaces over TLS.
 
+## Originating Identity
+
+Often a service broker will need to know the identity of the user that
+initiated the request from the platform. For example, this might be needed
+for auditing or authorization purposes. In order to facilitate this, the
+platform will need to provide this identification information to the broker
+on each request. If the platform wishes to provide this information then
+it MUST include an additional HTTP header on all requests it sends to the
```
pmorie: Is this really all requests?  Or just creational requests (ie, provision, bind)

pmorie: Also, 'MUST' seems like a backward-incompatible change.

spec.md
```diff
@@ -103,6 +104,26 @@ This header allows brokers to reject requests from marketplaces for versions the
...
+broker. The format of the header MUST be:
+
+```
+X-Broker-API-Originating-Identity: value[;charset=...]
+```
```
pmorie: Do we need to send a scheme as well?  Ie, 'I am sending a kubernetes identity'

pmorie:
We should note specifically in the spec body when the platform should send this header on a request by request basis.

spec.md
duglin: Its for all requests.

The "MUST" is predicated by the "if".  If you want to send this info then it MUST be done this way.

spec.md
duglin: I don't think so. Right now the assumption is that there's some out-of-band understanding between the platform and broker on how to interpret this data.

duglin:
@pmorie I like the idea of being more explicit - I think it makes sense to do it for the other http header(s) too - will give it a shot and see how it looks.

spec.md
duglin: correction - all except GET /v2/catalog

pmorie: I gotcha, missed the 'if' qualification

vaikas-google:
Can we discuss the base64 encoding brought up by @jim-minter ?

pmorie:
I also support @jim-minter's suggestion of base64 encoding the identity payload (comment: https://github.com/openservicebrokerapi/servicebroker/pull/222#issuecomment-307812990)

spec.md
```diff
@@ -380,6 +423,17 @@ The `:instance_id` of a service instance is provided by the platform. This ID wi
 | --- | --- | --- |
 | accepts_incomplete | boolean | A value of true indicates that the marketplace and its clients support asynchronous broker operations. If this parameter is not included in the request, and the broker can only provision a service instance of the requested plan asynchronously, the broker MUST reject the request with a `422 Unprocessable Entity` as described below. |
 
+##### Headers
+
+The following HTTP Headers are defined for this operation:
+
+| Header | Type | Description |
+| --- | --- | --- |
+| X-Broker-API-Version* | string | See [API Version Header](#api-version-header) |
```
pmorie: I think we should leave the other headers out of this proposal

duglin: I'm not comfortable leaving the spec in a 1/2 completed state. Either we add a table for http headers to each API and fill it out completely, or we don't (perhaps leaving it for a new PR). But I'm not keen on a PR that does 1/2 the job.

mattmcneeney: A swagger spec would resolve the need for this duplication

shalako: I'm in favor of having a headers table for each endpoint.

shalako:
Thoughts on base64 encoding the header? @duglin @mattmcneeney Could you check with broker authors? I'll send a ping to cf-dev@

spec.md
```diff
@@ -92,7 +93,7 @@ that do not understand them.
 
 Requests from the platform to the service broker MUST contain a header that declares the version number of the Service Broker API that the marketplace will use:
 
-`X-Broker-Api-Version: 2.11`
+`X-Broker-API-Version: 2.11`
```
pmorie: Opened https://github.com/openservicebrokerapi/servicebroker/pull/246 for this change

spec.md
```diff
@@ -103,6 +104,27 @@ This header allows brokers to reject requests from marketplaces for versions the
...
+broker, except the request to retrieve the catalog since that is not
+associated with a platform's user. The format of the header MUST be:
+
+```
+X-Broker-API-Originating-Identity: value[;charset=...]
+```
+Where:
+- its REQUIRED `value` is a non-empty string.
+- it MAY include a `charset` parameter specifying the character encoding of
+  the `value`. If not specified, the default value MUST be `us-ascii`.
```
pmorie: Did we want to give an example for CF and Kubernetes here?

pmorie: Also, do we want to discuss that the `value` may be any scheme?  For example, in Kubernetes, we'll probably want to send the base64 encoded userinfo.  See https://github.com/kubernetes-incubator/service-catalog/pull/939

spec.md
duglin: Any reason you created a dup PR?

spec.md
duglin: I'd kind of like to leave that up to the platform instead of being prescriptive since the needs of the platform may vary.

duglin:
From today's call I have a few questions about some of the things people were proposing:
- base64 encoding of the value. I'd like to understand why we should do this. I'm not totally against it but I'm wondering if its because people think it'll be used to send sensitive data?  If so we should discuss that because I don't think that was the intent - its meant to just convey the user initiating the request, nothing more.  As such, needing something like a scheme seems like an unnecessary thing to me.
- @mattmcneeney you said you wanted more structure/definition around the value itself. Can you provide a concrete proposal for what you're looking for and why if we limit it to just "the user"?
- if we extend this beyond a simple string and allow it to be something like a json object, I'm worried that people will use it to store tons of other data that's not really the intent of this header. People should probably be using the Body/Context to send this other info but I'd like to understand the usecases people have in mind so we can make the right decision.
- @pmorie you were mentioned using it to store data about where in the platform to provision stuff. W/o knowing more about the usecase that would seem to be better placed into the Context object, or someplace else in the body like the parameters, but we can get more info about this use?

At the last f2f we discussed moving the Context into and http header and realized that it would be problematic - not least of which because of the potential to hit a length limit.  Encouraging arbitrary data in this PR's header, and not limiting it to something as simple as an email/userID/uuid/etc.. will lead us back to that same state, so it might be best if we only focus on our immediate needs.

jim-minter:
> base64 encoding of the value
My intention when suggesting this was to reuse a standard, recognisable, simple to implement mechanism to allow small amounts of arbitrary data to be reliably exchanged.  Not to obfuscate sensitive data.

> json object
A relevant example here is sending both the user identifier and their group memberships, for example.  Not reams of information, but enough to make a compound object useful (and if it's base64 encoded by default, then there's protection against usernames with non-ascii characters, etc.  I'd hope people wouldn't intend to send more than what you might find in a HTTP Authorization header.

> People should probably be using the Body/Context to send this other info
Yes, this would probably better, but one issue is the body on DELETE problem, and it would be really nice to have access to user identification on delete to be able to check authorization.

mattmcneeney:
@duglin As far as I know, we don't have a use case in CF for this _today_. I can imagine us wanting to use this for either auditing purposes or possibly for asking brokers to perform actions on a users' behalf. The latter would probably require more than one piece of information (ie. an access token and a refresh token?), but without any concrete examples of how we want to use this, it's hard to plan for the future. (cc @avade @shalako in case I've missed something)

I totally agree with you though that I think storing complex objects in the header would be a bad idea as it puts onus on brokers to deal with errors when the value is not in the format it expects. We purposely avoided doing this when discussing how to send `context` to brokers.

duglin:
@jim-minter yea, DELETE is always the problem child :-)  but I'm hoping that the semantics of that operation shouldn't need additional info that would require it to be passed in the body.  Besides, there's always query parameters or other HTTP headers if people really need it.  Mashing non-user-identifying info into this header should seem non-kosher to me.

I can see why we may need more than a single string, and thus a struct would be nice - if we add words to make it clear that this isn't a grab-bag to stick any random stuff (even though we can't technically stop people).  Do we really need a scheme tag or can we just say its always a base64 json object?

jim-minter:
@duglin it's definitely valuable to have the user info on delete - as much as we're discussing at the moment - and the header is good enough to provide that.  I'm not personally aware of other information that's needed on delete beyond that (the fact that there's no context object on delete isn't blocking me, at least).

I think a scheme tag is valuable for at least the following reasons:
1- hopefully prevents the need to manually configure/match-up the identity scheme that a catalog sends and that a broker receives at broker registration time
2- makes it easy for a broker to tell what scheme it's dealing with and discard early if it doesn't understand it
3- allows implementation of brokers which can understand multiple schemes if needs be (i.e. understands what kubernetes uses, and what CF uses and can be a broker to either catalog).

duglin:
Sorry if i wasn't clear - I do want this header on delete

duglin:
re: scheme - I think the assumption here is that the broker knows the platform (and therefore scheme or data it should expect) based on the identity of the auth negotiation of the request itself - and therefore will know the scheme of this header.  It would need to know this, along with probably other info, in order to use it to do things like call back into the platform anyway.  Not for simple logging purposes though, obviously, but in that case I doubt it needs to understand the json, it would just need to log it.

To be clear, I'm not 100% against adding a scheme, but I want to make sure we're not adding it unnecessarily because the info is already available thru some other means - and it seems to me that in order for someone to really use the data in the json object they'll need to know a lot more than just the scheme anyway.  We'd then need to agree on (standardize) the schemes too, no?

jim-minter:
@duglin maybe I'm beginning to understand.  Perhaps the scheme isn't utterly essential, but I think it does provide some protection, and maybe mandating it in the spec also provides a small foundation in case in the future a platform could handle multiple identities on behalf of a user and send the right one to a broker.

duglin:
Do you have some sample schemes in mind so I can get a sense of how many (and their shape/values) you're thinking of?

jim-minter:
Nothing more than "the scheme Kubernetes uses" (user,groups,extra); "the scheme CloudFoundry uses" (sorry, don't know what that is).  Perhaps "username on its own" for good measure.

duglin:
hmmm would it make sense to push for it to be the same string that used as the "platform" field in the "context" object?  To get some consistency.

duglin:
revamped and ready for review!

profile.md
```diff
@@ -54,6 +59,75 @@ the [Open Service Broker API specification](spec.md) SHALL take precedence.
...
+
+Note that when both the originating identity HTTP Header and the Context
+object appear in the same message the `platform` value MUST be the same
+for both.
+
+### Cloud Foundry
+
+*`platform` Value*: `cloudfoundry`
+
+The following properties MUST appear within the JSON encoded `value`:
```
gberche-orange: I read both `user_name` and `user_id` MUST be provided by CF platforms to be compliant with the cloudfoundry profile.

 I can see cases where a cloudfoundry user would not want his user_name to be disclosed to untrusted brokers: e.g. `user_name` can be used to trigger denial of service by locking account after multiple unsuccessful login form attempts that use this `user_name` as user id. 

More specifically, a broker might be untrusted when offered by a 3rd party public Saas service provider, (provisionned by CF admins or possibly by space developpers through space private brokers), while some other corporate trusted brokers might indeed legitimaly need `user_name`.

Was it considered to publish to platform users the broker's intent to leverage `user_name`, e.g. by leveraging the existing catalog `requires ` key ?

Was it considered that the OSB spec provides rooms for cloudfoundry to omit `user_name` while remaining compliant ? For instance, during broker registration, admin may choose to opt-opt from exposing user_name to this broker.

duglin: That's a good point. I'm going to remove it since if its optional then people will need to deal with it being absent anyway so we might as well enforce consistency and just require user_id.

duglin: @pmorie it seems we may want to consider removing `username` from the Kube side of things for the same reason, no?

gberche-orange: > That's a good point. I'm going to remove it since if its optional then people will need to deal with it being absent anyway so we might as well enforce consistency and just require user_id.

I see cases where having `user_name` is useful: this avoids requiring trusted brokers that need user identity to have credentials with each platform and lookup user identity from user id in a platform specific way ( e.g. http://docs.cloudfoundry.org/api/uaa/#lookup-user-ids-usernames for CF). But this comes with a cost on platform side to control access/visibility of which broker are trusted or not.

duglin: Yup - there's a trade-off.  But for now I'd prefer to opt for the minimal amount of required data and see how people react to it - and only add optional data after we get feedback.

duglin:
Per the previous call, I made it clear that the header must be base64 encoded.

However, I decided to leave off charset. My thinking is that the json should use the same charset as the body and therefore I wonder if we really need to say anything at all about it. Thoughts?

shalako:
LGTM

avade:
After discussing with Doug and seeing the example header for auth [1], this LGTM

[1] https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization

profile.md
```diff
@@ -54,6 +59,73 @@ the [Open Service Broker API specification](spec.md) SHALL take precedence.
...
+
+### Kubernetes
+
+*`platform` Value*: `kubernetes`
+
+The following properties MUST appear within the JSON encoded `value`:
+
+| Property | Type | Description |
+| --- | --- | --- |
+| username | string | The `username` property from the Kubenernetes `UserInfo` object. |
```
pmorie: Nit: `UserInfo` -> `user.Info`

duglin: fixed

duglin:
fixed minor typo - @shalako @avade need your re-l.g.t.m

spec.md
```diff
@@ -112,6 +113,44 @@ This header allows brokers to reject requests from marketplaces for versions the
...
+```
+X-Broker-API-Originating-Identity: cloudfoundry eyANCiAgInVzZXJfaWQiOiAiNjgzZWE3NDgtMzA5Mi00ZmY0LWI2NTYtMzljYWNjNGQ1MzYwIiwNCiAgInVzZXJfbmFtZSI6ICJqb2VAZXhhbXBsZS5jb20iDQp9
+```
+
+Where the `value`, when decoded, is:
+```
+{
+  "user_id": "683ea748-3092-4ff4-b656-39cacc4d5360",
+  "user_name": "joe@example.com"
+}
```
shalako: This example doesn't match the one in profile.md, line 87. 

@avade @mattmcneeney Is CF going to send `user_id` or `user_id` and `user_name`?

shalako:
CF examples don't match between profile.md and spec.md

duglin:
good catch - fixed

shalako:
LGTM

fmui:
LGTM

duglin:
@shalako @fmui sorry - didn't notice the travis failure until today - fixed the formatting issues.  More l.g.t.m.'s needed.

angarg12:
LGTM

avade:
lgtm

shalako:
@mattmcneeney @avade Has this been implemented in CF yet?

mattmcneeney:
@shalako No not yet - we are waiting for this to get into validating through implementation, but if we don't think we need that step for this change, I can get it into our backlog soon ish

avade:
@mattmcneeney / @pmorie when do we expect feedback from CF / Service-Catalog on this?

Is it likely for 2.13 timeframe (early September)

duglin:
@avade I'll make sure we have an issue for it in svc-cat and see if we can get it scheduled soon.

duglin:
FYI - issue is here:  https://github.com/kubernetes-incubator/service-catalog/issues/934

mattmcneeney:
@avade Hoping to get this into a CF release in the next month, so will have feedback around then hopefully.

profile.md
mattmcneeney: @duglin in the kubernetes section you specify that the `value` field must be a base64 encoded JSON object, but there is no mention of the encoding scheme here. Should we add that?

duglin: yup -that's an oversight. The main spec says it needs to be base64 encoded ( https://github.com/openservicebrokerapi/servicebroker/pull/222/files#diff-958e7270f96f5407d7d980f500805b1bR137 ) so we should repeat it in the profile for both CF and Kube.

spec.md
```diff
@@ -112,6 +113,43 @@ This header allows brokers to reject requests from marketplaces for versions the
...
+## Originating Identity
+
+Often a service broker will need to know the identity of the user that
+initiated the request from the platform. For example, this might be needed
+for auditing or authorization purposes. In order to facilitate this, the
+platform will need to provide this identification information to the broker
+on each request. If the platform wishes to provide this information then
+it MUST include an additional HTTP header on all requests it sends to the
+broker, except the request to retrieve the catalog since that is not
+associated with a platform's user. The format of the header MUST be:
```
mattmcneeney: Two queries:
1. Do you want to audit who registered the service broker? In CF, whoever calls `cf create-service-broker` will implicitly call the catalog endpoint, so we could record whoever ran that command.
2. In some cases (ie. orphan mitigation), their will not be an originating user. In this case I guess the platform will just omit this header.
3. Following the above, is the sentence `If the platform wishes to provide this information then it MUST ...` strange? Maybe it could just read `Platforms wishing to provide this information CAN ...`. Combining a MUST with an optional thing seems odd reflecting back on this.

duglin: 1. perhaps but if a query for the latest catalog is a cron job who would be the user in that case? The same person who registered it originally? What if they've been deleted?
2. probably true - I added some text, see what you think.
3. "wishing" vs "if" is the same thing to me, so I don't mind changing that part if people want.  But, we do need a MUST in there so that people know they need to adhere to what we're defining if they're going to support this optional feature.  "CAN" isn't an RFC2119 keyword.

spec.md
```diff
@@ -126,6 +127,49 @@ response if the authentication fails.
...
+{
+  "user_id": "683ea748-3092-4ff4-b656-39cacc4d5360"
+}
+```
+
+Note that not all messages sent to a broker are initiated by an end-user
+of the platform. For example, during orphan mitigation or during the
+querying of the broker's catalog, the platform might not have an end-user
+with which to associate the request, therefore in those cases the
+originating identity header would not be included in those messages.
```
mattmcneeney: We should discuss how this applies to last operation endpoints. These endpoints may be polled by the platform in response to a user action (like an async create service instance), but it's the platform waiting for the operation to finish. It could be more work to support sending the header for this. What do you think @duglin ?

mattmcneeney:
Update: We're adding this to a CF release as we speak and will feedback soon once the implementation is complete (hopefully this week)

profile.md
```diff
@@ -55,6 +60,73 @@ the [Open Service Broker API specification](spec.md) SHALL take precedence.
...
+
+*`platform` Value*: `kubernetes`
+
+The following properties MUST appear within the JSON encoded `value`:
+
+| Property | Type | Description |
+| --- | --- | --- |
+| username | string | The `username` property from the Kubenernetes `user.info` object. |
+| uid | string | The `uid` property from the Kubenernetes `user.info` object. |
+| groups | string | The `groups` property from the Kubenernetes `user.info` object. |
```
MHBauer: Is `user.info` this API https://kubernetes.io/docs/api-reference/v1.7/#userinfo-v1beta1-authentication ?

It appears to be Group Version Kind: `Authentication v1beta1 UserInfo`.

liggitt: the `user.Info` object is an internal interface available to resthandlers in the request context, populated by the authentication layer. https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/authentication/user/user.go#L19-L42

it can be serialized as https://kubernetes.io/docs/api-reference/v1.7/#userinfo-v1-authentication (or the v1beta1 struct you referred to), or you can define your own type that serializes the same four fields.

liggitt: groups is `[]string`

mattmcneeney:
I learnt something interesting; prepending `X-` to headers was actually deprecated by the IETF in 2012 (see [this rfc](https://tools.ietf.org/html/rfc6648)). I think we should still use it though if only for consistency.

duglin:
Agreed

jmrodri:
@mattmcneeney @duglin +1 to keeping the `X-` despite that RFC. That RFC is news to me as well good for future knowledge.

profile.md
pmorie: this also needs to include the extrafields from the authorizer, or we'll be truncating some of the data about the user's identity.

Agree, @liggitt?

liggitt: Agree. Fully preserve the user.Info

spec.md
fmui: Re 3: "MAY" is a RFC2119 keyword that can be used here.

duglin: What if I reword this:
```
If the platform wishes to provide this information then it MUST include an additional HTTP
header on all requests it sends to the broker, except the request to retrieve the catalog
since that is not associated with a platform's user.
```
to be:
```
Platforms MAY support this feature. If they do, they MUST adhere to the following:
- for any OSB API request that is the result of an action taken by
  a platform's user, there MUST be an associated OriginatingIdentity header on that
  HTTP request.
- any OSB API request that is not associated with an action from a platform's user,
  such as the platform querying the marketplace, MAY exclude the header from that
  HTTP request.
- if present on a request, the OriginatingIdentity header MUST contain the identify
  information for the platform's user that took the action to cause the request to be sent.
```

spec.md
```diff
@@ -375,6 +429,17 @@ The request provides these query string parameters as useful hints for brokers.
...
 Note: Although the request query parameters `service_id` and `plan_id` are not mandatory, the platform SHOULD include them on all `last_operation` requests it makes to service brokers.
 
+#### Headers
+
+The following HTTP Headers are defined for this operation:
+
+| Header | Type | Description |
+| --- | --- | --- |
+| X-Broker-API-Version* | string | See [API Version Header](#api-version-header). |
+| X-Broker-API-Originating-Identity | string | See [Originating Identity](#originating-identity). |
```
pmorie: Is this the identity of the user that initiated the original action or the service account the platform component is running under?

I'm not sure that this makes sense to send on last operation, actually, because it seems like it's just checking the status of an action that's already been triggered.

Samze: In the Cloud Foundry implementation we omit the header during last_operation and certain orphan mitigation scenarios as it is the platform performing the action.

mattmcneeney:
A PR for this change has now been submitted to the Cloud Controller in CF as part of the validation through implementation: https://github.com/cloudfoundry/cloud_controller_ng/pull/877

profile.md
duglin: done

profile.md
duglin: done

spec.md
duglin: not hearing any objection, I made this change.

spec.md
duglin: removed

spec.md
duglin: I removed it from last op

mattmcneeney:
This has been merged into the Cloud Controller and the implementation has been validated.
https://github.com/cloudfoundry/cloud_controller_ng/pull/877

Calling for reviews!
@angarg12 @duglin @pmorie @skaegi @shalako @vaikas-google @fmui @avade

spec.md
```diff
@@ -126,6 +127,58 @@ response if the authentication fails.
...
 
+## Originating Identity
+
+Often a service broker will need to know the identity of the user that
+initiated the request from the platform. For example, this might be needed
+for auditing or authorization purposes. In order to facilitate this, the
+platform will need to provide this identification information to the broker
+on each request. Platforms MAY support this feature, and if they do, they
+MUST adhere to the following:
+- for any OSB API request that is the result of an action taken by
```
angarg12: Caps?

mattmcneeney:
Let's get this one merged folks. @duglin do you need to change anything following @angarg12's comment or are we good to go?

angarg12:
Yup, it also needs rebase.

spec.md
duglin: @angarg12 "Caps?" on which word(s) ?

angarg12: These 3 bullet points begin with a lowercase letter, - for, - any, - if

duglin: ah ok, done.

duglin:
addressed comments and rebased.

mattmcneeney:
LGTM 👏 👏 👏

angarg12:
LGTM

fmui:
LGTM

avade:
lgtm

PR #223, opened 2017-06-12
Make using auth/tls optional and allow for non-BasicAuth auth

### LINKED ISSUE #103: Improved authentication for security and federated provisioning
Given that a standardized API will be implemented and consumed by more parties, possibly accessed over the open Internet, does it make sense to allow and/or standardize on other auth mechanisms (i.e. such as OAuth?). The Spec currently specifies Basic Auth as the only mechanism for authentication.

**EDIT by @avade**

Working document: https://docs.google.com/document/d/1XLPMLJbf_lRXIdMg1XUyKNZUd2--fDROGvdEm25tMjc/edit

### LINKED ISSUE #146: Make it clear that auth/creds for brokes are optional

duglin:
This should now leave it open for us to add more text to the
Profile doc explaining how to do oAuth, or any other OPTIONAL
auth mechanism.

Closes #146

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -109,8 +109,22 @@ This header allows brokers to reject requests from marketplaces for versions the
 
 ## Authentication
 
-The marketplace MUST authenticate with the service broker using HTTP
-basic authentication (the `Authorization:` header) on every request. The broker is responsible for validating the username and password and returning a `401 Unauthorized` message if credentials are invalid. It is RECOMMENDED that brokers support secure communication from platform marketplaces over TLS.
+While the communication between a platform and broker
+MAY be unsecure, it is RECOMMENDED that all communications between a
+platform and a broker are secured via TLS and authenticated.
+
+Unless there is some out of band communication and agreement between a
+platform and a service broker, the marketplace MUST authenticate with the
+service broker using HTTP basic authentication (the `Authorization:` header)
+on every request. This specification does not specify how platform and brokers
+agree on other methods of authentication.
+
+If authentication is used, the broker MUST authenticate the request using the
+predetermined authentication mechanism and MUST return a `401 Unauthorized`
+response if the authentication fails.
+
+Note: using an authentication mechanism that is agreed to via out of band
+communications could lead to interoperability issues with other platforms.
 
 ## Catalog Management
 


spec.md
```diff
@@ -100,8 +100,19 @@ This header allows brokers to reject requests from marketplaces for versions the
...
 
-The marketplace MUST authenticate with the service broker using HTTP
-basic authentication (the `Authorization:` header) on every request. The broker is responsible for validating the username and password and returning a `401 Unauthorized` message if credentials are invalid. It is RECOMMENDED that brokers support secure communication from platform marketplaces over TLS.
+While the communication between a platform and broker
+MAY be unsecure, it is RECOMMENDED that all communications between a
+platform and a broker are secured via TLS and authenticated.
+
+Unless there is some out of band communication and agreement between a
+platform and a service broker, the marketplace MUST authenticate with the
+service broker using HTTP basic authentication (the `Authorization:` header)
```
bparees: the authorization header and basic auth aren't the same thing, contrary to what this seems to imply.  Seems like this should say "the marketplace must authenticate w/ the service broker using the authorization header.  The value of that header MAY be basic auth" (or it may be something else, which i think is what we're looking for here).

duglin: @bparees Are you focused on the linking of basic-auth with the authZ header?  If so, I think that's ok for this paragraph because we're saying what the default is - basic auth via authZ header.  If you want to separate out the header name discussion from the header value discussion, can we do that? Can we really mandate that all auth mechanisms MUST use the Authorization header? I would think that it would be up to the spec of that mechanism to decide that.

If you're talking about something else then can you elaborate because I think your proposed text says the same thing as what's in the PR (and in the spec today).

bparees: sorry, you're correct, I overlooked the semantics of the opening clause ("unless ......").  The contract is what i'd expect, allowing for non-basic-auth exchanges between the SC and Broker.

spec.md
```diff
@@ -100,8 +100,19 @@ This header allows brokers to reject requests from marketplaces for versions the
...
+on every request. This specification does not specify how the platform
+and brokers commnicate this out of band agreement.
```
shalako: Suggestion for last sentence: "This specification does not specify how platform and brokers agree on other methods of authentication."

shalako: I also want to make it clear that using methods of auth other than basic auth may make brokers incompatible with some platforms.

duglin: @shalako I picked up your new wording. On the 2nd comment... what kind of text are you looking for?  I'm having a hard time coming up with text that doesn't imply basicAuth is preferred, and I'm not sure that's what we want.

spec.md
```diff
@@ -109,8 +109,19 @@ This header allows brokers to reject requests from marketplaces for versions the
...
+platform and a broker are secured via TLS and authenticated.
+
+Unless there is some out of band communication and agreement between a
+platform and a service broker, the marketplace MUST authenticate with the
+service broker using HTTP basic authentication (the `Authorization:` header)
+on every request. This specification does not specify how platform and brokers
+agree on other methods of authentication.
+
+Regardless of the authentication mechanism that is used, the broker MUST
+validate the credentials from the platform and MUST return a
```
avade: "validate the credentials" - are certificates/tokens considered credentials? 

Perhaps "authenticate the request using the predetermined authentication mechanism" is clearer?

duglin: ok with me

duglin:
I've updated the top comment to remove the ref to #103 since I have a separate AI to create the profile text that explains how to use OAuth.  For now, I think this PR is ready for review since it just focuses on making it clear that auth (at all) and basicAuth (as the mechanism) is optional but recommended and the default.  So this PR can be reviewed/merged on its own.

mattmcneeney:
This looks good to me @duglin. We'll just have to make sure that, when the time comes, we add details of other supported auth mechanisms for each platform the the `profiles.md` file.

shalako:
@duglin Can we please have a sentence that warns that using an out-of-bound auth mechanism with a platform is a sure road to incompatibility with other platforms?

duglin:
@shalako done - see if the Note I added works for you.

shalako:
LGTM

spec.md
```diff
@@ -109,8 +109,22 @@ This header allows brokers to reject requests from marketplaces for versions the
...
 ## Authentication
 
-The marketplace MUST authenticate with the service broker using HTTP
-basic authentication (the `Authorization:` header) on every request. The broker is responsible for validating the username and password and returning a `401 Unauthorized` message if credentials are invalid. It is RECOMMENDED that brokers support secure communication from platform marketplaces over TLS.
+While the communication between a platform and broker
+MAY be unsecure, it is RECOMMENDED that all communications between a
+platform and a broker are secured via TLS and authenticated.
+
+Unless there is some out of band communication and agreement between a
+platform and a service broker, the marketplace MUST authenticate with the
```
liggitt: It reads oddly to pair a MUST with a RECOMMENDED. It also means that to be spec compliant, by default I must be willing to send basic auth credentials over http? That makes me cringe.

duglin: I'm not sure this is a MUST+RECOMMEND - where do you see that?

It doesn't mean you MUST support basic auth, it just says you MUST support it IF you don't have some other mechanism agreed to between you and the brokers.

duglin:
this could use some LGTMs:  @pmorie @mattmcneeney @vaikas-google @avade @angarg12

angarg12:
LGTM

avade:
lgtm

mattmcneeney:
This change will allow platforms to start testing out non-BasicAuth workflows, which seems to be of interest to many people, so LGTM.

Once platforms start working on this, we should provide feedback and PR any further changes into the spec and the profiles file, to ensure broker authors can easily add support for this.

mattmcneeney:
Accidentally merged, fixing now...

PR #233, opened 2017-06-15
Markdown can not have a class

jpeeler:
The application of the note class makes the markdown link text render as plain
text and not as a link.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -293,7 +293,7 @@ Brokers that support sychronous responses for provision, update, and delete can 
 
 ### Asynchronous Operations
 
-<p class="note">Note: Asynchronous operations are currently supported only for provision, update, and deprovision.</p>
+Note: Asynchronous operations are currently supported only for provision, update, and deprovision.
 
 For a broker to return an asynchronous response, the query parameter `accepts_incomplete=true` MUST be included the request. If the parameter is not included or is set to `false`, and the broker cannot fulfill the request synchronously (guaranteeing that the operation is complete on response), then the broker SHOULD reject the request with the status code `422 Unprocessable Entity` and the following body:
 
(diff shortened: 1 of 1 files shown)


duglin:
Did you push the right change? It doesn't seem to match the text of the issue

jpeeler:
I did push the correct change. The markdown link "[Catalog](#catalog-management)" was removed, which was tagged with the paragraph element and note class. Maybe it's actually just being content in a paragraph that doesn't cause the link to show up, not sure.

duglin:
Ah ok I see now.  The anchor is still there (see line 107, in my copy anyway).  I think perhaps the right fix is to remove the <p> tag instead.

mattmcneeney:
Hey @jpeeler. Is this problem arising when pulling these docs into platform-specific docs?

jpeeler:
@mattmcneeney The problem is when rendered on github, though I assume the same would apply with any markdown rendering elsewhere. https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#binding

mattmcneeney:
@jpeeler so it is, I didn't realise that. We'd had some trouble sync'ing these docs with other tools so just wanted to make sure it wasn't another problem along those lines.

This change looks good to me.

duglin:
I don't think we should remove the href, let's remove the \<p\> and class instead. Its odd that we have a class w/o its definition anyway.

shalako:
I'm ok with removing the `<p>` but then we should do it for another one in the spec also.

mattmcneeney:
I'm OK with this too. This was for the CF docs, but we are moving to a solution where this won't be required.

duglin:
@jpeeler you ok with making the proposed change? If you don't have time I can do it.

jpeeler:
@duglin the latest commit made the change, I think it's good to go.

duglin:
@jpeeler I think there are 4 \<p\> / Notes in the spec that would fall into this category, do you want to do the others too?

jpeeler:
Sorry didn't realize you wanted them all gone, but now they are!

duglin:
excellent - thanks!
LGTM

shalako:
LGTM

angarg12:
LGTM

duglin:
need one more easy L.G.T.M

duglin:
@pmorie?

avade:
lgtm

PR #234, opened 2017-06-15
Change example to match field order of request body

jpeeler:
This doesn't have to be merged, but slightly helps readability I think.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -415,12 +415,12 @@ The `:instance_id` of a service instance is provided by the platform. This ID wi
 
 ```
 {
+  "service_id": "service-guid-here",
+  "plan_id": "plan-guid-here",
   "context": {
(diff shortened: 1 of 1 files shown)


duglin:
If we do reorder things,i think I'd prefer to change the table so that context is first mainly because to me it's like global info (almost like a header) so putting it first kind of separates it a bit from the info of the real data of the request.

mattmcneeney:
The other fields are also getting deprecated at some point in the future, but I'm happy with reordering things for readability.

shalako:
LGTM

Service and plan ids at the top work for me.

pmorie:
LGTM

angarg12:
LGTM

shalako:
LGTM

duglin:
LGTM

PR #237, opened 2017-06-19
Use full URLs in profile.md

mattmcneeney:
CF docs pull this file in automatically, however the links don't work unless they are absolute. This will fix this issue.

### DIFF
--- a/profile.md
+++ b/profile.md
@@ -2,7 +2,7 @@
 
 ## Abstract
 
-The [Open Service Broker API specification](spec.md) allows for extensions
+The [Open Service Broker API specification](https://github.com/openservicebrokerapi/servicebroker/blob/v2.12/spec.md) allows for extensions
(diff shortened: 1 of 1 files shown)


mattmcneeney:
@shalako @duglin @vaikas-google @pmorie @angarg12 
Would appreciate help getting this merged (if no objections) so we can get our CF docs updated :)

angarg12:
LGTM

shalako:
Just want to confirm we're going to start linking out to the OSBAPI repo from CF docs rather than use relative links...

mattmcneeney:
Yep that's right. The other option would be to link to profile.md.html.erb,
but something tells me that isn't a good idea!

mattmcneeney:
As per @duglin's comment on the other similar PR, this isn't a suitable solution, so closing.

PR #238, opened 2017-06-19
Use absolute URLs for links to the spec

mattmcneeney:
I just submitted a PR to fix this on the v2.12 branch. This is to do the same on master, and so when we next cut a release, we have a simple change to make (swap `master` for `v...`). Thanks!

### DIFF
--- a/profile.md
+++ b/profile.md
@@ -2,7 +2,7 @@
 
 ## Abstract
 
-The [Open Service Broker API specification](spec.md) allows for extensions
+The [Open Service Broker API specification](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md) allows for extensions
 and variations based on the environments in which it is being used; this
 document contains the recommended usage pattern for some of those variants.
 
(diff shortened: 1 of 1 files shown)


mattmcneeney:
@shalako @duglin @vaikas-google @pmorie @angarg12
Same again, this time on master!

shalako:
Relative links won't work? I'm wondering if we can continue the having the CF docs link to each other, or if we have to start linking out to the OSBAPI docs.

mattmcneeney:
So this change will make the CF docs link out to the repository. We could
consider just linking directly out to the spec.md from the side menu and
not doing any syncing at all.

duglin:
I'm confused as to why relative links don't work.

duglin:
This change will break any other fork/branch, right?

mattmcneeney:
Good point @duglin, didn't think about that. This isn't an acceptable solution. Closing these PRs, will research another way!

duglin:
I'm not 100% sure what issue this was trying to resolve, but I get the sense it was related to other issues we've seen where the CF docs would be better served by having some kind of pre-processor in place to tweak things to deal with CF specific needs.  Is that correct?

mattmcneeney:
Yep, that is correct!

PR #242, opened 2017-06-20
moving content for volume mounts from cf docs into spec

shalako:
The spec for the `volume_mounts` field in the bind response was documented as a separate file, then removed from the CF docs by accident so it was missed when we migrated the spec. 

We have since added the doc back to the CF docs: http://docs.cloudfoundry.org/services/volume-services.html. Once this is merged, we'll remove the doc from the CF docs. I will submit a PR to the 2.12 branch as well.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -640,9 +640,9 @@ did not include a `"requires":["route_forwarding"]` property.
 
 There are a class of services that provide network storage to applications
 via volume mounts in the application container. A create binding response
-from one of these services MUST include a `volume_mount`.
+from one of these services MUST include `volume_mounts`.
 
-Brokers MUST NOT include a `volume_mounts` in a create binding response
+Brokers MUST NOT include `volume_mounts` in a create binding response
 if the associated [Catalog](#catalog-management) entry for the service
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -640,9 +640,9 @@ did not include a `"requires":["route_forwarding"]` property.
 
 There are a class of services that provide network storage to applications
 via volume mounts in the application container. A create binding response
-from one of these services MUST include a `volume_mount`.
+from one of these services MUST include `volume_mounts`.
```
pmorie: I'm cool with moving this information in, but we might want to add a note saying that for route and volume services, users should check to see that their platforms support those services.

shalako: I'd like such a note to be accompanied by a link to a section in profile.md that specifies what features are (or are not) supported by example platforms. Can we do that in a separate PR?

duglin:
LGTM

angarg12:
LGTM

avade:
lgtm

duglin:
one more l.g.t.m needed

zrob:
LGTM

PR #243, opened 2017-06-20
moving content for volume mounts from cf docs into 2.12

shalako:


spec.md
```diff
@@ -734,7 +734,25 @@ For success responses, the following fields are supported. Others will be ignore
...
 | route_service_url | string | A URL to which the platform MUST proxy requests for the address sent with `bind_resource.route` in the request body. `"requires":["route_forwarding"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the platform can consider the response invalid. |
-| volume_mounts | array-of-objects | An array of configuration for mounting volumes. `"requires":["volume_mount"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the platform can consider the response invalid. |
+| volume_mounts | array-of-objects | An array of configuration for remote storage devices to be mounted into an application container filesystem. `"requires":["volume_mount"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the platform can consider the response invalid. |
+
+##### Volume Mounts Object
+
+| Response Field | Type | Description |
+| --- | --- | --- |
+| driver | string | Name of the volume driver plugin which manages the device |
```
duglin: More often than not I think we end the sentences in the last column with a period.

shalako: A convention I picked up from the CF docs team is to only use a period after a sentence. I can go either way.

duglin: Just submitted a PR to check for stuff like this.

spec.md
```diff
@@ -734,7 +734,25 @@ For success responses, the following fields are supported. Others will be ignore
...
+| container_dir | string | The directory to mount inside the application container |
```
duglin: This could be interpreted to mean "the **existing** directory in the container to do something with". I think this might be more clear: `The path in the application container onto which the volume will be mounted. Any existing resource at that path will not be visible to the container.`

spec.md
```diff
@@ -734,7 +734,25 @@ For success responses, the following fields are supported. Others will be ignore
...
+| mode | string | "r" to mount the volume read-only, or "rw" to mount it read-write |
+| device_type | string | A string specifying the type of device to mount. Currently the only supported value is "shared"  |
```
duglin: "device to mount"... I wonder if this should be "volume to mount" instead. Or can we use this to mount non-volume resources?

spec.md
```diff
@@ -734,7 +734,25 @@ For success responses, the following fields are supported. Others will be ignore
...
+| device | device-object | Device object containing device_type specific details. Currently the only supported value is "shared_device" |
+
+##### Shared Device Object
+
+| Field | Type | Description |
+| --- | --- | --- |
+| volume_id | string | ID of the shared volume to mount on every app instance |
```
duglin: `This REQUIRED field is the ID of...`

spec.md
```diff
@@ -734,7 +734,25 @@ For success responses, the following fields are supported. Others will be ignore
...
+| mount_config | object | Configuration object to be passed to the driver when the volume is mounted (optional) |
```
duglin: `This OPTIONAL field contains the configuration object to be passed to the driver when the volume is mounted.`

spec.md
duglin: `This REQUIRED field is the name of...`
Or we should use the asterisk convention used in the rest of the spec.

shalako: I have always felt that the asterisk mechanism is too easy to miss. Just never got around to changing it. I like the idea of either putting it in the description or maybe better, adding a required column to parameter/header/body tables.

duglin: I agree with leaving the possible move from asterisks to another PR, but I think this PR will need to use them since some of these fields are required within the scope of the optional objects.  Right?

spec.md
shalako: We're checking with the container management team to find out if the mechanism that bind-mounts a list of dirs checks whether the targets exist or not.

spec.md
julian-hj: ...not now, but eventually theoretically we might mount other kinds of devices like WORM storage, cameras, whatever.  To me "device" seems correct.

spec.md
duglin: On the period... we should add it to these descriptions since that's consistent with the rest of the doc.

spec.md
duglin: ok - if it turns out that this is just an impl detail of Kube and not true for everyone then only the first sentence of my recommended change should be picked up.  Or, better.... `This specification does not mandate what action the platform is to take if the path specified already exists in the container.`.  Although, I'd prefer if we did to provide a consistent UX.

spec.md
duglin: If that's the case, should we rename this section "device mounts" or something instead of "volumes" ?

spec.md
duglin: Based on above discussions, let's add an asterisk

spec.md
shalako: Unfortunately, at the top-level field name is already `volume_mounts` in the spec against which volume service brokers have been built, the ship may have sailed.

spec.md
shalako: Adding periods and asterisks...

shalako:
@duglin I updated the PR per your feedback

spec.md
julian-hj: My understanding from the Garden team is that bind mounts in Garden behave very similarly to bind mounts in k8s.  Any existing resource will get masked by the bind mount. So your initial suggestion looks correct to me.  Does it make sense to mandate that behavior in the spec though?  Maybe?  It appears to be standard linux bind mount behavior, but I wonder about Windows runtimes, and whether they behave the same way.

shalako: I have added Doug's suggested caveat: `This specification does not mandate what action the platform is to take if the path specified already exists in the container.`

duglin:
LGTM

angarg12:
LGTM

avade:
LGTM

zrob:
LGTM

PR #248, opened 2017-06-21
Better define what a "CLI-friendly string" is

mattmcneeney:
I don't mind if we want to add a generic definition for this and then change the field type for these two cases, but if everyone is happy with this, I'm fine with it.

Thanks to Laird Nelson for the note about this in slack.


shalako:
LGTM

duglin:
LGTM

zrob:
LGTM

shalako:
One more needed

angarg12:
LGTM

PR #263, opened 2017-07-10
Add section in spec for API routes.

### LINKED ISSUE #257: Allow paths as suffixes for broker request URLs
### Goal
We have several scenarios where brokers need to be exposed from endpoints that are not just a base hostname/port pair, but also contain path elements. We would like to be able to expose the broker API endpoints as path suffixes on an existing base path, and include the path as part of the broker URL when registering it with the platform.

### Current model
Currently brokers must be registered as root hosts, e.g.,

```
Broker: http://mybroker.mycompany.com:8080
Username: me
Password: supersecret
```

Calls must always be made to broker spec paths as the root path, e.g., `http://mybroker.mycompany.com:8080/v2/catalog`. This is because they are specified in the spec as HTTP calls like `GET /v2/catalog`.

### Proposed change
We would like to be able to register brokers including a base path, such as

```
Broker: http://mybroker.mycompany.com:8080/some/base/path
Username: me
Password: supersecret
```

Broker spec paths are then appended to whatever broker URI has been registered, e.g., `http://mybroker.mycompany.com:8080/some/base/path/v2/catalog`.

bmelville:
The intent of this section is to clarify requirements of the endpoint routes for the API.

Fixes #257.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -8,6 +8,7 @@
     - [Changes Since v2.11](#changes-since-v211)
   - [API Version Header](#api-version-header)
   - [Authentication](#authentication)
+  - [API Routes](#api-routes)
   - [Catalog Management](#catalog-management)
     - [Adding a Broker to the Platform](#adding-a-broker-to-the-platform)
   - [Synchronous and Asynchronous Operations](#synchronous-and-asynchronous-operations)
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -112,6 +113,16 @@ This header allows brokers to reject requests from marketplaces for versions the
 The marketplace MUST authenticate with the service broker using HTTP
 basic authentication (the `Authorization:` header) on every request. The broker is responsible for validating the username and password and returning a `401 Unauthorized` message if credentials are invalid. It is RECOMMENDED that brokers support secure communication from platform marketplaces over TLS.
 
+## API Routes
+
+All endpoint routes in the Service Broker API are defined as URL suffixes. The Service Broker API MAY be exposed from a common URL prefix. An individual broker MAY also choose to expose multiple Service Broker APIs from different URL prefixes, but each MUST have a unique URL prefix.
```
duglin: Perhaps a slight rewording to make it clear that they can't change prefixes across APIs:
```
All endpoint routes defined in this specification are to be interpreted as URL suffixes.
When a broker is registered with a platform, and provides a URL prefix, all supported
endpoints MUST be based on that URL prefix. In other words, within the scope of a
single broker's use for a platform, the URL prefix MUST NOT change.
```
I think we can stop there, but if people want to be very explicit we could add something like:
```
While a broker MAY be registered to multiple platforms, or even if it is registered to the
same platform more than once,  each registration MUST be treated as a different broker
for the purpose of the URL prefix/suffix scoping.
```

mattmcneeney: @bmelville did you want to change anything following @duglin's feedback above?

bmelville:
Yes this fell off the priority list for the last two weeks, but I plan to follow up shortly.

duglin:
tap tap tap

avade:
1-week closing warning.

n3wscott:
I think this PR is not needed. The spec says `service-broker-url` not `service-broker-host`. `service-broker-url` implies that any sub-resource may provide an OSBAPI.

PR #283, opened 2017-07-31
Revert "Make using auth/tls optional and allow for non-BasicAuth auth"

mattmcneeney:
Reverts openservicebrokerapi/servicebroker#223

Whoops, that didn't take long. Forgot that we shouldn't merge things that are still being validated! My bad.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -109,22 +109,8 @@ This header allows brokers to reject requests from marketplaces for versions the
 
 ## Authentication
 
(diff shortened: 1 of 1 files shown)


mattmcneeney:
Speaking with Alex, we think this is actually OK to go into the spec as it isn't adding any new features. New auth workflows will be added in a new PR; this is just a wording change to the spec allowing other workflows, but no platforms support these at this time.

PR #289, opened 2017-08-09
add context object to binding calls

mattmcneeney:
We have some use cases where knowing contextual information would be useful when creating a service binding.

What do y'all think of this?

### DIFF
--- a/profile.md
+++ b/profile.md
@@ -130,6 +130,7 @@ The following table specifies which properties MUST appear in each API:
 | --- | --- |
 | `PUT /v2/service_instances/:instance_id` | `organization_guid`<br>`space_guid` |
 | `PATCH /v2/service_instances/:instance_id` | `organization_guid`<br>`space_guid` |
+| `PUT /v2/service_instances/:instance_id/service_bindings/:binding_id` | `organization_guid`<br>`space_guid` |
 
 Example:
 
@@ -168,6 +169,7 @@ The following table specifies which properties MUST appear in each API:
 | --- | --- |
 | `PUT /v2/service_instances/:instance_id` | `namespace` |
 | `PATCH /v2/service_instances/:instance_id` | `namespace` |
+| `PUT /v2/service_instances/:instance_id/service_bindings/:binding_id` | `namespace` |
 
 Example:
 
--- a/spec.md
+++ b/spec.md
@@ -795,6 +795,7 @@ it to correlate the resource it creates.
 
 | Request Field | Type | Description |
 | --- | --- | --- |
+| context | object | Contextual data under which the service binding is created. |
 | service_id* | string | ID of the service from the catalog. MUST be a non-empty string. |
 | plan_id* | string | ID of the plan from the catalog. MUST be a non-empty string. |
 | app_guid | string | Deprecated in favor of `bind_resource.app_guid`. GUID of an application associated with the binding to be created. If present, MUST be a non-empty string. |
@@ -828,6 +829,10 @@ binding requests.
 
 ```
 {
+  "context": {
+    "platform": "cloudfoundry",
+    "some_field": "some-contextual-data"
+  },
   "service_id": "service-id-here",
   "plan_id": "plan-id-here",
   "bind_resource": {
@@ -844,6 +849,10 @@ binding requests.
 #### cURL
 ```
 $ curl http://username:password@broker-url/v2/service_instances/:instance_id/service_bindings/:binding_id -d '{
+  "context": {
+    "platform": "cloudfoundry",
+    "some_field": "some-contextual-data"
+  },
   "service_id": "service-id-here",
   "plan_id": "plan-id-here",
   "bind_resource": {


duglin:
not that I see a problem with it but I'm curious to know why org/space are not part of that message exchange already?

mattmcneeney:
I'm not sure why. @zrob any ideas?

mattmcneeney:
Does anyone else have any thoughts on this?

pmorie:
We should send this on bind, I am +1 on this.

mattmcneeney:
If folks are happy with this, please can we give it the LGT_
@duglin @pmorie @shalako @avade @vaikas-google @angarg12

angarg12:
LGTM

mattmcneeney:
The implementation of this is now being validated. I will post feedback when I have it, at which point we can begin the review process.

profile.md
```diff
@@ -130,6 +130,7 @@ The following table specifies which properties MUST appear in each API:
 | --- | --- |
 | `PUT /v2/service_instances/:instance_id` | `organization_guid`<br>`space_guid` |
 | `PATCH /v2/service_instances/:instance_id` | `organization_guid`<br>`space_guid` |
+| `PUT /v2/service_instances/:instance_id/service_bindings/:binding_id` | `organization_guid`<br>`space_guid` |
```
duglin: Let's add the appropriate stuff for Kube too, which would be the "namespace" just like all of the other calls.

gberche-orange:
@mattmcneeney related to cloudfoundry's future support for sharing services and our conversation onto  this [cf-dev@ thread](http://cf-dev.70369.x6.nabble.com/Re-cf-dev-Re-Proposal-Sharing-service-instances-across-orgs-and-spaces-tp7101p7819.html), the `organization_guid` and `space_guid` in the `PUT /v2/service_instances/:instance_id/service_bindings/:binding_id`  call represent where the binding is requested. 

In order for a broker to reject a binding requested from a shared service instances, it needs to maintain state between the `PUT /v2/service_instances/:instance_id` and the `PUT /v2/service_instances/:instance_id/service_bindings/:binding_id` calls (to compare the `organization_guid` and `space_guid` fields provided in each context).

Adding new "hint" fields to the binding context, (such as `service_instance.{organization_guid= space_guid=}`) would help support stateless brokers that need to reject binding from service instances shared across spaces.

profile.md
mattmcneeney: Done!

mattmcneeney:
@gberche-orange I agree with your proposal for supporting stateless brokers, however this is probably a spec-wide change so I think this should be tackled in a separate issue. What do you think?

duglin:
LGTM

gberche-orange:
@mattmcneeney while there are some existing similar hints such as in https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#parameters-3 to reduce statefulnesss across OSB API calls without involving spec-wide changes, in this specific case the additional hints on `organization_guid` and `space_guid` would need to be namespaced, and this would deserve a more general discussion to provide consistency across "cross-calls hints". 

I looked for existing issues and PR and could only find stateless support for specific parts of the spec, or broker state themselves https://github.com/openservicebrokerapi/servicebroker/issues?utf8=%E2%9C%93&q=%20is%3Aopen%20stateless%20 

Would you open a spec-wide issue ?

Samze:
Would a `context` field on the bind request be redundant or at least confusing with the existing [`bind_resource`](https://github.com/openservicebrokerapi/servicebroker/blob/v2.12/spec.md#bind-resource-object) field? The bind resource is described as an object containing "platform specific information related to the context in which the service will be used.", and the example fields in this object are `app_guid` and `route` but the spec states that "Platforms MAY choose to add additional ones as needed."

Additionally, the spec states `bind_resource.app_guid` will be a kubernetes namespace. Adding a context object would duplicate this information. 

If we do decide to have both `bind_resource` & `context` then we should clarify the descriptions.

Sam & @jenspinney

mattmcneeney:
@samze @jenspinney Good spot. I'd still support having a context object for consistency but it does raise a good point as to what we should do with the bind_resource object and its fields.

@duglin Do you have any thoughts on this?

duglin:
I agree with @mattmcneeney on this. I think for consistency it would be good to place the org/space/namespace info in the same spot for all messages that need it and I view that information as broader than info related to the app (or the thing to which we will bind the service).

`bind_resource` is the entity to which we will bind the service. This means, to me, that while it may share some of the same properties as `context` its scope is a bit different.  At a minimum it will include the app_guid which is not part of `context`.  The fact that Kube uses the namespace as the `app_guid` is just a coincidence and could change in the future.

Let me offer a scenario that we don't support today (at least i don't think we do) to explain why I think there's a difference between `bind_resource` and `context`.... let's say we create a service in org/space called john/dev. In that case the context will have john/dev in it.  Now, let's suppose that we support the idea of binding that service instance to an app running in mary/prod (yes I know we can't do this today in CF and the "sharing idea" might solve it differently, but indulge me :-) ).  In that case I can image the `bind_resource` one day having its own org/space (mary/prod in this case) that tells us where the app is because it might be in a different location from the service instance.

mattmcneeney:
Interestingly enough we are currently working on just that scenario! Thanks
for the feedback.

mattmcneeney:
@duglin I have changed some of the text to try to make it clearer what `context` and `bind_resource` are used for. Could you take a look and let me know what you think?

cc @utako @jenspinney

duglin:
rebase needed

mattmcneeney:
Rebased and ready @duglin @pmorie @shalako @avade @vaikas-google @angarg12

zrob:
LGTM

pmorie:
LGTM

vaikas-google:
LGTM

shalako:
LGTM

mattmcneeney:
Woohoo! Thanks all!

PR #293, opened 2017-08-14
Provide more clarity around the scope of app_guid

### LINKED ISSUE #183: Scope/use of org, space and appGUID
While we've agreed that org and space are too CF specific, and are in the process of replace them with "context", appGUID and CF's use of these three fields are things we should probably discuss.

In particular, CF will not allow the same app to bind to the same service instance more than once. Because of this, today's service brokers do not need to worry about things like seeing two different createBinding() requests that share the same org, space and appGUID.  The question is, should they care? Should the spec ban this situation?

I can't really think of a good reason why the spec should codify CF's current usage pattern with respect to this.  I think it might be fair, for some reason I can't think of right now, for a broker to reject two requests with the same values for org/space/appGUID - but that's an impl choice.

I'd like to see the spec add some text around this, something that says:
- the spec places no restrictions on how many bindings for the same instance an app can have
- brokers are free to use context & appGUID for whatever purpose they want. Including contacting the platform and using those fields to specify the context of the request, or using it to limit the usage pattern (e.g. only one binding per app)

Related to this though is the question of appGUID.  Right now the spec doesn't really place any restrictions on this property, ie. must it be unique? And it doesn't really define what an "app" is either.  

I suspect we probably can't be too prescriptive here since an "app" may mean different things in different platforms.  In CF is clear. In Kube its less clear.  It could mean a set of replicated images/containers, or it could mean a set of unrelated containers that just happen to share the same label. Or it could even be a namespace. The scope to which the binding applies may need to remain open.

My proposal for this would be to add text that says something like:
- "app" is meant to represent a platform specific scope to which a binding applies. And give some examples like CF's app, Kube's Deployments, Kube's namespaces....  The definition of what an app's scope is platform specific and not required to be static across all binding requests.  For example, in one it might refer to a namespace, but in another it might refer to a Deployment. Its a platform impl choice.
- app_guid MUST be unique within the platform

Thoughts?

duglin:
Closes #183

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -123,7 +123,7 @@ If authentication is used, the broker MUST authenticate the request using the
 predetermined authentication mechanism and MUST return a `401 Unauthorized`
 response if the authentication fails.
 
-Note: using an authentication mechanism that is agreed to via out of band
+Note: Using an authentication mechanism that is agreed to via out of band
 communications could lead to interoperability issues with other platforms.
 
 ## Catalog Management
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -123,7 +123,7 @@ If authentication is used, the broker MUST authenticate the request using the
 predetermined authentication mechanism and MUST return a `401 Unauthorized`
 response if the authentication fails.
 
-Note: using an authentication mechanism that is agreed to via out of band
+Note: Using an authentication mechanism that is agreed to via out of band
```
duglin: Just noticed the other `Note:`s start with a cap.

jmrodri:
LGTM

fmui:
LGTM

angarg12:
LGTM

duglin:
2 more reviews needed: @avade @mattmcneeney @pmorie @shalako @vaikas-google @zrob

mattmcneeney:
LGTM

avade:
LGTM

PR #299, opened 2017-08-29
Add warning text about using weird chars in IDs

### LINKED ISSUE #273: ServiceId and PlanID - valid chars
ServiceID and PlanID are defined as a non-empty strings w/o any restrictions on the exact set of characters that can be used - however, the spec does RECOMMEND a GUID be used.

This leaves the question of whether the intent was to really leave it totally wide open for any characters (including things like forward/back-slashes, punctuation and spaces).

If in Kubernetes we were to use these IDs as the primary look-up key then we'd run into issues because the rules in Kube for these fields are: should be up to maximum length of 253 characters and consist of lower case alphanumeric characters, -, and .

Question for OSB API group to consider: can/should we be more specific about the valid character set of these IDs w/o requiring a version bump in the spec?

duglin:
Closes #273

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -8,6 +8,7 @@
     - [Changes Since v2.11](#changes-since-v211)
   - [API Version Header](#api-version-header)
   - [Authentication](#authentication)
+  - [URL Properties](#url-properties)
   - [Catalog Management](#catalog-management)
     - [Adding a Broker to the Platform](#adding-a-broker-to-the-platform)
   - [Synchronous and Asynchronous Operations](#synchronous-and-asynchronous-operations)
@@ -125,6 +126,27 @@ response if the authentication fails.
 
 Note: Using an authentication mechanism that is agreed to via out of band
 communications could lead to interoperability issues with other platforms.
+
+## URL Properties
+
+This specification defines the following properties that might appear within
+URLs:
+- `service_id`
+- `plan_id`
+- `instance_id`
+- `binding_id`
+- `operation`
+
+While this specification places no restriction on the set of characters used
+within these strings, it is RECOMMENDED that these properties only contain
+characters from the "Unreserved Characters" as defined by
+[RFC3986](https://tools.ietf.org/html/rfc3986#section-2.3). In other words:
+uppercase and lowercase letters, decimal digits, hyphen, period, underscore
+and tilde.
+
+If a character outside of the "Unreserved Characters" set is used, then
+it SHOULD be percent-encoded prior to being used as part of the HTTP request,
+per [RFC3986](https://tools.ietf.org/html/rfc3986#section-2.1).
 
 ## Catalog Management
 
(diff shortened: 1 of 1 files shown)


angarg12:
LGTM

mattmcneeney:
LGTM. Nice one.

avade:
I like the way that you have explained WHY by stating the properties are used in URLs

LGTM

spec.md
```diff
@@ -126,6 +127,25 @@ response if the authentication fails.
...
+
+While this specification places no restriction on the set of characters used
+within these strings, it is RECOMMENDED that these properties only contain
+characters from the "Unreserved Characters" as defined by
+[RFC3986](https://www.ietf.org/rfc/rfc3986.txt). In other words: uppercase
+and lowercase letters, decimal digits, hypen, period, underscore and tilde.
+
+If a character outside of the "Unreserved Characters" set is used, then
+it SHOULD be percent-encoded prior to be used as part of the HTTP request,
+per [RFC3986](https://www.ietf.org/rfc/rfc3986.txt).
```
MHBauer: Is it possible to link directly to the section like this? 
https://tools.ietf.org/html/rfc3986#section-2.1
If that kind of non-pure-text link is inappropriate, can you add to the text direction to see specifically section 2.1?

duglin:
Addressed @mhbauer's minor URL comment. Another round of reviews needed.

spec.md
```diff
@@ -126,6 +127,27 @@ response if the authentication fails.
...
+- `service_id`
+- `plan_id`
+- `instance_id`
+- `binding_id`
+- `operation`
+
+While this specification places no restriction on the set of characters used
+within these strings, it is RECOMMENDED that these properties only contain
+characters from the "Unreserved Characters" as defined by
+[RFC3986](https://www.ietf.org/rfc/rfc3986.txt#section-2.3). In other words:
```
MHBauer: This link doesn't go to the section. It's the text spec, and the second link is the html spec.

duglin: fixed

angarg12:
LGTM

mattmcneeney:
LGTM

duglin:
2 more reviews needed

spec.md
```diff
@@ -126,6 +127,27 @@ response if the authentication fails.
...
+- `plan_id`
+- `instance_id`
+- `binding_id`
+- `operation`
+
+While this specification places no restriction on the set of characters used
+within these strings, it is RECOMMENDED that these properties only contain
+characters from the "Unreserved Characters" as defined by
+[RFC3986](https://tools.ietf.org/html/rfc3986#section-2.3). In other words:
+uppercase and lowercase letters, decimal digits, hypen, period, underscore
```
vaikas-google: s/hypen/hyphen/

duglin: fixed

vaikas-google:
LGTM

avade:
lgtm

angarg12:
LGTM

shalako:
LGTM

PR #300, opened 2017-08-29
Add clarity around concurrent updates

duglin:
From the f2f:
Proposal: In the “Blocking Operations” section change it to say “...that mutate the same resource…” instead of “...that act on the same set of resources...”

&& creating Binding is not necessary considered mutating an Instance, it’s a Broker’s choice

&& move “Blocking Operations” out from under Async

&& change 422 in createBinding response status code table to include this concurrency stuff

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -566,7 +566,8 @@ For a service broker to return an asynchronous response, the query parameter
 not included or is set to `false`, and the service broker cannot fulfill the
 request synchronously (guaranteeing that the operation is complete on
 response), then the service broker SHOULD reject the request with the status
-code `422 Unprocessable Entity` and the following body:
+code `422 Unprocessable Entity` and the following body
+(see [Service Broker Errors](#service-broker-errors)):
 
 ```
 {
@@ -589,11 +590,11 @@ progress of the operation.
 
 ## Blocking Operations
 
-Service brokers do not have to support concurrent requests that act on the
-same set of resources. If a service broker receives a request that it is not
+Service Brokers do not have to support concurrent requests that mutate the
+same resource.  If a service broker receives a request that it is not
 able to process due to other activity being done on that resource then the
 service broker MUST reject the request with a HTTP `422 Unprocessable
-Entity` error and the following body:
+Entity` error and the following body (see [Service Broker Errors](#service-broker-errors):
 
 ```
 {
@@ -801,7 +802,7 @@ $ curl http://username:password@service-broker-url/v2/service_instances/:instanc
 | 202 Accepted | MUST be returned if the service instance provisioning is in progress. This triggers the platform marketplace to poll the [Service Instance Last Operation Endpoint](#polling-last-operation) for operation status. Note that a re-sent `PUT` request MUST return a `202 Accepted`, not a `200 OK`, if the service instance is not yet fully provisioned. |
 | 400 Bad Request | MUST be returned if the request is malformed or missing mandatory data. |
 | 409 Conflict | MUST be returned if a service instance with the same id already exists but with different attributes. The expected response body is `{}`, though the description field MAY be used to return a user-facing error message, as described in [Service Broker Errors](#service-broker-errors). |
-| 422 Unprocessable Entity | MUST be returned if the service broker only supports asynchronous provisioning for the requested plan and the request did not include `?accepts_incomplete=true`. The expected response body is: `{ "error": "AsyncRequired", "description": "This service plan requires client support for asynchronous service operations." }`, as described below. |
+| 422 Unprocessable Entity | MUST be returned if the service broker only supports asynchronous provisioning for the requested plan and the request did not include `?accepts_incomplete=true`. The expected response body is: `{ "error": "AsyncRequired", "description": "This service plan requires client support for asynchronous service operations." }`, as described below (see [Service Broker Errors](#service-broker-errors). |
 
 Responses with any other status code will be interpreted as a failure. Service brokers can include a user-facing message in the `description` field; for details see [Service Broker Errors](#service-broker-errors).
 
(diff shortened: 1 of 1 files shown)


angarg12:
LGTM

jmrodri:
@duglin PR title has a TYPO: conurrent should be concurrent

duglin:
@jmrodri fixed!  thanks

avade:
LGTM

mattmcneeney:
LGTM

duglin:
just one more review needed

jmrodri:
Not sure I count but LGTM :)

duglin:
Added text about orphan mitigation per the weekly meeting.

duglin:
I changed the error code from 422 to "409 Conflict" - seemed more appropriate, but what do others think?

jmrodri:
@duglin `409 Conflict` does sound more appropriate.  👍

angarg12:
LGTM

duglin:
reviews needed

spec.md
```diff
@@ -782,7 +789,7 @@ $ curl http://username:password@broker-url/v2/service_instances/:instance_id/ser
...
 | 400 Bad Request | MUST be returned if the request is malformed or missing mandatory data. |
 | 409 Conflict | MUST be returned if a service binding with the same id, for the same service instance, already exists but with different parameters. The expected response body is `{}`, though the description field MAY be used to return a user-facing error message, as described in [Broker Errors](#broker-errors). |
-| 422 Unprocessable Entity | MUST be returned if the broker requires that `app_guid` be included in the request body. The expected response body is: `{ "error": "RequiresApp", "description": "This service supports generation of credentials through binding an application only." }`. |
+| 422 Unprocessable Entity | MUST be returned if the broker requires that `app_guid` be included in the request body. The expected response body is: `{ "error": "RequiresApp", "description": "This service supports generation of credentials through binding an application only." }`. Additionally, if the broker rejects the request due to a concurrent request to create a binding for the same service instance, then this error MUST be returned (see [Blocking Operations](#blocking-operations)). |
```
vaikas-google: Above, we return 409 conflict when there are concurrent operations ongoing, but here we do 422 which seems a bit inconsistent.

duglin: damn - I'll fix that later today

duglin: fixed

duglin:
updated and ready for reviews

mattmcneeney:
LGTM

angarg12:
LGTM

spec.md
```diff
@@ -455,22 +456,28 @@ If the query parameter described above is present, and the broker executes the r
...
   "description": "Another operation for this service instance is in progress"
 }
 ```
 
+Note that per the [Orphans](#orphans) section, this error response does not
+cause orphan mitigation to be initiated. Therefore, platforms receiving
+this error response SHOULD resend the request at a later time.
+
+Brokers MAY choose to treat the creation of a binding as a mutation of
+the corresponding service instance - it is an implementation choice.
```
duglin: Add text about how this might cause platforms to serialize binding create() requests.

duglin:
ready for next round of reviews

mattmcneeney:
LGTM

fmui:
The proposal defines a fixed error body. A broker can neither change the description nor add additional data that might be useful to identify why there is a conflict. Can we make this a bit more flexible?

duglin:
@fmui See:  https://github.com/duglin/servicebroker/blob/6cbba3d423f3d7e5ebadd2e650f245ce531e1e2f/spec.md#broker-errors  it says: "For error responses, the following fields are valid. Others will be ignored. ".  While it would be nice if it was a bit clearer, I view that as saying the broker may include other fields (extensions) but they might be ignored by the platform.  Does that cover it?

fmui:
Not really. The proposal says:
"...then the broker **MUST reject** the request with an HTTP `409 Conflict` **and the following body:** ..."

This is much more restrictive than defined in the Broker Error section. It doesn't even allow to return a different description text. I don't see a reason why this error should have stricter rules than other errors.

duglin:
I don't think it was meant to be more restrictive. I think an generic "error PR" to make us more consistent would be good to answer these concerns. In particular:
- which fields in the error json MUST be as exactly specified or can be tweaked (e.g description)
- are additional fields (extensions) allowed and how are they added? (wrapper or top-level)

and we should touch all error text to ensure consistency.

vaikas-google:
+1 to allowing for additional fields. We have already run into some issues because brokers are built on platforms that provide some fixed fields for all error responses.

duglin:
@fmui can we clarify all of this in a follow-on PR? I'd prefer to keep this one more focused on just this one usecase.

zrob:
I have a few concerns here.

1. changing the MUST from 422 to 409 is a backwards incompatible change
1. similar to fmui concern, this feels like a very restrictive error message, where brokers may want to include more or different details that could be helpful in resolving the issue
1. I don't understand the benefits of making this change.  I can see that 409 may be somewhat more correct, but do we have platforms that will perform a different behavior based on this?  As long as we are returning an error and a useful message I don't see a reason to be more prescriptive than we already are.

duglin:
I reverted back to 422 and decided to add the "extensibility" text into this PR instead of creating a new one.  See what people think.

mattmcneeney:
@duglin merge conflict!

duglin:
rebased

mattmcneeney:
This is going to give me merge conflict fun times in the async bindings PR. But LGTM anyway!

[Approved with PullApprove](https://pullapprove.com/openservicebrokerapi/servicebroker/pull-request/300/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=openservicebrokerapi/servicebroker)

angarg12:
LGTM

[Approved with PullApprove](https://pullapprove.com/openservicebrokerapi/servicebroker/pull-request/300/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=openservicebrokerapi/servicebroker)

fmui:
LGTM

avade:
lgtm

PR #307, opened 2017-09-01
Fix some scoping text

duglin:
- plan_id's are globally unique, not scoped to the service
- I also fixed it so that we don't place requirements on the usage of IDs
  since that isn't necessary. Only the creation of the ID should state
  things like "its globally unique" or "must be non-empty string".
  The users of that ID can't control
  that aspect of it. However, we can control where the ID comes from so
  I added text to say things like "it MUST be the ID of a previously
  provisioned service".

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -490,8 +490,8 @@ The request provides these query string parameters as useful hints for brokers.
 
 | Query-String Field | Type | Description |
 | --- | --- | --- |
-| service_id | string | ID of the service from the catalog. If present, MUST be a non-empty string. |
-| plan_id | string | ID of the plan from the catalog. If present, MUST be a non-empty string. |
+| service_id | string | If present, it MUST be the ID of the service being used. |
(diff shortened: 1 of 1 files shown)


angarg12:
The OSBAPI website seems to be down and that is making the build fail. Should we move images to this repo?

angarg12:
LGTM

duglin:
we need to switch our OSB API URLs to be http instead of https.

duglin:
Rebased and ready for another round of reviews

angarg12:
LGTM

mattmcneeney:
LGTM

duglin:
rebased and ready for reviews

angarg12:
LGTM

avade:
LGTM

duglin:
2 more reviews needed

mattmcneeney:
LGTM

duglin:
one more review needed

pmorie:
LGTM

PR #326, opened 2017-09-27
fix readme formatting

mattmcneeney:
Missing newlines are making the readme look bad, and fix the release notes link to point to the latest release

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,13 +4,13 @@
 
 ## Latest Release: v2.13
 [Read the specification](https://github.com/openservicebrokerapi/servicebroker/blob/v2.13/spec.md)
+
 [Read the profile](https://github.com/openservicebrokerapi/servicebroker/blob/v2.13/profile.md)
 
-[Release Notes for past versions](release-notes.md)
+[Release Notes for past versions](https://github.com/openservicebrokerapi/servicebroker/blob/v2.13/release-notes.md)
 
 ## Working Changes
-[Read the working changes to the spec](spec.md)
-[Read the working changes to the profile](profile.md)
+[Read the working changes to the documentation](https://github.com/openservicebrokerapi/servicebroker/tree/master)
 
 [Roadmap and release planning](https://github.com/openservicebrokerapi/servicebroker/projects/1)
 


README.md
```diff
@@ -4,13 +4,15 @@
...
 
-[Release Notes for past versions](release-notes.md)
+[Release Notes for past versions](https://github.com/openservicebrokerapi/servicebroker/blob/v2.13/release-notes.md)
 
 ## Working Changes
-[Read the working changes to the spec](spec.md)
-[Read the working changes to the profile](profile.md)
+[Read the working changes to the spec](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md)
+
+[Read the working changes to the profile](https://github.com/openservicebrokerapi/servicebroker/blob/master/profile.md)
```
duglin: I'm ok with the new lines, but I'd prefer to keep these relative so that local traversals work.

mattmcneeney: That does mean though if somebody gets linked to v2.13 of the spec, then those links will show v2.13 of spec/profile, rather than master. Are we happy with that?

duglin: What if instead of pointing to both docs individually (which could be kind of annoying since the # of docs may change after we cut a release) we just do something like:

[Read the latest working changes of the documentation.](https://github.com/openservicebrokerapi/servicebroker/)

mattmcneeney: But then if I'm on the v2.13 branch, then I think that link would take me to a readme that looked exactly the same 😆

duglin: yes, the latest doc would always point to itself if there's a link to the latest doc :-)

mattmcneeney:
@duglin took your advice. How's this now?

angarg12:
The build fails with broken links to v2.13

e.g.

> tools/../_spec.md: Can't load url: https://github.com/openservicebrokerapi/servicebroker/blob/v2.13/spec.md#schema-object

duglin:
I don't see that error locally, could be a network issue. But I do get a 404 for: https://github.com/openservicebrokerapi/servicebroker/blob/master

duglin:
if we more network burps we can update the linker tool - the latest one has an option to retry.

mattmcneeney:
@duglin it picked up an actual bug - fixing now

angarg12:
LGTM

duglin:
LGTM grazie!

duglin:
2 more reviews needed

vaikas-google:
LGTM

pmorie:
LGTM

PR #327, opened 2017-09-27
Change object to JSON object in tables

mattmcneeney:
Consistency!


duglin:
image

LGTM

n3wscott:
The spec calls out when objects should be encoded as JSON. JSON is an encoding not an object type right? It seems like defining the type as just object is clear in the context. To my eye defining the type as 'JSON object' seems strange.

duglin:
my 2 cents... I don't care which we show in the tables, but let's be consistent. Right now we alternate based on a reason that I don't think most users would pick on.

mattmcneeney:
I agree with both of the above comments. JSON object seemed a little odd to
me, but I wanted consistency across the spec. I’d be happy to change all to
object if folks are happy with that?

duglin:
go for it

mattmcneeney:
Updated @duglin @n3wscott

n3wscott:
I think that reads better, and perhaps one last edit for this could be in the description field, sometimes it reads 'object' and sometimes 'JSON object'. I think it does make sense to have JSON in the description, example:

'An opaque object...' to 'An opaque JSON object...', etc

duglin:
LGTM
I'm ok w/ or w/o scott's latest suggestion.

vaikas-google:
LGTM

duglin:
LGTM

vaikas-google:
LGTM

angarg12:
LGTM

avade:
lgtm

PR #332, opened 2017-10-03
Add defines for missing terms.

n3wscott:
Please help me define these terms. I took a guess, but I could be wrong.

I also think `marketplace` should be defined but it is used in two different ways in the spec, once as `service marketplace` and another as `platform marketplace` and then alone as `marketplace`.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -62,6 +62,9 @@ be interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
 
 This specification defines the following terms:
 
+- *Application*: The software that will make use of, and be bound to, a Service
+  Instance.
+
 - *Platform*: The software that will manage the cloud environment into which
   Applications and Service Brokers are provisioned.  Users will not directly
   provision Services from Service Brokers, rather they will ask the Platform
@@ -73,19 +76,22 @@ This specification defines the following terms:
   some action. However, there can also be non-interactive Services that can
   perform the desired actions without direct prompting from the Application.
 
+- *Service Binding*: The representation of an association between an
+  Application and a Service Instance. Service Bindings will often
+  contain the credentials that the Application will use to communicate
+  with the Service Instance.
+
 - *Service Broker*: Service Brokers manage the lifecycle of Services. Platforms
   interact with Service Brokers to provision, and manage, Service Instances
   and Service Bindings.
 
 - *Service Instance*: An instantiation of a Service offering.
 
-- *Service Binding*: The representation of an association between an
-  Application and a Service Instance. Often, Service Bindings, will
-  contain the credentials that the Application will use to communicate
-  with the Service Instance.
+- *Service Offerings*: Generic bindable endpoints to extend or tie services
+  together. Examples include log drain and routing services.
 
-- *Application*: The software that will make use of, be bound to, a Service
-  Instance.
+- *Service Plan*: The representation of the costs and benefits for a given
+  service, potentially as a tier.
 
 ## Changes
 
--- a/spec.md
+++ b/spec.md
@@ -62,8 +62,13 @@ be interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
 
 This specification defines the following terms:
 
-- *Application*: The software that will make use of, and be bound to, a Service
-  Instance.
+- *Application*: Often the entity using a Service Instance will be a piece of
+  software, however, this does not need to be the case. For the purposes of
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -62,6 +62,9 @@ be interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
 
 This specification defines the following terms:
 
+- *Application*: The software that will make use of, and be bound to, a Service
```
duglin: I wonder if we can reword this to say that an app is just one of the things that a service can be bound to. For example, in kube its not an app as much as its the namespace.

spec.md
```diff
@@ -73,19 +76,22 @@ This specification defines the following terms:
   some action. However, there can also be non-interactive Services that can
   perform the desired actions without direct prompting from the Application.
 
+- *Service Binding*: The representation of an association between an
+  Application and a Service Instance. Service Bindings will often
+  contain the credentials that the Application will use to communicate
+  with the Service Instance.
```
duglin: I think we need to loosen this a bit since you can request a binding w/o knowing (or ever having) an app to bind to.  Perhaps something like:
```
*Service Binding*: Represents the request to use a Service Instance. As part of
this request there might be a reference to the entity that will use the 
Service Instance. Service Bindings will often contain the credentials that
can then be used to communicate with the Service Instance.
```

duglin: Yea, I know you just moved it but I think we need to tweak it since we're in this section anyway

spec.md
duglin: Perhaps:
```
*Application*: Often the entity using a Service Instance will be a piece of software,
however, this does not need to be the case. For the purposes of this
specification, the term "Application" will be used to represent all entities that
might make use of, and be bound to, a Service Instance.
```

duglin:
I agree,let's define marketplace too

duglin:
ping @n3wscott

n3wscott:
Thanks @duglin for the review, sorry for the delay. I took a stab at marketplace but feel free to suggest something else. I am still getting a handle on OSB.

spec.md
```diff
@@ -62,6 +62,14 @@ be interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
 
 This specification defines the following terms:
 
+- *Application*: Often the entity using a Service Instance will be a piece of
+  software, however, this does not need to be the case. For the purposes of
+  this specification, the term "Application" will be used to represent all
+  entities that might make use of, and be bound to, a Service Instance.
+
+- *Marketplace*: The collective offerings brought by a Service Broker for a
```
avade: How about:

Marketplace: An aggregation of services provided by Service Brokers that may have been curated before presented to end users of a platform.

n3wscott: I like the direction, I worked it in and updated.

spec.md
```diff
@@ -74,19 +81,22 @@ This specification defines the following terms:
...
   interact with Service Brokers to provision, and manage, Service Instances
   and Service Bindings.
 
 - *Service Instance*: An instantiation of a Service offering.
 
-- *Service Binding*: The representation of an association between an
-  Application and a Service Instance. Often, Service Bindings, will
-  contain the credentials that the Application will use to communicate
-  with the Service Instance.
+- *Service Offerings*: Generic bindable endpoints to extend or tie services
```
duglin: This one sounds a bit odd since it feels like its trying to define something other than just a normal service.
What about:
```
*Service Offering*: The advertisement of a service that a Service Broker supports.
```
and then , if needed, we could define "Service" too since don't actually do that today.

spec.md
```diff
@@ -74,19 +81,22 @@ This specification defines the following terms:
...
+  together. Examples include log drain and routing services.
 
-- *Application*: The software that will make use of, be bound to, a Service
-  Instance.
+- *Service Plan*: The representation of the costs and benefits for a given
+  service, potentially as a tier.
```
duglin: s/service/variant of the service/

spec.md
n3wscott: Service is defined.

duglin: whoa, no idea how i missed that earlier today

n3wscott:
Ok I have fixed the merge problems. Please remove the label.

duglin:
LGTM

mattmcneeney:
LGTM

vaikas-google:
LGTM

angarg12:
LGTM

avade:
LGTM

PR #333, opened 2017-10-10
Add GET endpoints for fetching a service instance/binding

mattmcneeney:
See issue #159

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -20,13 +20,13 @@
   - [Polling Last Operation](#polling-last-operation)
     - [Polling Interval and Duration](#polling-interval-and-duration)
   - [Provisioning](#provisioning)
+  - [Fetching a Service Instance](#fetching-a-service-instance)
   - [Updating a Service Instance](#updating-a-service-instance)
   - [Binding](#binding)
     - [Types of Binding](#types-of-binding)
+  - [Fetching a Service Binding](#fetching-a-service-binding)
   - [Unbinding](#unbinding)
   - [Deprovisioning](#deprovisioning)
-  - [Fetching a Service Instance](#fetching-a-service-instance)
-  - [Fetching a Service Binding](#fetching-a-service-binding)
   - [Orphans](#orphans)
 
 ## API Overview
@@ -886,6 +886,54 @@ For success responses, the following fields are defined:
 }
 ```
 
+## Fetching a Service Instance
+
+If `"instances_retrievable" :true` is declared for a service in the
+[Catalog](#catalog-management) endpoint, Service Brokers MUST support this
+endpoint for all plans of the service.
+
+### Request
+
+##### Route
+`GET /v2/service_instances/:instance_id`
+
+`:instance_id` is the identifier of a previously provisioned instance.
+
+##### cURL
+```
+$ curl 'http://username:password@broker-url/v2/service_instances/:instance_id' -X GET -H "X-Broker-API-Version: 2.13"
+```
+
+### Response
+
+| Status Code | Description |
+| --- | --- |
+| 200 OK | The expected response body is below. |
+| 404 Not Found | MUST be returned if the Service Instance does not exist or if a provisioning operation is still in progress. |
+
+Responses with any other status code will be interpreted as a failure and the
+Platform MUST continue to remember the Service Instance.
+
+##### Body
+
+For success responses, the following fields are defined:
+
+| Response field | Type | Description |
+| --- | --- | --- |
+| service_id | string | The ID of the service from the catalog that is associated with the Service Instance. |
+| plan_id | string | The ID of the plan from the catalog that is associated with the Service Instance. |
+| dashboard_url | string | The URL of a web-based management user interface for the Service Instance; we refer to this as a service dashboard. The URL MUST contain enough information for the dashboard to identify the resource being accessed (`9189kdfsk0vfnku` in the example below). Note: a Service Broker that wishes to return `dashboard_url` for a Service Instance MUST return it with the initial response to the provision request, even if the service is provisioned asynchronously. |
+| parameters | object | Configuration parameters for the Service Instance. |
+
+```
+{
+  "dashboard_url": "http://example-dashboard.example.com/9189kdfsk0vfnku",
+  "parameters": {
+    "billing-account": "abcde12345"
+  }
+}
+```
+
 ## Updating a Service Instance
 
 By implementing this endpoint, Service Broker authors can enable users to
@@ -1291,6 +1339,63 @@ can be mounted on all app instances simultaneously.
 }
 ```
 
+## Fetching a Service Binding
+
+If `"bindings_retrievable" :true` is declared for a service in the
+[Catalog](#catalog-management) endpoint, Service Brokers MUST support this
+endpoint for all plans of the service.
+
+### Request
+
+##### Route
+`GET /v2/service_instances/:instance_id/service_bindings/:binding_id`
+
+The `:instance_id` is the ID of a previously provisioned Service Instance. The
+`:binding_id` is the ID of a previously provisioned binding for that instance.
+
+##### cURL
+```
+$ curl 'http://username:password@broker-url/v2/service_instances/:instance_id/service_bindings/:binding_id' -X GET -H "X-Broker-API-Version: 2.13"
+```
+
+### Response
+
+| Status Code | Description |
+| --- | --- |
+| 200 OK | The expected response body is below. |
+| 404 Not Found | MUST be returned if the Service Binding does not exist or if a binding operation is still in progress. |
+
+Responses with any other status code will be interpreted as a failure and the
+Platform MUST continue to remember the Service Binding.
+
+##### Body
+
+For success responses, the following fields are defined:
+
+| Response Field | Type | Description |
+| --- | --- | --- |
+| credentials | object | A free-form hash of credentials that can be used by applications or users to access the service. |
+| syslog_drain_url | string | A URL to which logs MUST be streamed. `"requires":["syslog_drain"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform MUST consider the response invalid. |
+| route_service_url | string | A URL to which the Platform MUST proxy requests for the address sent with `bind_resource.route` in the request body. `"requires":["route_forwarding"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
+| volume_mounts | array-of-objects | An array of configuration for mounting volumes. `"requires":["volume_mount"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
+| parameters | object | Configuration parameters for the Service Binding. |
+
+```
+{
+  "credentials": {
+    "uri": "mysql://mysqluser:pass@mysqlhost:3306/dbname",
+    "username": "mysqluser",
+    "password": "pass",
+    "host": "mysqlhost",
+    "port": 3306,
+    "database": "dbname"
+  },
+  "parameters": {
+    "billing-account": "abcde12345"
+  }
+}
+```
+
 ## Unbinding
 
 Note: Service Brokers that do not provide any bindable services or plans do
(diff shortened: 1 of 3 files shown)


spec.md
```diff
@@ -1367,6 +1373,111 @@ Errors](#service-broker-errors).
...
+| route_service_url | string | A URL to which the platform MUST proxy requests for the address sent with `bind_resource.route` in the request body. `"requires":["route_forwarding"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the platform can consider the response invalid. |
+| volume_mounts | array-of-objects | An array of configuration for mounting volumes. `"requires":["volume_mount"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the platform can consider the response invalid. |
+| parameters | JSON object | Configuration options for the service instance. Service brokers SHOULD ensure that the configuration parameters match the service binding schema they provide to platforms via the [Catalog](#catalog-management) endpoint (this will enable platforms to offer additional features such as pre-populated form fields in UIs). If service brokers cannot fetch all configuration parameters, they MUST NOT return any. |
+
+```
+{
+  "credentials": {
+    "uri": "mysql://mysqluser:pass@mysqlhost:3306/dbname",
+    "username": "mysqluser",
+    "password": "pass",
```
n3wscott: Feels weird to be returning a password in clear text.

spec.md
```diff
@@ -1367,6 +1373,111 @@ Errors](#service-broker-errors).
 }
 ```
 
+## Fetching a Service Instance
+
+If `"instance_retrievable" :true` is declared for a service in the [Catalog](#catalog-management) endpoint, brokers MUST support this endpoint for all plans of the service.
```
kibbles-n-bytes: If `"instance_retrievable": false`, does that mean the endpoint is should not be supported for ANY plans? Or simply that not ALL plans support it, and it's on a case-by-case basis?

I think the former gives us more information, and since this field is not on a plan-by-plan basis we should make it a sweeping statement where either all plans support it or none do.

spec.md
```diff
@@ -1401,6 +1407,111 @@ Errors](#service-broker-errors).
...
+Responses with any other status code will be interpreted as a failure.
+Brokers can include a user-facing message in the `description` field; for
+details see [Broker Errors](#service-broker-errors).
+
+##### Body
+
+For success responses, a broker MUST return the following fields. For error
+responses, see [Broker Errors](#service-broker-errors).
+
+| Response field | Type | Description |
```
n3wscott: This will need `plan_id` returned as well.

spec.md
```diff
@@ -327,6 +329,8 @@ users when they have to type it as an argument on the command line.
 | tags | array-of-strings | Tags provide a flexible mechanism to expose a classification, attribute, or base technology of a service, enabling equivalent services to be swapped out without changes to dependent logic in applications, buildpacks, or other services. E.g. mysql, relational, redis, key-value, caching, messaging, amqp. |
 | requires | array-of-strings | A list of permissions that the user would have to give the service, if they provision it. The only permissions currently supported are `syslog_drain`, `route_forwarding` and `volume_mount`. |
 | bindable* | boolean | Specifies whether Service Instances of the service can be bound to applications. This specifies the default for all plans of this service. Plans can override this field (see [Plan Object](#plan-object)). |
+| instance_retrievable | boolean | Specifies whether the [Fetching a Service Instance](#fetching-a-service-instance) endpoint is supported for all plans. |
+| binding_retrievable | boolean | Specifies whether the [Fetching a Service Binding](#fetching-a-service-binding) endpoint is supported for all plans. |
```
mattmcneeney: Should be plural?

spec.md
```diff
@@ -1420,6 +1426,113 @@ Errors](#service-broker-errors).
...
+| Response field | Type | Description |
+| --- | --- | --- |
+| service_id | string | The ID of the service from the catalog that is associated with the Service Instance. |
+| plan_id | string | The ID of the plan from the catalog that is associated with the Service Instance. |
+| dashboard_url | string | The URL of a web-based management user interface for the Service Instance; we refer to this as a service dashboard. The URL MUST contain enough information for the dashboard to identify the resource being accessed (`9189kdfsk0vfnku` in the example below). Note: a broker that wishes to return `dashboard_url` for a Service Instance MUST return it with the initial response to the provision request, even if the service is provisioned asynchronously. |
+| parameters | JSON object | Configuration options for the Service Instance. Service Brokers SHOULD ensure that the configuration parameters match the Service Instance update schema they provide to Platforms via the [Catalog](#catalog-management) endpoint (this will enable Platforms to offer additional features such as pre-populated form fields in UIs). If brokers cannot fetch all configuration parameters, they MUST NOT return any. |
```
mattmcneeney: This should say parameters, not options

mattmcneeney: And "Service Brokers" not "brokers"

mattmcneeney: Should this be **all** configuration parameters rather than only those used for update? (Schemas for resources would help UIs display this info)

spec.md
```diff
@@ -1420,6 +1426,113 @@ Errors](#service-broker-errors).
...
+| syslog_drain_url | string | A URL to which logs MUST be streamed. `"requires":["syslog_drain"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform MUST consider the response invalid. |
+| route_service_url | string | A URL to which the Platform MUST proxy requests for the address sent with `bind_resource.route` in the request body. `"requires":["route_forwarding"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
+| volume_mounts | array-of-objects | An array of configuration for mounting volumes. `"requires":["volume_mount"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
+| parameters | JSON object | Configuration options for the Service Instance. Service Brokers SHOULD ensure that the configuration parameters match the Service Binding schema they provide to Platforms via the [Catalog](#catalog-management) endpoint (this will enable Platforms to offer additional features such as pre-populated form fields in UIs). If Service Brokers cannot fetch all configuration parameters, they MUST NOT return any. |
```
mattmcneeney: Same comments as above (also, there is no such thing as a "Service Binding schema" - it is the create service binding schema)

spec.md
```diff
@@ -777,7 +783,7 @@ The following HTTP Headers are defined for this operation:
   "service_id": "service-id-here",
   "plan_id": "plan-id-here",
   "context": {
-    "platform": "cloudfoundry",
+    "Platform": "cloudfoundry",
```
mattmcneeney: Fix this

spec.md
```diff
@@ -932,7 +938,7 @@ the user did not explicitly specify in their request for the update.
 ```
 {
   "context": {
-    "platform": "cloudfoundry",
+    "Platform": "cloudfoundry",
```
mattmcneeney: Fix this

mattmcneeney: Fix all

spec.md
```diff
@@ -219,9 +221,9 @@ X-Broker-API-Originating-Identity: Platform value
 
 `Platform` MUST be a non-empty string indicating the Platform from which
 the request is being sent. The specific value SHOULD match the values
-defined in the [profile](profile.md) document for the `context.platform`
+defined in the [profile](profile.md) document for the `context.Platform`
```
mattmcneeney: Fix these too

spec.md
```diff
@@ -1420,6 +1426,113 @@ Errors](#service-broker-errors).
...
+
+For success responses, a Service Broker MUST return the following fields. For error
+responses, see [Broker Errors](#service-broker-errors).
+
+| Response field | Type | Description |
+| --- | --- | --- |
+| service_id | string | The ID of the service from the catalog that is associated with the Service Instance. |
+| plan_id | string | The ID of the plan from the catalog that is associated with the Service Instance. |
+| dashboard_url | string | The URL of a web-based management user interface for the Service Instance; we refer to this as a service dashboard. The URL MUST contain enough information for the dashboard to identify the resource being accessed (`9189kdfsk0vfnku` in the example below). Note: a Service Broker that wishes to return `dashboard_url` for a Service Instance MUST return it with the initial response to the provision request, even if the service is provisioned asynchronously. |
+| parameters | JSON object | Configuration parameters for the Service Instance. If Service Brokers cannot fetch all configuration parameters, they MUST NOT return any. |
```
mattmcneeney:
```
If Service Brokers cannot fetch all configuration parameters, they MUST NOT return any.
```

I can't remember the logic behind adding this restriction but it seems overly restrictive now. Any thoughts anyone?

duglin: I think it was because there's no way to know what "absence" means... not there are all? Didn't change? something else?

mattmcneeney: So if we assume a broker returns an error if it "can't fetch" parameters for some reason, then what about if a broker only saves some parameters for whatever reason?

Whilst that wouldn't lead to the nicest UI, being able to show some information about how it was configured seems better than none.

spec.md
```diff
@@ -1420,6 +1426,113 @@ Errors](#service-broker-errors).
...
+```
+
+### Response
+
+| Status Code | Description |
+| --- | --- |
+| 200 OK | The expected response body is below. |
+| 404 Not Found | MUST be returned if the Service Binding does not exist or if a binding operation is still in progress. The expected response body is `{}`. |
+
+Responses with any other status code will be interpreted as a failure database.
```
mhrivnak: I suspect the word "database" is not meant to be there.

spec.md
mhrivnak: Consider simply ``object``, which is what the rest of the spec uses, instead of ``JSON object``.

spec.md
```diff
@@ -1420,6 +1426,113 @@ Errors](#service-broker-errors).
...
+ignored. For error responses, see [Broker Errors](#service-broker-errors).
+
+| Response Field | Type | Description |
+| --- | --- | --- |
+| credentials | object | A free-form hash of credentials that can be used by applications or users to access the service. |
+| syslog_drain_url | string | A URL to which logs MUST be streamed. `"requires":["syslog_drain"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform MUST consider the response invalid. |
+| route_service_url | string | A URL to which the Platform MUST proxy requests for the address sent with `bind_resource.route` in the request body. `"requires":["route_forwarding"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
+| volume_mounts | array-of-objects | An array of configuration for mounting volumes. `"requires":["volume_mount"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
+| parameters | JSON object | Configuration parameters for the Service Binding. If Service Brokers cannot fetch all configuration parameters, they MUST NOT return any. |
```
mhrivnak: Consider simply ``object``, which is what the rest of the spec uses, instead of ``JSON object``.

spec.md
mattmcneeney: Good spot, thanks!

spec.md
mattmcneeney: Good spot, thanks!

spec.md
mattmcneeney: Removed, thanks!

mattmcneeney:
I've removed the `If Service Brokers cannot fetch all configuration parameters, they MUST NOT return any.` line from this PR. It seems overly restrictive for little gain, and raises questions like "what about parameters that can be set at create not update time?" etc. Schemas will help here.

Obviously this PR is on hold, but since I'm changing this and k8s has already implemented it, I need to hear from @kibbles-n-bytes @pmorie @vaikas-google etc on whether or not you support this or need this back in!

spec.md
```diff
@@ -1379,6 +1385,109 @@ For success responses, the following fields are defined:
...
+| Status Code | Description |
+| --- | --- |
+| 200 OK | The expected response body is below. |
+| 404 Not Found | MUST be returned if the Service Binding does not exist or if a binding operation is still in progress. |
+
+Responses with any other status code will be interpreted as a failure.
+
+##### Body
+
+For success responses, the following fields are defined:
```
jenspinney: Does "the following fields are defined" imply that all of the fields are optional or mandatory? I think it'd be clearer if we were more explicit. (Same with the instance parameter fields above). In the descriptions below, it seems to say "if this feature is not defined in the catalog, then including this field is invalid", but if the feature is defined in the catalog, is it a requirement that this field be present? Is it allowed to return multiple of these fields?

duglin: The optionality of a field is defied by whether it is marked with an asterisk or not; * == required.  By using the word "defined" we're saying that the spec has some well-known fields w/o implying that extension fields can't also be defined.  Which specific text are you referring to in the 2nd part of your question?

mattmcneeney:
As promised, this PR has been rebased and is ready to go! Let's review this!

mattmcneeney:
@n3wscott I added the updates to the swagger and openapi docs. Could you take a look at them and make sure they look ok to you? Same request @duglin Thank you!

spec.md
n3wscott: I'm ok

jmrodri:
LGTM

vaikas-google:
LGTM

nilebox:
@mattmcneeney can you remove the "DO NOT MERGE." text from the PR description, if it's good to go now?

mattmcneeney:
Good point thanks @nilebox - removed

spec.md
```diff
@@ -25,6 +25,8 @@
     - [Types of Binding](#types-of-binding)
   - [Unbinding](#unbinding)
   - [Deprovisioning](#deprovisioning)
+  - [Fetching a Service Instance](#fetching-a-service-instance)
```
duglin: Minor thing but the order of the TOC seems to be: provision, update instance, bind, unbind, deprovision - which is the normal lifecycle of things. I would think "fetching" should be in the middle of that flow, not at the end.

spec.md
```diff
@@ -1428,6 +1434,109 @@ For success responses, the following fields are defined:
...
+```
+$ curl 'http://username:password@broker-url/v2/service_instances/:instance_id' -X GET -H "X-Broker-API-Version: 2.13"
+```
+
+### Response
+
+| Status Code | Description |
+| --- | --- |
+| 200 OK | The expected response body is below. |
+| 404 Not Found | MUST be returned if the Service Instance does not exist or if a provisioning operation is still in progress. |
```
duglin: In other ops we return a 410 not 404 when the Instance doesn't exist. Should we be consistent?

mattmcneeney:
All issues addressed except the 404 vs 410... needs discussion

spec.md
duglin: during 4/11 F2F we agreed to keeping it a 404

duglin:
LGTM

fmui:
LGTM

spec.md
```diff
@@ -386,6 +388,8 @@ It is therefore RECOMMENDED that implementations avoid such strings.
 | tags | array of strings | Tags provide a flexible mechanism to expose a classification, attribute, or base technology of a service, enabling equivalent services to be swapped out without changes to dependent logic in applications, buildpacks, or other services. E.g. mysql, relational, redis, key-value, caching, messaging, amqp. |
 | requires | array of strings | A list of permissions that the user would have to give the service, if they provision it. The only permissions currently supported are `syslog_drain`, `route_forwarding` and `volume_mount`. |
 | bindable* | boolean | Specifies whether Service Instances of the service can be bound to applications. This specifies the default for all plans of this service. Plans can override this field (see [Plan Object](#plan-object)). |
+| instances_retrievable | boolean | Specifies whether the [Fetching a Service Instance](#fetching-a-service-instance) endpoint is supported for all plans. |
+| bindings_retrievable | boolean | Specifies whether the [Fetching a Service Binding](#fetching-a-service-binding) endpoint is supported for all plans. |
```
dotchev: For backward compatibility the default should be `false`. State it explicitly.
For both  `instances_retrievable` and `bindings_retrievable`.

mattmcneeney: @dotchev We cannot state that new fields are required in the catalog. Platforms should treat the absence of the field the same as it would treat the presence of the field set to false.

PR #334, opened 2017-10-10
Support asynchronous bindings (requires #333)

mattmcneeney:
See issue #137

This depends on #333

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1369,9 +1369,9 @@ For success responses, the following fields are defined:
 | Response Field | Type | Description |
 | --- | --- | --- |
 | credentials | object | A free-form hash of credentials that can be used by applications or users to access the service. MUST be returned for a synchronous binding if the Service Broker supports generation of credentials. |
-| syslog_drain_url | string | A URL to which logs MUST be streamed. If the service declares `"requires":["syslog_drain"]` in the [Catalog](#catalog-management) endpoint, this MUST be returned for a synchronous binding or the Platform can consider the response invalid. |
-| route_service_url | string | A URL to which the Platform MUST proxy requests for the address sent with `bind_resource.route` in the request body. If the service declares `"requires":["route_forwarding"]` in the [Catalog](#catalog-management) endpoint and `"bind_resource":{"route":"some-address.com"}` is included in the request, this MUST be returned for a synchronous binding or the Platform can consider the response invalid. |
-| volume_mounts | array of [VolumeMount](#volume-mount-object) objects | An array of configuration for remote storage devices to be mounted into an application container filesystem. If the service declares `"requires":["volume_mount"]` in the [Catalog](#catalog-management) endpoint, this MUST be returned for a synchronous binding or the Platform can consider the response invalid. |
+| syslog_drain_url | string | A URL to which logs MUST be streamed. For synchronous bindings, `"requires":["syslog_drain"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
+| route_service_url | string | A URL to which the Platform MUST proxy requests for the address sent with `bind_resource.route` in the request body. For synchronous bindings, `"requires":["route_forwarding"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
+| volume_mounts | array of [VolumeMount](#volume-mount-object) objects | An array of configuration for remote storage devices to be mounted into an application container filesystem. For synchronous bindings, `"requires":["volume_mount"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
 | operation | string | For asynchronous responses, Service Brokers MAY return an identifier representing the operation. The value of this field MUST be provided by the Platform with requests to the [Polling Last Operation for Service Bindings](#polling-last-operation-for-service-bindings) endpoint in a URL encoded query parameter. If present, MUST be a non-empty string. |
 
 ##### Volume Mount Object
@@ -1469,9 +1469,9 @@ For success responses, the following fields are defined:
 | Response Field | Type | Description |
 | --- | --- | --- |
 | credentials | object | A free-form hash of credentials that can be used by applications or users to access the service. MUST be returned if the Service Broker supports generation of credentials and the Service Binding was provisioned asynchronously. |
-| syslog_drain_url | string | A URL to which logs MUST be streamed. MUST be returned if the service declares `"requires":["syslog_drain"]` in the [Catalog](#catalog-management) endpoint and the Service Binding was provisioned asynchronously. |
-| route_service_url | string | A URL to which the Platform MUST proxy requests for the address sent with `bind_resource.route` in the request body. MUST be returned if the service declares `"requires":["route_forwarding"]` in the [Catalog](#catalog-management) endpoint and the Service Binding was provisioned asynchronously and the original request contained `"bind_resource":{"route":"some-address.com"}`. |
-| volume_mounts | array of [VolumeMount](#volume-mount-object) objects | An array of configuration for mounting volumes. MUST be returned if the service declares `"requires":["volume_mount"]` in the [Catalog](#catalog-management) endpoint and the Service Binding was provisioned asynchronously. |
+| syslog_drain_url | string | A URL to which logs MUST be streamed. For asynchronous bindings, `"requires":["syslog_drain"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
+| route_service_url | string | A URL to which the Platform MUST proxy requests for the address sent with `bind_resource.route` in the request body. For asynchronous bindings, `"requires":["route_forwarding"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
+| volume_mounts | array of [VolumeMount](#volume-mount-object) objects | An array of configuration for mounting volumes. For asynchronous bindings, `"requires":["volume_mount"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
 | parameters | object | Configuration parameters for the Service Binding. |
 
 ```
--- a/spec.md
+++ b/spec.md
@@ -696,7 +696,7 @@ to cease polling.
 #### Route
 `GET /v2/service_instances/:instance_id/last_operation`
 
-`:instance_id` MUST be a globally unique non-empty string.
+`:instance_id` MUST be the ID of a previously provisioned Service Instance.
 
 #### Parameters
 
@@ -779,8 +779,10 @@ regarding the Service Binding can then immediately be fetched using the
 #### Route
 `GET /v2/service_instances/:instance_id/service_bindings/:binding_id/last_operation`
 
-`:instance_id` MUST be a globally unique non-empty string.
-`:binding_id` MUST be a globally unique non-empty string.
+`:instance_id` MUST be the ID of a previously provisioned Service Instance.
+
+`:binding_id` MUST be the ID of a previously provisioned binding for that
+instance.
 
 #### Parameters
 
@@ -973,7 +975,8 @@ any circumstances.
 ##### Route
 `GET /v2/service_instances/:instance_id`
 
-`:instance_id` is the identifier of a previously provisioned instance.
+`:instance_id` MUST be the ID of a previously provisioned Service Instance.
+instance.
 
 ##### cURL
 ```
@@ -1441,8 +1444,10 @@ circumstances.
 ##### Route
 `GET /v2/service_instances/:instance_id/service_bindings/:binding_id`
 
-The `:instance_id` is the ID of a previously provisioned Service Instance. The
-`:binding_id` is the ID of a previously provisioned binding for that instance.
+`:instance_id` MUST be the ID of a previously provisioned Service Instance.
+
+`:binding_id` MUST be the ID of a previously provisioned binding for that
+instance.
 
 ##### cURL
 ```
--- a/spec.md
+++ b/spec.md
@@ -809,7 +809,7 @@ $ curl http://username:password@broker-url/v2/service_instances/:instance_id/ser
 | --- | --- |
 | 200 OK | MUST be returned upon successful processing of this request. The expected response body is below. |
 | 400 Bad Request | MUST be returned if the request is malformed or missing mandatory data. |
-| 410 Gone | Appropriate only for asynchronous delete operations. The Platform MUST consider this response a success and remove the resource from its database. The expected response body is `{}`. Returning this while the Platform is polling for create operations SHOULD be interpreted as an invalid response and the Platform SHOULD continue polling. |
+| 410 Gone | Appropriate only for asynchronous delete operations. The Platform MUST consider this response a success and remove the resource from its database. Returning this while the Platform is polling for create operations SHOULD be interpreted as an invalid response and the Platform SHOULD continue polling. |
 
 Responses with any other status code SHOULD be interpreted as an error or
 invalid response. The Platform SHOULD continue polling until the broker returns
@@ -1351,7 +1351,7 @@ $ curl http://username:password@service-broker-url/v2/service_instances/:instanc
 | --- | --- |
 | 200 OK | MUST be returned if the binding already exists and the requested parameters are identical to the existing binding. The expected response body is below. |
 | 201 Created | MUST be returned if the binding was created as a result of this request. The expected response body is below. |
-| 202 Accepted | MUST be returned if the binding is in progress. The `operation` string MUST match that returned for the original request. This triggers the Platform to poll the [Polling Last Operation for Service Bindings](#polling-last-operation-for-service-bindings) endpoint for operation status. Information regarding the Service Binding (i.e. credentials) MUST NOT be returned in this asynchronous request. Note that a re-sent `PUT` request MUST return a `202 Accepted`, not a `200 OK`, if the binding is not yet fully created. |
+| 202 Accepted | MUST be returned if the binding is in progress. The `operation` string MUST match that returned for the original request. This triggers the Platform to poll the [Polling Last Operation for Service Bindings](#polling-last-operation-for-service-bindings) endpoint for operation status. Information regarding the Service Binding (i.e. credentials) MUST NOT be returned in this response. Note that a re-sent `PUT` request MUST return a `202 Accepted`, not a `200 OK`, if the binding is not yet fully created. |
 | 400 Bad Request | MUST be returned if the request is malformed or missing mandatory data. |
 | 409 Conflict | MUST be returned if a Service Binding with the same id, for the same Service Instance, already exists but with different parameters. |
 | 422 Unprocessable Entity | MUST be returned if the Service Broker requires that `app_guid` be included in the request body. The response body MUST contain error code `"RequiresApp"` (see [Service Broker Errors](#service-broker-errors)). The error response MAY include a helpful error message in the `description` field such as `"This Service supports generation of credentials through binding an application only."`. Additionally, if the Service Broker rejects the request due to a concurrent request to create a binding for the same Service Instance, then this error MUST be returned (see [Blocking Operations](#blocking-operations)). This MUST also be returned if the Service Broker only supports asynchronous bindings for the Service Instance and the request did not include `?accepts_incomplete=true`. In this case, the response body MUST contain error code `"AsyncRequired"` (see [Service Broker Errors](#service-broker-errors)). The error response MAY include a helpful error message in the `description` field such as `"This Service Instance requires client support for asynchronous binding operations."`. |
@@ -1368,10 +1368,10 @@ For success responses, the following fields are defined:
 
 | Response Field | Type | Description |
 | --- | --- | --- |
-| credentials | object | A free-form hash of credentials that can be used by applications or users to access the service. MUST be returned for a synchronous request if the Service Broker supports generation of credentials. |
-| syslog_drain_url | string | A URL to which logs MUST be streamed. If the service declares `"requires":["syslog_drain"]` in the [Catalog](#catalog-management) endpoint, this MUST be returned for a synchronous request or the Platform can consider the response invalid. |
-| route_service_url | string | A URL to which the Platform MUST proxy requests for the address sent with `bind_resource.route` in the request body. If the service declares `"requires":["route_forwarding"]` in the [Catalog](#catalog-management) endpoint and `"bind_resource":{"route":"some-address.com"}` is included in the request, this MUST be returned for a synchronous request or the Platform can consider the response invalid. |
-| volume_mounts | array of [VolumeMount](#volume-mount-object) objects | An array of configuration for remote storage devices to be mounted into an application container filesystem. If the service declares `"requires":["volume_mount"]` in the [Catalog](#catalog-management) endpoint, this MUST be returned for a synchronous request or the Platform can consider the response invalid. |
+| credentials | object | A free-form hash of credentials that can be used by applications or users to access the service. MUST be returned for a synchronous binding if the Service Broker supports generation of credentials. |
+| syslog_drain_url | string | A URL to which logs MUST be streamed. If the service declares `"requires":["syslog_drain"]` in the [Catalog](#catalog-management) endpoint, this MUST be returned for a synchronous binding or the Platform can consider the response invalid. |
+| route_service_url | string | A URL to which the Platform MUST proxy requests for the address sent with `bind_resource.route` in the request body. If the service declares `"requires":["route_forwarding"]` in the [Catalog](#catalog-management) endpoint and `"bind_resource":{"route":"some-address.com"}` is included in the request, this MUST be returned for a synchronous binding or the Platform can consider the response invalid. |
+| volume_mounts | array of [VolumeMount](#volume-mount-object) objects | An array of configuration for remote storage devices to be mounted into an application container filesystem. If the service declares `"requires":["volume_mount"]` in the [Catalog](#catalog-management) endpoint, this MUST be returned for a synchronous binding or the Platform can consider the response invalid. |
 | operation | string | For asynchronous responses, Service Brokers MAY return an identifier representing the operation. The value of this field MUST be provided by the Platform with requests to the [Polling Last Operation for Service Bindings](#polling-last-operation-for-service-bindings) endpoint in a URL encoded query parameter. If present, MUST be a non-empty string. |
 
 ##### Volume Mount Object
@@ -1436,8 +1436,6 @@ endpoint for all services and plans that support bindings (`"bindable": true`)
 and this endpoint MUST be available immediately after the
 [Polling Last Operation for Service Bindings](#polling-last-operation-for-service-bindings)
 endpoint returns `"state": "succeeded"` for a [Binding](#binding) operation.
-Otherwise, Platforms SHOULD NOT attempt to call this endpoint under any
-circumstances.
 
 ### Request
 
--- a/spec.md
+++ b/spec.md
@@ -1362,6 +1362,14 @@ created on the Service Broker. However, while the platform will attempt
 to send an unbind request, Service Brokers MAY automatically delete
 any resources associated with the failed bind request on their own.
 
+#### Body (for a `202 Accepted` response code)
+
+For a `202 Accepted` response code, the following fields are defined:
+
+| Response Field | Type | Description |
+| --- | --- | --- |
+| operation | string | For asynchronous responses, Service Brokers MAY return an identifier representing the operation. The value of this field MUST be provided by the Platform with requests to the [Polling Last Operation for Service Bindings](#polling-last-operation-for-service-bindings) endpoint in a URL encoded query parameter. If present, MUST be a non-empty string. |
+
 #### Body (for `200 OK` and `201 Created` response codes)
 
 For `200 OK` and `201 Created` response codes, the following fields are defined:
@@ -1372,14 +1380,6 @@ For `200 OK` and `201 Created` response codes, the following fields are defined:
 | syslog_drain_url | string | A URL to which logs MUST be streamed. `"requires":["syslog_drain"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
 | route_service_url | string | A URL to which the Platform MUST proxy requests for the address sent with `bind_resource.route` in the request body. `"requires":["route_forwarding"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
 | volume_mounts | array of [VolumeMount](#volume-mount-object) objects | An array of configuration for remote storage devices to be mounted into an application container filesystem. `"requires":["volume_mount"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
-
-#### Body (for a `202 Accepted` response code)
-
-For a `202 Accepted` response code, the following fields are defined:
-
-| Response Field | Type | Description |
-| --- | --- | --- |
-| operation | string | For asynchronous responses, Service Brokers MAY return an identifier representing the operation. The value of this field MUST be provided by the Platform with requests to the [Polling Last Operation for Service Bindings](#polling-last-operation-for-service-bindings) endpoint in a URL encoded query parameter. If present, MUST be a non-empty string. |
 
 ##### Volume Mount Object
 
--- a/spec.md
+++ b/spec.md
@@ -1191,9 +1191,7 @@ services do not need to implement the endpoint for bind requests.
 #### Credentials
 
 Credentials are a set of information used by an Application or a user to
-utilize the Service Instance. If the Service Broker supports generation of
-credentials it MUST return `credentials` in the response for a request to
-create a Service Binding. Credentials SHOULD be unique whenever possible, so
+utilize the Service Instance. Credentials SHOULD be unique whenever possible, so
 access can be revoked for each binding without affecting consumers of other
 bindings for the Service Instance.
 
@@ -1205,10 +1203,6 @@ logs needs information for the location to which it will stream logs. A create
 binding response from a Service Broker that provides one of these services MUST
 include a `syslog_drain_url`. The Platform MUST use the `syslog_drain_url` value
 when sending logs to the service.
-
-Service Brokers MUST NOT include a `syslog_drain_url` in a create binding
-response if the associated [Catalog](#catalog-management) entry for the
-service did not include a `"requires":["syslog_drain"]` property.
 
 #### Route Services
 Route services are a class of Service Offerings that intermediate requests to
@@ -1231,19 +1225,11 @@ proxy application requests through instances of the service is managed
 out-of-band, the Service Broker MUST NOT return `route_service_url` in the
 response.
 
-Service Brokers MUST NOT include a `route_service_url` in a create binding
-response if the associated [Catalog](#catalog-management) entry for the
-service did not include a `"requires":["route_forwarding"]` property.
-
 #### Volume Services
 
 There are a class of services that provide network storage to applications
 via volume mounts in the application container. A create binding response from
 one of these services MUST include `volume_mounts`.
-
-Service Brokers MUST NOT include `volume_mounts` in a create binding response
-if the associated [Catalog](#catalog-management) entry for the service
-did not include a `"requires":["volume_mount"]` property.
 
 ### Request
 
@@ -1376,7 +1362,7 @@ For `200 OK` and `201 Created` response codes, the following fields are defined:
 
 | Response Field | Type | Description |
 | --- | --- | --- |
-| credentials | object | A free-form hash of credentials that can be used by applications or users to access the service. |
+| credentials | object | A free-form hash of credentials that can be used by applications or users to access the service. MUST be returned if the Service Broker supports generation of credentials. |
 | syslog_drain_url | string | A URL to which logs MUST be streamed. `"requires":["syslog_drain"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
 | route_service_url | string | A URL to which the Platform MUST proxy requests for the address sent with `bind_resource.route` in the request body. `"requires":["route_forwarding"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
 | volume_mounts | array of [VolumeMount](#volume-mount-object) objects | An array of configuration for remote storage devices to be mounted into an application container filesystem. `"requires":["volume_mount"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
--- a/spec.md
+++ b/spec.md
@@ -1362,16 +1362,23 @@ created on the Service Broker. However, while the platform will attempt
 to send an unbind request, Service Brokers MAY automatically delete
 any resources associated with the failed bind request on their own.
 
-#### Body
+#### Body (for `200 OK` and `201 Created` response codes)
 
-For success responses, the following fields are defined:
+For `200 OK` and `201 Created` response codes, the following fields are defined:
 
 | Response Field | Type | Description |
 | --- | --- | --- |
-| credentials | object | A free-form hash of credentials that can be used by applications or users to access the service. MUST be returned for a synchronous binding if the Service Broker supports generation of credentials. MUST NOT be returned if the Service Broker is creating the Service Binding asynchronously. |
-| syslog_drain_url | string | A URL to which logs MUST be streamed. `"requires":["syslog_drain"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. MUST NOT be returned if the Service Broker is creating the Service Binding asynchronously. |
-| route_service_url | string | A URL to which the Platform MUST proxy requests for the address sent with `bind_resource.route` in the request body. `"requires":["route_forwarding"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. MUST NOT be returned if the Service Broker is creating the Service Binding asynchronously. |
-| volume_mounts | array of [VolumeMount](#volume-mount-object) objects | An array of configuration for remote storage devices to be mounted into an application container filesystem. `"requires":["volume_mount"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. MUST NOT be returned if the Service Broker is creating the Service Binding asynchronously. |
+| credentials | object | A free-form hash of credentials that can be used by applications or users to access the service. |
+| syslog_drain_url | string | A URL to which logs MUST be streamed. `"requires":["syslog_drain"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
+| route_service_url | string | A URL to which the Platform MUST proxy requests for the address sent with `bind_resource.route` in the request body. `"requires":["route_forwarding"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
+| volume_mounts | array of [VolumeMount](#volume-mount-object) objects | An array of configuration for remote storage devices to be mounted into an application container filesystem. `"requires":["volume_mount"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
+
+#### Body (for a `202 Accepted` response code)
+
+For a `202 Accepted` response code, the following fields are defined:
+
+| Response Field | Type | Description |
+| --- | --- | --- |
 | operation | string | For asynchronous responses, Service Brokers MAY return an identifier representing the operation. The value of this field MUST be provided by the Platform with requests to the [Polling Last Operation for Service Bindings](#polling-last-operation-for-service-bindings) endpoint in a URL encoded query parameter. If present, MUST be a non-empty string. |
 
 ##### Volume Mount Object
--- a/spec.md
+++ b/spec.md
@@ -17,8 +17,9 @@
     - [Synchronous Operations](#synchronous-operations)
     - [Asynchronous Operations](#asynchronous-operations)
   - [Blocking Operations](#blocking-operations)
-  - [Polling Last Operation](#polling-last-operation)
-    - [Polling Interval and Duration](#polling-interval-and-duration)
+  - [Polling Last Operation for Service Instances](#polling-last-operation-for-service-instances)
+  - [Polling Last Operation for Service Bindings](#polling-last-operation-for-service-bindings)
+  - [Polling Interval and Duration](#polling-interval-and-duration)
   - [Provisioning](#provisioning)
   - [Fetching a Service Instance](#fetching-a-service-instance)
   - [Updating a Service Instance](#updating-a-service-instance)
@@ -650,8 +651,7 @@ operations."`.
 
 If the query parameter described above is present, and the Service Broker
 executes the request asynchronously, the Service Broker MUST return the
-asynchronous response `202 Accepted`. The response body SHOULD be the same as
-if the Service Broker were serving the request synchronously.
+asynchronous response `202 Accepted`.
 
 An asynchronous response triggers the Platform to poll the endpoint
 `GET /v2/service_instances/:instance_id/last_operation` until the Service Broker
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -1367,6 +1468,122 @@ Errors](#service-broker-errors).
...
+| 404 Not Found | MUST be returned if the service binding does not exist or if a binding operation is still in progress. The expected response body is `{}`. |
+
+Responses with any other status code will be interpreted as a failure database.
+Brokers can include a user-facing message in the `description` field; for
+details see [Broker Errors](#service-broker-errors).
+
+##### Body
+
+For success responses, the following fields are supported. Others will be
+ignored. For error responses, see [Broker Errors](#service-broker-errors).
```
vaikas-google: Do we need this 'Others will be ignored'? I assume the intent here is that some brokers might return some other fields due to the platforms/systems they are built on and we want to allow for that? Question is, will there be some brokers that return additional fields that might not be in the spec yet but the platforms will use. I'm just thinking, who ignores these fields and is the intent that they _must_ be ignored, or something. Just curious :)

spec.md
```diff
@@ -1126,6 +1209,7 @@ $ curl http://username:password@service-broker-url/v2/service_instances/:instanc
 | --- | --- |
 | 200 OK | MUST be returned if the binding already exists and the requested parameters are identical to the existing binding. The expected response body is below. |
 | 201 Created | MUST be returned if the binding was created as a result of this request. The expected response body is below. |
+| 202 Accepted | MUST be returned if the binding is in progress. This triggers the platform marketplace to poll the [Polling Last Operation for Service Bindings](#polling-last-operation-for-service-bindings) endpoint for operation status. Information regarding the service binding (i.e. credentials) MUST NOT be returned in this asynchronous request. Note that a re-sent `PUT` request MUST return a `202 Accepted`, not a `200 OK`, if the binding is not yet fully created. |
```
vaikas-google: re-sent `PUT`, does it have to be identical PUT?

spec.md
mattmcneeney: I just brought this across for consistency. 'Others will be ignored' is always kinda weird for a JSON object. If we remove this across the spec I'll happily remove here!

spec.md
mattmcneeney: Yes!

duglin:
Still discussing this internally but some immediate feedback I received is that requiring GET might be a bit of a security issue or hardship. This will force the broker to not just persist the credentials during the bind() and last_operation() phase, but then to persist it forever.

nilebox:
> This will force the broker to not just persist the credentials during the bind() and last_operation() phase, but then to persist it forever.

Isn't that already true?

According to the spec, if the broker receives a duplicate `bind` request (with the same parameters), it should return `200 OK` with existing credentials:
> 

200 OK | MUST be returned if the binding already exists and the requested parameters are identical to the existing binding. The expected response body is below.
-- | --

This duplicate binding request can occur at any time, not necessarily within a short amount of time after the initial binding request.

duglin:
hmmm that's a good point. I'm curious, for the CF guys who support stateless brokers, how can this check be performed if the broker is stateless?

spec.md
```diff
@@ -1367,6 +1468,122 @@ Errors](#service-broker-errors).
...
+```
+$ curl 'http://username:password@broker-url/v2/service_instances/:instance_id' -X GET -H "X-Broker-API-Version: 2.11"
+```
+
+### Response
+
+| Status Code | Description |
+| --- | --- |
+| 200 OK | The expected response body is below. |
+| 404 Not Found | MUST be returned if the service instance does not exist or if a provisioning operation is still in progress. The expected response body is `{}`. |
```
nilebox: What should happen if the instance **update** request is in progress (i.e. there is an existing resource, but the update processing is not finished yet)?

spec.md
```diff
@@ -673,10 +679,82 @@ Errors](#service-broker-errors).
...
+endpoint to obtain the state of the last requested operation. The broker
+response MUST contain the field `state` and MAY contain the field `description`.
+
+Valid values for `state` are `in progress`, `succeeded`, and `failed`. The
+platform will poll the `last_operation` endpoint as long as the broker returns
+`"state": "in progress"`. Returning `"state": "succeeded"` or
+`"state": "failed"` will cause the platform to cease polling and, in the case of
+a [Binding](#binding) request, information on the service binding can then
+immediately be fetched using the
+[Fetching a Service Binding](#fetching-a-service-binding) endpoint. The value
```
nilebox: But what if the catalog endpoint returns `"binding_retrievable": false`? Do we still support async bindings (without being able to fetch any result)?

kibbles-n-bytes: I think that we should explicitly add that if a binding supports async operations, it must have `"binding_retrievable": true`.

spec.md
kibbles-n-bytes: Seems to me like this needs to be outlined here as a `422 - Unprocessable Entity` situation.

spec.md
kibbles-n-bytes: The wording with `binding_retrievable` is a bit ambiguous. It says:

```If `"binding_retrievable" :true` is declared for a service in the Catalog endpoint, brokers MUST support this endpoint for all plans of the service.```

It doesn't say that having `"binding_retrievable": false` means the endpoint is unavailable for all plans, so you could potentially have some plans able to support it and some not, and thus some plans able to be asynchronously bound to and some not.

I think we should tighten the language on the `binding_retrievable` definition.

Comment (mattmcneeney) [reply spec.md]: I think this is sensible, thanks!

File: spec.md (hunk 3 above)
Comment (mattmcneeney) [reply spec.md]: Done

**Comment by vaikas-google** [comment]
Is there anything else that needs to be done before we can move this forward? Service Catalog is working on implementing this and we would like to ensure this is agreed upon asap. Can we make the decisions during tomorrows call for the remaining items?
As far as I understand there's only one remaining item that's been lingering for awhile now?
GET of the Bindings vs some other mechanism?

**Comment by mattmcneeney** [comment]
@vaikas-google I totally agree. Let's get this discussed on today's call and get this marked as being validated so we can work on getting this supported in platforms.

**Comment by mattmcneeney** [comment]
Please all review this PR as we are planning on moving this to `validation through implementation` on next week's call.

File: spec.md [hunk 5]
```diff
@@ -1399,6 +1500,127 @@ Errors](#service-broker-errors).
...
+details see [Broker Errors](#service-broker-errors).
+
+##### Body
+
+For success responses, the following fields are supported. Others will be
+ignored. For error responses, see [Broker Errors](#service-broker-errors).
+
+| Response Field | Type | Description |
+| --- | --- | --- |
+| credentials | object | A free-form hash of credentials that can be used by applications or users to access the service. |
```
jmrodri: In the *Types of Binding* section of the spec it reads:
> Credentials are a set of information... If the Service Broker supports generation of credentials it MUST return credentials in the response for a request to create a Service Binding.

I think we should incorporate the MUST in the table. Most of the time I refer to the table, and example responses, to see what I need to return. 

Something like this:
> A free-form hash of credentials that can be used by applications or users to access the service. If the Service Broker supports generate of credentials it MUST return this in the response.

spec.md
```diff
@@ -1399,6 +1500,127 @@ Errors](#service-broker-errors).
...
+| syslog_drain_url | string | A URL to which logs MUST be streamed. `"requires":["syslog_drain"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the platform MUST consider the response invalid. |
+| route_service_url | string | A URL to which the platform MUST proxy requests for the address sent with `bind_resource.route` in the request body. `"requires":["route_forwarding"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the platform can consider the response invalid. |
+| volume_mounts | array-of-objects | An array of configuration for mounting volumes. `"requires":["volume_mount"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the platform can consider the response invalid. |
```
jmrodri: Same comment about `volume_mounts`. The *Types of Bindings* section reads:
>  A create binding response from one of these services MUST include volume_mounts.

but the table doesn't mention that this field MUST be returned if the broker supports volume mounts.

jmrodri:
Other than the above clarification points, I approve of this PR!

spec.md

spec.md
mattmcneeney: also done!

jmrodri:
LGTM

spec.md
```diff
@@ -590,9 +597,8 @@ code `422 Unprocessable Entity` and the following body
 ```
 
 If the query parameter described above is present, and the Service Broker
-executes the request asynchronously, the Service Broker MUST return the
-asynchronous response `202 Accepted`. The response body SHOULD be the same as
-if the Service Broker were serving the request synchronously.
```
duglin: we've dropped the statement about the response looking the same between async and sync - is that really ok? 

If I were to read the spec w/o any history of what was there in the past, I would probably return a 202 with just {} in the Body for async ops. Is that now ok even for provisioning? And, as a platform author I would probably expect that too.

spec.md
```diff
@@ -697,10 +703,82 @@ Errors](#service-broker-errors).
...
+  "description": "Creating service (10% complete)."
+}
+```
+
+## Polling Last Operation for Service Bindings
+
+When a broker returns status code `202 Accepted` for [Binding](#binding) or
+[Unbinding](#unbinding), the platform will begin polling the
+`/v2/service_instances/:instance_id/service_bindings/:binding_id/last_operation`
+endpoint to obtain the state of the last requested operation. The broker
```
duglin: Can we make the format of this consistent with provision polling? Meaning, a table with the expected fields, rather than a sentence as we have here.

spec.md
```diff
@@ -697,10 +703,82 @@ Errors](#service-broker-errors).
...
+response MUST contain the field `state` and MAY contain the field `description`.
+
+Valid values for `state` are `in progress`, `succeeded`, and `failed`. The
+platform will poll the `last_operation` endpoint as long as the broker returns
+`"state": "in progress"`. Returning `"state": "succeeded"` or
+`"state": "failed"` will cause the platform to cease polling and, in the case of
+a [Binding](#binding) request, information on the Service Binding can then
+immediately be fetched using the
+[Fetching a Service Binding](#fetching-a-service-binding) endpoint. The value
+provided for `description` will be passed through to the platform API client and
```
duglin: we don't have the same normative keywords here like we do for provisioning -we should be consistent.

spec.md
duglin: never mind - I was mixing up sections - this is consistent.

spec.md
duglin: grrr - we repeat ourselves w/o the normative language in both spots.  Ignore this, will try to remember to open a new PR to fix this.

spec.md
```diff
@@ -1005,19 +1083,19 @@ services do not need to implement the endpoint for bind requests.
 
 Credentials are a set of information used by an application or a user to
 utilize the service instance. If the Service Broker supports generation of
-credentials it MUST return `credentials` in the response for a request to
-create a Service Binding. Credentials SHOULD be unique whenever possible, so
-access can be revoked for each binding without affecting consumers of other
-bindings for the service instance.
+credentials it MUST return `credentials` in the response for a synchronous
```
duglin: This could be interpreted as only seeing the creds for sync, not async. I think its ok to keep it generic:
```
If the Service Broker supports generation of credentials it MUST return
credentials in the response to create a Service Binding. 
```
And people will know that in the async case the "real response" flows later - that's the point of async.

spec.md
```diff
@@ -1005,19 +1083,19 @@ services do not need to implement the endpoint for bind requests.
...
 
 #### Log Drain
 
 There are a class of service offerings that provide aggregation, indexing, and
 analysis of log data. To utilize these services an application that generates
 logs needs information for the location to which it will stream logs. A create
-binding response from a Service Broker that provides one of these services
-MUST include a `syslog_drain_url`. The platform MUST use the
-`syslog_drain_url` value when sending logs to the service.
+binding response for a synchronous request from a Service Broker that provides
```
duglin: again, I'd drop the "synchronous" part

spec.md
```diff
@@ -1050,8 +1128,8 @@ service did not include a `"requires":["route_forwarding"]` property.
 #### Volume Services
 
 There are a class of services that provide network storage to applications
-via volume mounts in the application container. A create binding response
-from one of these services MUST include `volume_mounts`.
+via volume mounts in the application container. A create binding response for a
+synchronous request from one of these services MUST include `volume_mounts`.
```
duglin: let's drop the "synchronous" qualifier

spec.md
```diff
@@ -1181,10 +1265,11 @@ Errors](#service-broker-errors).
 
 | Response Field | Type | Description |
 | --- | --- | --- |
-| credentials | object | A free-form hash of credentials that can be used by applications or users to access the service. |
-| syslog_drain_url | string | A URL to which logs MUST be streamed. `"requires":["syslog_drain"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the platform MUST consider the response invalid. |
+| credentials | object | A free-form hash of credentials that can be used by applications or users to access the service. MUST be returned for a synchronous request if the Service Broker supports generation of credentials. |
```
duglin: let's drop "synch" from all of these since its misleading - the properties apply to both sync and async

spec.md
duglin: I wonder if 409 is more appropriate:
409 Conflict
    Indicates that the request could not be processed because of conflict in the request, such as an edit conflict between multiple simultaneous updates.

spec.md
```diff
@@ -1399,6 +1500,127 @@ Errors](#service-broker-errors).
...
+```
+$ curl 'http://username:password@broker-url/v2/service_instances/:instance_id' -X GET -H "X-Broker-API-Version: 2.11"
+```
+
+### Response
+
+| Status Code | Description |
+| --- | --- |
+| 200 OK | The expected response body is below. |
+| 404 Not Found | MUST be returned if the service instance does not exist or if a provisioning operation is still in progress. The expected response body is `{}`. |
```
duglin: While I prefer 404's, we don't use 404's in other places due to some concern about proxy errors - is that not a concern here?

spec.md
```diff
@@ -1399,6 +1500,127 @@ Errors](#service-broker-errors).
...
+
+##### Body
+
+For success responses, a broker MUST return the following fields. For error
+responses, see [Broker Errors](#service-broker-errors).
+
+| Response field | Type | Description |
+| --- | --- | --- |
+| dashboard_url | string | The URL of a web-based management user interface for the service instance; we refer to this as a service dashboard. The URL MUST contain enough information for the dashboard to identify the resource being accessed (`9189kdfsk0vfnku` in the example below). Note: a broker that wishes to return `dashboard_url` for a service instance MUST return it with the initial response to the provision request, even if the service is provisioned asynchronously. |
+| parameters | JSON object | Configuration options for the service instance. Service brokers SHOULD ensure that the configuration parameters match the service instance update schema they provide to platforms via the [Catalog](#catalog-management) endpoint (this will enable platforms to offer additional features such as pre-populated form fields in UIs). If brokers cannot fetch all configuration parameters, they MUST NOT return any. |
```
duglin: I'm not sure about the "SHOULD ensure" sentence since we can't do much about it on the GET side of things - this is more of a check for the PUT operation. If PUT allows something bad to get thru then GET MUST return the same bad results - especially per the last sentence of this paragraph.

Also, we should be more clear and say that the response MUST match what was provided on the PUT - dropping params would be bad.

spec.md
```diff
@@ -1399,6 +1500,127 @@ Errors](#service-broker-errors).
...
+
+## Fetching a Service Binding
+
+If `"binding_retrievable" :true` is declared for a service in the
+[Catalog](#catalog-management) endpoint, brokers MUST support this endpoint for
+all services and plans that support bindings (`"bindable": true`) and this
+endpoint MUST be available immediately after the
+[Polling Last Operation for Service Bindings](#polling-last-operation-for-service-bindings)
+endpoint returns `"state": "succeeded"` for a [Binding](#binding) operation.
+Otherwise, platforms SHOULD NOT attempt to call this endpoint under any
```
duglin: Do we need the last sentence? Wouldn't they just get back a 404 and no harm is done?  Same question for GET on instances too I guess.

spec.md
duglin: +1 to a follow-on PR to remove it.

spec.md
duglin: Where is the MUST that was asked for? I don't see it in the above table.

spec.md
duglin: is the MUST someplace else in the doc now?

spec.md
```diff
@@ -1399,6 +1500,127 @@ Errors](#service-broker-errors).
...
+| syslog_drain_url | string | A URL to which logs MUST be streamed. `"requires":["syslog_drain"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the platform MUST consider the response invalid. |
+| route_service_url | string | A URL to which the platform MUST proxy requests for the address sent with `bind_resource.route` in the request body. `"requires":["route_forwarding"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the platform can consider the response invalid. |
+| volume_mounts | array-of-objects | An array of configuration for mounting volumes. `"requires":["volume_mount"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the platform can consider the response invalid. |
+| parameters | JSON object | Configuration options for the service instance. Service brokers SHOULD ensure that the configuration parameters match the Service Binding schema they provide to platforms via the [Catalog](#catalog-management) endpoint (this will enable platforms to offer additional features such as pre-populated form fields in UIs). If service brokers cannot fetch all configuration parameters, they MUST NOT return any. |
```
duglin: Can comment as in GET for instances.
Also, s/service instance/service binding/ in the first line.

duglin:
overall +1 to the design, just a few syntax comments - nothing that should really stop us from moving to the implementation phase

spec.md
mattmcneeney: Ah, missed that bit. I personally think it's clear enough in the types of binding section, but can add this in here too.

spec.md
mattmcneeney: Copied here too

spec.md
mattmcneeney: I hope so, it was only a SHOULD. Otherwise we can add text stating that async instance provision SHOULD return the same as sync, but leave this open for async bindings? But I'd prefer to just remove this...

spec.md
mattmcneeney: Changed!

spec.md
mattmcneeney: 👍

spec.md
mattmcneeney: 👍

spec.md
mattmcneeney: 👍

spec.md
mattmcneeney: 👍

spec.md
mattmcneeney: Do they? I thought for async, the GET endpoint should be used to fetch credentials?

spec.md
mattmcneeney: This is a comment for #159 I believe. Do we really want to not support 404s going forward because of possible proxy issues?

spec.md
mattmcneeney: re #159

spec.md
mattmcneeney: re #159

spec.md
mattmcneeney: re #159

mattmcneeney:
Thank you @duglin - most comments addressed (except [this one](https://github.com/openservicebrokerapi/servicebroker/pull/334#discussion_r150855739) which needs some discussion)

spec.md
duglin: I'm just worried about the dashboard_url. This change pushes us towards not returning it on the provision response during an async provision. While I would prefer this model, I wasn't sure how existing brokers would feel about this shift in direction within the v2 time frame.

spec.md
duglin: Yes, but I'd prefer if we didn't have to qualify so many sentences with "sync" vs "async" since it then leaves the reader wonder what should happen in the opposite case. Which means we get very verbose or repetitive. I think it would be cleaner to just say "this json is what is returned from this op. Whether you get it on the initial response (sync) or later during a GET (async) doesn't change the format of the json".

spec.md
duglin: I would love it if we could use 404's the way they were intended to be use :-)   but I'm just trying to be consistent with why we weren't allowed to use them in the past and why 410's were added.

spec.md
mattmcneeney: Yeah. We could explicity change the `dashboard_url` in the response table for provision to state that it MUST be returned immediately even in the async case. Do we want to lock ourselves into that though? Or maybe we feel that we already are, no matter what the wording says?

If CF wanted to get dashboard_url at a later date, it would have to use the GET endpoints, which are going to require a change to CF anyway to call those. But I guess for provision the only data is this URL so maybe we can just enforce that to be returned immediately. We haven't had much feedback from service authors on that front as far as I remember.

mattmcneeney: @duglin we actually already have that on line 922:
```
Note: a Service Broker that wishes to return `dashboard_url` for a service instance MUST return it with the initial response to the provision request, even if the service is provisioned asynchronously. If present, MUST be a non-empty string.
```

duglin: feels a little inconsistent to have a SHOULD in one spot (or implied SHOULD (or less) now), and a MUST in another.  Seems we need to first decide is the MUST should live on - which I suspect it should since that would be a breaking change. If so, then I think we may need to be more explicit that an async provision is not the same as async binding because async provision requires a non-empty response.

spec.md
jmrodri: I prefer 404s regardless of proxy issue.

mattmcneeney:
I'm moving this forward to the next stage following agreement on last night's call!

spec.md
```diff
@@ -719,7 +797,7 @@ If the successful response includes a `state` of `failed` then the Platform
 MUST send a deprovision request to the Service Broker to prevent an orphan
```
kibbles-n-bytes: Not just deprovision; in the case of a binding, it should also attempt to unbind.

mattmcneeney: Good spot, fixed!

spec.md
```diff
@@ -638,21 +644,16 @@ corresponding Service Instance - it is an implementation choice. Doing so would
...
+state of the last requested operation.
 
-Valid values for `state` are `in progress`, `succeeded`, and `failed`. The
-Platform will poll the `last_operation` endpoint as long as the Service Broker
-returns `"state": "in progress"`. Returning `"state": "succeeded"` or `"state":
-"failed"` will cause the Platform to cease polling. The value provided for
-`description` will be passed through to the Platform API client and can be used
-to provide additional detail for users about the progress of the operation.
+Returning `"state": "succeeded"` or `"state": "failed"` will cause the Platform
+to cease polling.
```
mattmcneeney: This is for simplification and removing redundant information

kibbles-n-bytes:
Documenting for those who haven't been part of the calls:

Google has had async binding support as part of the Google Cloud service broker since October, and we have been using it in conjunction with alpha async binding support in the Kubernetes Service Catalog platform since around the same time. On both fronts, the implementation went smoothly, helped greatly by the fact that it mostly mirrors the currently existing async instance flow. Early access users of the broker have had no issues, and we've given multiple public demos of the interaction that have all gone well.

spec.md
```diff
@@ -1394,14 +1491,128 @@ For success responses, the following fields are defined:
...
+
+Responses with any other status code will be interpreted as a failure.
+
+##### Body
+
+For success responses, the following fields are defined:
+
+| Response Field | Type | Description |
+| --- | --- | --- |
+| credentials | object | A free-form hash of credentials that can be used by applications or users to access the service. |
```
mattmcneeney: MUST be returned if the service binding was provisioned asynchronously.

spec.md
```diff
@@ -1394,14 +1491,128 @@ For success responses, the following fields are defined:
...
+| syslog_drain_url | string | A URL to which logs MUST be streamed. `"requires":["syslog_drain"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform MUST consider the response invalid. |
```
mattmcneeney: If the service declares `"requires":["syslog_drain"]` in the [Catalog](#catalog-management) endpoint, this MUST be returned if the Service Binding was provisioned asynchronously.

spec.md
```diff
@@ -1200,10 +1279,11 @@ For success responses, the following fields are defined:
...
 | --- | --- | --- |
-| credentials | object | A free-form hash of credentials that can be used by applications or users to access the service. |
-| syslog_drain_url | string | A URL to which logs MUST be streamed. `"requires":["syslog_drain"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform MUST consider the response invalid. |
+| credentials | object | A free-form hash of credentials that can be used by applications or users to access the service. MUST be returned for a synchronous request if the Service Broker supports generation of credentials. |
+| syslog_drain_url | string | A URL to which logs MUST be streamed. If the service declares `"requires":["syslog_drain"]` in the [Catalog](#catalog-management) endpoint, this MUST be returned for a synchronous request or the Platform can consider the response invalid. |
 | route_service_url | string | A URL to which the Platform MUST proxy requests for the address sent with `bind_resource.route` in the request body. `"requires":["route_forwarding"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
```
mattmcneeney: If the service declares `"requires":["route_forwarding"]` in the [Catalog](#catalog-management) endpoint, this MUST be returned for a synchronous request or the Platform can consider the response invalid.

spec.md
```diff
@@ -1394,14 +1491,128 @@ For success responses, the following fields are defined:
...
+
+##### Body
+
+For success responses, the following fields are defined:
+
+| Response Field | Type | Description |
+| --- | --- | --- |
+| credentials | object | A free-form hash of credentials that can be used by applications or users to access the service. |
+| syslog_drain_url | string | A URL to which logs MUST be streamed. `"requires":["syslog_drain"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform MUST consider the response invalid. |
+| route_service_url | string | A URL to which the Platform MUST proxy requests for the address sent with `bind_resource.route` in the request body. `"requires":["route_forwarding"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
```
mattmcneeney: If the service declares `"requires":["route_forwarding"]` in the [Catalog](#catalog-management) endpoint, this MUST be returned if the Service Binding was provisioned asynchronously.

spec.md
```diff
@@ -1394,14 +1491,128 @@ For success responses, the following fields are defined:
...
+| volume_mounts | array-of-objects | An array of configuration for mounting volumes. `"requires":["volume_mount"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
```
mattmcneeney: If the service declares `"requires":["volume_mount"]` in the [Catalog](#catalog-management) endpoint, this MUST be returned if the Service Binding was provisioned asynchronously.

spec.md
```diff
@@ -1200,10 +1279,11 @@ For success responses, the following fields are defined:
 
 | Response Field | Type | Description |
 | --- | --- | --- |
-| credentials | object | A free-form hash of credentials that can be used by applications or users to access the service. |
-| syslog_drain_url | string | A URL to which logs MUST be streamed. `"requires":["syslog_drain"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform MUST consider the response invalid. |
+| credentials | object | A free-form hash of credentials that can be used by applications or users to access the service. MUST be returned for a synchronous request if the Service Broker supports generation of credentials. |
```
mattmcneeney: MUST be returned for a synchronous request if the Service supports generation of credentials.

mattmcneeney:
@kibbles-n-bytes and co
I updated the [Fetching a Service Binding](https://github.com/mattmcneeney/servicebroker/blob/4e1cc88b5fa1dd92215d2b11d9b7fb5415859136/spec.md#fetching-a-service-binding) section to say that the relevant binding information MUST be returned if the binding was created asynchronously. Otherwise Platform's would have no way of guaranteeing they can get the credentials/syslog_drain_url/other info associated with a binding.

spec.md
```diff
@@ -1200,10 +1279,11 @@ For success responses, the following fields are defined:
...
-| volume_mounts | array-of-objects | An array of configuration for remote storage devices to be mounted into an application container filesystem. `"requires":["volume_mount"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
+| credentials | object | A free-form hash of credentials that can be used by applications or users to access the service. MUST be returned for a synchronous request if the Service Broker supports generation of credentials. |
+| syslog_drain_url | string | A URL to which logs MUST be streamed. If the service declares `"requires":["syslog_drain"]` in the [Catalog](#catalog-management) endpoint, this MUST be returned for a synchronous request or the Platform can consider the response invalid. |
+| route_service_url | string | A URL to which the Platform MUST proxy requests for the address sent with `bind_resource.route` in the request body. If the service declares `"requires":["route_forwarding"]` in the [Catalog](#catalog-management) endpoint, this MUST be returned for a synchronous request or the Platform can consider the response invalid. |
```
mattmcneeney: @duglin @pmorie etc Are you happy with the MUST in this description?

Before this change, we said that:
* if a broker returns `route_service_url`, then it MUST declare `"requires":["volume_mount"]` in its catalog

After this change, we say that:
* if a broker delcares `"requires":["volume_mount"]` in its catalog, it MUST return `route_service_url`

The same applies for `syslog_drain_url` and `volume_mounts`. Is everyone happy with this?

cc @jenspinney @samze

duglin: I think the desired intent is ok but the wording is a bit off. I think its trying to say:
- - - 
A Service Broker MUST NOT return a `route_service_url` unless the corresponding
service declares `"requires":["route_forwarding"]` in the [Catalog](#catalog-management)
endpoint and the Platform has requested it by including a `bind_resource.route`
in the request body.
- - - 
Or something like that.  Right?

mattmcneeney: I think I added something close to this on the latest commit. See what you think...

spec.md
```diff
@@ -1184,6 +1262,7 @@ $ curl http://username:password@service-broker-url/v2/service_instances/:instanc
...
 | 400 Bad Request | MUST be returned if the request is malformed or missing mandatory data. |
 | 409 Conflict | MUST be returned if a Service Binding with the same id, for the same Service Instance, already exists but with different parameters. |
 | 422 Unprocessable Entity | MUST be returned if the Service Broker requires that `app_guid` be included in the request body. The response body MUST contain error code `"RequiresApp"` (see [Service Broker Errors](#service-broker-errors)). The error response MAY include a helpful error message in the `description` field such as `"This Service supports generation of credentials through binding an application only."`. Additionally, if the Service Broker rejects the request due to a concurrent request to create a binding for the same Service Instance, then this error MUST be returned (see [Blocking Operations](#blocking-operations)). |
```
Samze: We should expand this to include the "AsyncRequired" error code in addition to "RequiresApp".

spec.md
```diff
@@ -1307,23 +1391,36 @@ The following HTTP Headers are defined for this operation:
...
 
 ### Response
 
 | Status Code | Description |
 | --- | --- |
 | 200 OK | MUST be returned if the binding was deleted as a result of this request. The expected response body is `{}`. |
+| 202 Accepted | MUST be returned if the unbinding is in progress. This triggers the Platform to poll the [Polling Last Operation for Service Bindings](#polling-last-operation-for-service-bindings) endpoint for operation status. Note that a re-sent `DELETE` request MUST return a `202 Accepted`, not a `200 OK`, if the unbinding request has not completed yet. |
 | 400 Bad Request | MUST be returned if the request is malformed or missing mandatory data. |
-| 410 Gone | MUST be returned if the binding does not exist. The expected response body is `{}`. |
+| 410 Gone | MUST be returned if the binding does not exist. |
```
Samze: We should add 422 here detailing the "AsyncRequired" error code case.

spec.md
mattmcneeney: Done, thanks!

spec.md
mattmcneeney: Done, thanks!

jmrodri:
@mattmcneeney LGTM

nmaslarski:
Hi colleagues,
We have some kind of a dilemma here about the orphan mitigation of async bindings:
According to the table under the orphan mitigation section in case of `200 with malformed response` we should not perform orphan mitigation. There is also one statement above the table stating:
`the Platform SHOULD attempt to delete resources it cannot be sure were successfully created`

In case of async binding, we make a call to 
`GET /v2/service_instances/:instance_id/service_bindings/:binding_id/last_operation`
to see wether the resource has been created. And then another call to
`GET /v2/service_instances/:instance_id/service_bindings/:binding_id`
to get the actual credentials. The second endpoint returns `Status Code 200 OK`, and in case of malformed credentials according to the table we should not perform orphan mitigation. However without credentials the binding is not really successfully created, also in case of synchronous binding we do perform orphan mitigation if we get malformed credentials.

I am not sure wether the spec is clear enough for handling this case.

Best Regards,
Niki

n3wscott:
@nmaslarski It sounds like the broker you are interacting with is not written to the spec. I am understanding `malformed credentials` to mean they are empty in the GET response? If so, that broker should return `404 - MUST be returned if the Service Binding does not exist or if a binding operation is still in progress. ` Not `200`.

Can you link to the line you reference with `200 with malformed response` because I can only find `400 Bad Request - MUST be returned if the request is malformed or missing mandatory data.`

nilebox:
@nmaslarski @n3wscott I don't think we have orphan mitigation for async instances, nor should we have it for async bindings.
i.e. the initial request ends with `202 Accepted` that is a healthy response code. And it doesn't matter what codes are returned by `last_operation` polling. If the binding fails, the broker should continue returnung `in progress` until it cleans up, and only then it should return `failed` status.

nilebox:
@nmaslarski Also to add to above: we only perform orphan mitigation for `PUT` ("create") requests, and there is no orphan mitigation for `GET` and `PATCH` (update).

nmaslarski:
@n3wscott @nilebox 
Hi, thanks for the answers!
I'm not sure if I presented my problem correctly.
Yes, I am talking about the case where the broker is not compliant with the spec, but the question was what should we do from a platform perspective in this case.
Lets say, the `last_operation` endpoint returns `state: succeed` at some point, but then the endpoint for fetching the service binding is not compliant? Lets say it returns a `string` for credentials instead of `JSON object`. How should the platform treat this? In my opinion this is a broken binding, even though the `last_operation` says it is successful

nilebox:
> In my opinion this is a broken binding, even though the last_operation says it is successful

I agree that this is a broken binding. It's up to platform how to react to this, I think. Service Catalog will most probably mark the "create binding" operation as failed, without retrying or cleanup. It's up to the user then to act on a broken binding.

Most likely the scenario for your example will be:
1. Binding is marked as failed with "Invalid credentials" error message
2. Developer of the broker will issue a bugfix release.
3. User will need to manually delete a broken binding and create a new one.

Samze:
I think this also goes beyond a badly coded broker. The general question is what should a platform do if it cannot obtain a bind result.

My understanding of orphan mitigation is that is triggered if something has gone wrong AND the response indicates that the broker may have created a resource. So `40x` responses are not mitigated as they indicate a client error and so it is assumed the broker never created a resource, while `50x` are. I guess the table in the spec was written to assume that the resource creation would be captured in a single request.

For asynchronous bindings this is the first time the platform will need to make three calls in-order to satisfy a bind request (async instances only needs 2):
1. Create binding
1. Poll last operation
1. Get binding

The spec defines orphan mitigation for failure scenarios if the create fails or if last_operation results in a failure state. If get binding failures, we know that the broker has reported it has successfully created a resource but the platform cannot obtain a result for it. If we orphan mitigate in the other cases, I do not see why we would not orphan mitigate in this case.

nilebox:
> The spec defines orphan mitigation for failure scenarios if the create fails or if last_operation results in a failure state.

@Samze sorry, can you please point to a place in the spec where it says that `last_operation` resulting in a `failed` state requires orphan mitigation?

As I said above, currently we only do orphan mitigation in response to the errors occuring in `PUT` ("create") requests, and if the broker responds with `202 Accepted` (i.e. no error), it's a responsibility of the broker to cleanup if the asynchronous operation fails from now on.

Samze:
Hey @nilebox, sure thing.

`
If the successful response includes a state of failed then the Platform MUST send a deprovision request to the Service Broker to prevent an orphan being created on the Service Broker. However, while the Platform will attempt to send a deprovision request, Service Brokers MAY automatically delete any resources associated with the failed bind request on their own.
` [link](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#body-1)

There is an [equivalent statement](https://github.com/mattmcneeney/servicebroker/blob/10c34aca3593049190b50570d4ce770ad6fb741c/spec.md#body-2) for unbinding on an async binding last_operation failure.

kibbles-n-bytes:
@nmaslarski @Samze I think whether this falls into the bucket of "orphan mitigation" is a bit of a red herring. Orphan mitigation doesn't actually specify the timing of when the deletion request needs to occur; it is just telling the Platform "hey, don't just assume these things didn't get successfully created; make sure they're gone if you're going to forget about them". And then the text from #353 is trying to additionally say "hey, don't just assume these failed things didn't leave any broken artifacts on the broker; make sure everything's cleaned up if you're going to forget about them". It's unfortunate that #353 uses the term "orphan" but isn't part of the orphan mitigation section, but they're subtly different, so it's more an ambiguous choice of words.

The end result here that we're actually worried about is whether a deletion request should *ever* be sent to the broker. And in that case, I think the answer is clear: the Platform knows that *something* happened on the broker side that was not a rejected request, and so a deletion request would need to be sent at some point before the Platform forgets about the resource. That could be immediately, or that could be when the user explicitly requests to delete the resource.

We had talked about this at the last OSB F2F, and ultimately decided to get rid of the orphan mitigation section in favor of more clear failure and cleanup explanations. I got started on that here: #407 , but haven't yet gotten around to completing it. I'll make sure to get that fixed up and finished before the async binding PR goes in, but in the meantime I'm 100% confident that the Platform in this case should at some time send a delete request for the resource if it wants to get rid of it, and that the timing for it is up to the Platform.

mattmcneeney:
Note to self:
- [ ] Remember to update the [workflows](https://github.com/openservicebrokerapi/servicebroker/blob/master/diagram.md) diagram in the next release

duglin:
I can't remember, what's the status of this?

spec.md
```diff
@@ -1380,10 +1463,10 @@ For success responses, the following fields are defined:
 
 | Response Field | Type | Description |
 | --- | --- | --- |
-| credentials | object | A free-form hash of credentials that can be used by applications or users to access the service. |
-| syslog_drain_url | string | A URL to which logs MUST be streamed. `"requires":["syslog_drain"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform MUST consider the response invalid. |
-| route_service_url | string | A URL to which the Platform MUST proxy requests for the address sent with `bind_resource.route` in the request body. `"requires":["route_forwarding"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
-| volume_mounts | array-of-objects | An array of configuration for mounting volumes. `"requires":["volume_mount"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
+| credentials | object | A free-form hash of credentials that can be used by applications or users to access the service. MUST be returned if the Service Broker supports generation of credentials and the Service Binding was provisioned asynchronously. |
```
mattmcneeney: I can see this new line being a point of contention. What about brokers that don't want to hold onto credentials or only want to return them once for security issues? I know we've discussed this before but can't remember what we all agreed on?

nilebox: > return them once

@mattmcneeney I don't think it's realistic to "return only once". What about retries from the platform side?

If brokers want to remain stateless, they still have an option of creating binding synchronously. But even in that case create request with unchanged parameters should lead to `200 OK` response with implied unchanged credentials (rather than generating new credentials on retry) according to the spec.

spec.md
```diff
@@ -763,13 +756,87 @@ For success responses, the following fields are defined:
...
+$ curl http://username:password@broker-url/v2/service_instances/:instance_id/service_bindings/:binding_id/last_operation
+```
+
+### Response
+
+| Status Code | Description |
+| --- | --- |
+| 200 OK | MUST be returned upon successful processing of this request. The expected response body is below. |
+| 400 Bad Request | MUST be returned if the request is malformed or missing mandatory data. |
+| 410 Gone | Appropriate only for asynchronous delete operations. The Platform MUST consider this response a success and remove the resource from its database. The expected response body is `{}`. Returning this while the Platform is polling for [Binding](#binding) operations SHOULD be interpreted as an invalid response and the Platform SHOULD continue polling. |
```
Samze: `Returning this while the Platform is polling for [Binding](#binding) operations SHOULD be interpreted as an invalid response and the Platform SHOULD continue polling.`

Should be 

`Returning this while the Platform is polling for create operations SHOULD be interpreted as an invalid response and the Platform SHOULD continue polling.`

spec.md
```diff
@@ -763,13 +756,87 @@ For success responses, the following fields are defined:
...
+| --- | --- | --- |
+| state* | string | Valid values are `in progress`, `succeeded`, and `failed`. While `"state": "in progress"`, the Platform SHOULD continue polling. A response with `"state": "succeeded"` or `"state": "failed"` MUST cause the Platform to cease polling. |
+| description | string | A user-facing message that can be used to tell the user details about the status of the operation. If present, MUST be a non-empty string. |
+
+\* Fields with an asterisk are REQUIRED.
+
+```
+{
+  "state": "in progress",
+  "description": "Creating service (10% complete)."
```
Samze: Creating binding (10% complete).

mattmcneeney:
Needs checking over from the svc-cat team to make sure this is aligned with their implementation. We are happy with this from the CF side. So... it's review time!

spec.md
```diff
@@ -680,21 +680,16 @@ corresponding Service Instance - it is an implementation choice. Doing so would
...
 directed at the same Service Instance if concurrent updates are not supported.
 
-## Polling Last Operation
+## Polling Last Operation for Service Instances
 
 When a Service Broker returns status code `202 Accepted` for
 [Provision](#provisioning), [Update](#updating-a-service-instance), or
 [Deprovision](#deprovisioning), the Platform will begin polling the
 `/v2/service_instances/:instance_id/last_operation` endpoint to obtain the
-state of the last requested operation. The Service Broker response MUST contain
```
duglin: For those following later on..... we removed this sentence to remove redundant info.

spec.md
```diff
@@ -763,13 +756,87 @@ For success responses, the following fields are defined:
...
+to cease polling and, in the case of a [Binding](#binding) request, information
+regarding the Service Binding can then immediately be fetched using the
+[Fetching a Service Binding](#fetching-a-service-binding) endpoint.
+
+### Request
+
+#### Route
+`GET /v2/service_instances/:instance_id/service_bindings/:binding_id/last_operation`
+
+`:instance_id` MUST be a globally unique non-empty string.
```
duglin: Technically, this should be:
`:instance_id` is the identifier of a previously provisioned instance.
same for binding_id

spec.md
```diff
@@ -763,13 +756,89 @@ For success responses, the following fields are defined:
...
+$ curl http://username:password@broker-url/v2/service_instances/:instance_id/service_bindings/:binding_id/last_operation
+```
+
+### Response
+
+| Status Code | Description |
+| --- | --- |
+| 200 OK | MUST be returned upon successful processing of this request. The expected response body is below. |
+| 400 Bad Request | MUST be returned if the request is malformed or missing mandatory data. |
+| 410 Gone | Appropriate only for asynchronous delete operations. The Platform MUST consider this response a success and remove the resource from its database. The expected response body is `{}`. Returning this while the Platform is polling for create operations SHOULD be interpreted as an invalid response and the Platform SHOULD continue polling. |
```
duglin: In other spots you removed "The expected response body is `{}`.".  Should we do that here too?

spec.md
```diff
@@ -1247,7 +1328,7 @@ binding requests.
 
 #### cURL
 ```
-$ curl http://username:password@service-broker-url/v2/service_instances/:instance_id/service_bindings/:binding_id -d '{
+$ curl http://username:password@service-broker-url/v2/service_instances/:instance_id/service_bindings/:binding_id?accepts_incomplete=true -d '{
```
zrob: it's interesting that the example curl here now includes an optional query param

the provision curl request does not include it.

do we care either way?

spec.md
```diff
@@ -1270,9 +1351,10 @@ $ curl http://username:password@service-broker-url/v2/service_instances/:instanc
 | --- | --- |
 | 200 OK | MUST be returned if the binding already exists and the requested parameters are identical to the existing binding. The expected response body is below. |
 | 201 Created | MUST be returned if the binding was created as a result of this request. The expected response body is below. |
+| 202 Accepted | MUST be returned if the binding is in progress. The `operation` string MUST match that returned for the original request. This triggers the Platform to poll the [Polling Last Operation for Service Bindings](#polling-last-operation-for-service-bindings) endpoint for operation status. Information regarding the Service Binding (i.e. credentials) MUST NOT be returned in this asynchronous request. Note that a re-sent `PUT` request MUST return a `202 Accepted`, not a `200 OK`, if the binding is not yet fully created. |
```
duglin: s/asynchronous request/asynchronous response/    -  maybe?

spec.md
```diff
@@ -1450,23 +1543,37 @@ The following HTTP Headers are defined for this operation:
 
 ```
 $ curl 'http://username:password@service-broker-url/v2/service_instances/:instance_id/
-  service_bindings/:binding_id?service_id=service-id-here&plan_id=plan-id-here' -X DELETE -H "X-Broker-API-Version: 2.13"
+  service_bindings/:binding_id?service_id=service-id-here&plan_id=plan-id-here&accepts_incomplete=true' -X DELETE -H "X-Broker-API-Version: 2.13"
```
zrob: similar as previous note about including optional query params in curl example

spec.md
```diff
@@ -1286,10 +1368,11 @@ For success responses, the following fields are defined:
...
-| credentials | object | A free-form hash of credentials that can be used by applications or users to access the service. |
-| syslog_drain_url | string | A URL to which logs MUST be streamed. `"requires":["syslog_drain"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform MUST consider the response invalid. |
-| route_service_url | string | A URL to which the Platform MUST proxy requests for the address sent with `bind_resource.route` in the request body. `"requires":["route_forwarding"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
-| volume_mounts | array of [VolumeMount](#volume-mount-object) objects | An array of configuration for remote storage devices to be mounted into an application container filesystem. `"requires":["volume_mount"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
+| credentials | object | A free-form hash of credentials that can be used by applications or users to access the service. MUST be returned for a synchronous request if the Service Broker supports generation of credentials. |
```
duglin: s/sync request/sync response/
I think

n3wscott: yes, as a followup?

spec.md
```diff
@@ -1286,10 +1368,11 @@ For success responses, the following fields are defined:
...
+| syslog_drain_url | string | A URL to which logs MUST be streamed. If the service declares `"requires":["syslog_drain"]` in the [Catalog](#catalog-management) endpoint, this MUST be returned for a synchronous request or the Platform can consider the response invalid. |
```
duglin: I don't thin the "if" sentence is right since what you wrote now implies that if the "requires" is there then this field MUST be there which isn't true since it might not want to return one for this.

spec.md
```diff
@@ -1286,10 +1368,11 @@ For success responses, the following fields are defined:
...
+| route_service_url | string | A URL to which the Platform MUST proxy requests for the address sent with `bind_resource.route` in the request body. If the service declares `"requires":["route_forwarding"]` in the [Catalog](#catalog-management) endpoint and `"bind_resource":{"route":"some-address.com"}` is included in the request, this MUST be returned for a synchronous request or the Platform can consider the response invalid. |
+| volume_mounts | array of [VolumeMount](#volume-mount-object) objects | An array of configuration for remote storage devices to be mounted into an application container filesystem. If the service declares `"requires":["volume_mount"]` in the [Catalog](#catalog-management) endpoint, this MUST be returned for a synchronous request or the Platform can consider the response invalid. |
```
duglin: same comment as above, I don't think the "if" is worded correctly.

spec.md
```diff
@@ -1349,15 +1432,22 @@ can be mounted on all app instances simultaneously.
 
 If `"bindings_retrievable" :true` is declared for a service in the
 [Catalog](#catalog-management) endpoint, Service Brokers MUST support this
-endpoint for all plans of the service.
+endpoint for all services and plans that support bindings (`"bindable": true`)
+and this endpoint MUST be available immediately after the
+[Polling Last Operation for Service Bindings](#polling-last-operation-for-service-bindings)
+endpoint returns `"state": "succeeded"` for a [Binding](#binding) operation.
+Otherwise, Platforms SHOULD NOT attempt to call this endpoint under any
```
duglin: why not?  Wouldn't they just get a 404 if its not ready/there?

spec.md
mattmcneeney: done

spec.md
mattmcneeney: done

spec.md
mattmcneeney: done

spec.md
mattmcneeney: all done

spec.md
mattmcneeney: removed

spec.md
zrob: nevermind, curl examples are the same, was looking at the top level request example

spec.md
zrob: nevermind, curl examples are the same, was looking at the top level request example

zrob:
lgtm

duglin:
LGTM

zrob:
lgtm

fmui:
LGTM

PR #335, opened 2017-10-11
Remove "web-friendly" definition since its never actually used

duglin:
Removing a sentence about "web-friendly" fields since the term
"web-friendly" is actually never used.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -296,11 +296,11 @@ $ curl http://username:password@service-broker-url/v2/catalog -H "X-Broker-API-V
 
 #### Body
 
-CLI and web clients have different needs with regard to service and plan names.
-A CLI-friendly string is all lowercase, with no spaces.
-Keep it short -- imagine a user having to type it as an argument for a longer
-command.
-A web-friendly display name is camel-cased with spaces and punctuation supported.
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -300,7 +300,6 @@ CLI and web clients have different needs with regard to service and plan names.
 A CLI-friendly string is all lowercase, with no spaces.
 Keep it short -- imagine a user having to type it as an argument for a longer
 command.
-A web-friendly display name is camel-cased with spaces and punctuation supported.
```
vaikas-google: Just a thought, while I'm all for removing this, I wonder if we should rejigger the previous sentences because it talks about web clients, yet then only talks about CLI friendly string.
Maybe just give a definition of a CLI friendly string.
In the following sections CLI-friendly name is all lowercase, with no spaces. It should be kept short -- imagine a user having to type it as an argument for a longer command.

duglin: good idea - working on it

vaikas-google:
LGTM

avade:
LGTM

angarg12:
LGTM

mattmcneeney:
LGTM

PR #341, opened 2017-10-17
Terminology: Service Binding

n3wscott:
Finding usages of mixed case Service Bindings and correcting it.

Adding a tool to help prevent "service bindings" in the future.

### DIFF
--- a/profile.md
+++ b/profile.md
@@ -35,12 +35,6 @@ The terminology defined below are defined by the
 for convenience. If there are any inconsistencies between their definitions in
 this document and the [Open Service Broker API specification](spec.md), then
 the [Open Service Broker API specification](spec.md) SHALL take precedence.
-
-- *Service Binding*: Represents the request to use a Service Instance. As part
-  of this request there might be a reference to the entity, also known as the
-  Application, that will use the Service Instance. Service Bindings will often
-  contain the credentials that can then be used to communicate with the Service
-  Instance.
  
 - *Platform*: The software that will manage the cloud environment into which
   Applications and Service Brokers are provisioned.  Users will not directly
@@ -58,6 +52,12 @@ the [Open Service Broker API specification](spec.md) SHALL take precedence.
   and Service Bindings.
 
 - *Service Instance*: An instantiation of a Service offering.
+
+- *Service Binding*: Represents the request to use a Service Instance. As part
+  of this request there might be a reference to the entity, also known as the
+  Application, that will use the Service Instance. Service Bindings will often
+  contain the credentials that can then be used to communicate with the Service
+  Instance.
 
 - *Application*: The software that uses a Service Instance via Service Binding.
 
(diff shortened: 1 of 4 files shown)


n3wscott:
I will hold off adding more PRs until this is merged, I want to use the tool. :D

tools/verify-terms.sh
```diff
@@ -0,0 +1,99 @@
...
+  # echo scanning $file
+  dir=$(dirname $file)
+
+  [[ -n "$verbose" ]] && echo "> $file"
+
+  lineNum=0
+  cat ${file} | while read line; do 
+    ((lineNum++)) || true
+
+	for rfc in "Service Binding"; do
```
duglin: Would it make sense to also check for the other terms like "Service Instance", etc... ?

mattmcneeney: +1 to this. It may also be better to add the tool to a separate commit once the terms are fixed on master, so that the build will always pass on master (if we add service instance to this without fixing master, it would fail)

n3wscott: This is true. But it would cause failures at the moment. Here is the plan:

I am going to work through all the terms defined and Caps them. And in that PR add the term to this list. But I am going to do them one term at a time. Then future PRs will not be able to add uncapped terms!

n3wscott: If we need to I can break out the script in another PR. I felt like it was easy enough to understand together. Please advise.

profile.md
```diff
@@ -53,10 +53,11 @@ the [Open Service Broker API specification](spec.md) SHALL take precedence.
...
 
-- *Service Binding*: The representation of an association between an
-  Application and a Service Instance. Often, Service Bindings, will
-  contain the credentials that the Application will use to communicate
-  with the Service Instance.
+- *Service Binding*: Represents the request to use a Service Instance. As part
+  of this request there might be a reference to the entity, also known as the
+  Application, that will use the Service Instance. Service Bindings will often
+  contain the credentials that can then be used to communicate with the Service
+  Instance.
```
avade: Suggested edit. Can discuss on the call.

*Service Binding*: Represents the intent to use a Service Instance. As part
 of a service binding there might be a reference to the entity, also known as the
 Application, that will use the Service Instance. Service Bindings will often
 contain the credentials and co-ordianates that can then be used to communicate 
 with the Service Instance.

n3wscott: Note this is in profile, and we should have a followup PR to sync the term tables from the spec and profile docs.

vaikas-google:
LGTM

profile.md
avade: I will PR the profile doc to reference the spec for the definitions. 

WDYT of my proposed text?

angarg12:
LGTM

duglin:
LGTM

pmorie:
LGTM

PR #343, opened 2017-10-18
Feedback Requested: Move json examples from spec text

n3wscott:
Here is my thinking:

 - Json examples in the spec are huge and distract from the text. 
 - Json files could be used to host a fake broker and act as the responses for a live swagger demo.
 - We can run a json validation script on the .json files to validate we are making valid json by hand.

### DIFF
--- a/json/v2/catalog-get.json
+++ b/json/v2/catalog-get.json
@@ -1,0 +1,122 @@
+{
+  "services": [{
+    "name": "fake-service",
+    "id": "acb56d7c-XXXX-XXXX-XXXX-feb140a59a66",
+    "description": "fake service",
+    "tags": ["no-sql", "relational"],
+    "requires": ["route_forwarding"],
+    "bindable": true,
+    "metadata": {
+      "provider": {
+        "name": "The name"
+      },
+      "listing": {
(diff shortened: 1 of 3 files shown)


angarg12:
LGTM

I like the idea of JSON validation.

duglin:
I like the idea of being able to do json validation but I’m not sure of pulling it out of the spec. I think that might hurt readability by forcing people to click to see the samples.  I would prefer if we created a tool to do the validation while it’s in the spec.  We should be able to find and extract it pretty easily.

n3wscott:
I am suggesting removing the json blobs because when you print this spec, there are 2+ pages of json that is redundant from the defines above it. An example is useful but the target audience of this spec can construct a json blob in their mind while reading, or they can click the link.

A potentially more compelling example would be a side repo that you could launch a fake broker and hit it with swagger and get these canned responses back.

Another reason for removing the json blobs is to allow support for non-json communication in the future. For example I can see support for gRPC in the future for these types of services.

avade:
I worry that pulling these out of the spec will mean they get forgotten and outdated.

We have also had feedback from broker authors that inline request/response examples are a very useful reference to develop against.

PR #354, opened 2017-10-27
Add text around what brokers should do upon failures

duglin:
Provide more guidance for what should happen when things go wrong
so we can ensure a more consistent semantics and interop.
    
This should only be reviewed after #353 - so only look at the 2nd commit.
    
### DIFF
--- a/spec.md
+++ b/spec.md
@@ -723,7 +723,9 @@ For success responses, the following fields are defined:
 
 If the successful response includes a `state` of `failed` then the Platform
 MUST send a deprovision request to the Service Broker to prevent an orphan
-being created on the Service Broker.
+being created on the Service Broker. However, while the Platform will attempt
+to send a deprovision request, Service Brokers MAY automatically delete
+any resources associated with the failed bind request on their own.
 
 ### Polling Interval and Duration
 
@@ -826,7 +828,9 @@ $ curl http://username:password@service-broker-url/v2/service_instances/:instanc
 
 Responses with any other status code will be interpreted as a failure and a
 deprovision request MUST be sent to the Service Broker to prevent an orphan
-being created on the Service Broker.
+being created on the Service Broker. However, while the platform will attempt
+to send a deprovision request, Service Brokers MAY automatically delete
+any resources associated with the failed provisioning request on their own.
 
 #### Body
 
@@ -977,6 +981,8 @@ $ curl http://username:password@service-broker-url/v2/service_instances/:instanc
 | 422 Unprocessable entity | MUST be returned if the requested change is not supported or if the request cannot currently be fulfilled due to the state of the Service Instance (e.g. Service Instance utilization is over the quota of the requested plan). Additionally, a `422 Unprocessable Entity` can also be returned if the Service Broker only supports asynchronous update for the requested plan and the request did not include `?accepts_incomplete=true`; in this case the response body MUST contain a error code `"AsyncRequired"` (see [Service Broker Errors](#service-broker-errors)). The error response MAY include a helpful error message in the `description` field such as `"This Service Plan requires client support for asynchronous service operations."`. |
 
 Responses with any other status code will be interpreted as a failure.
+When the response includes a 4xx status code, the Service Broker MUST NOT
+apply any of the requested changes to the Service Instance.
 
 #### Body
 
@@ -1169,7 +1175,9 @@ $ curl http://username:password@service-broker-url/v2/service_instances/:instanc
 
 Responses with any other status code will be interpreted as a failure and an
 unbind request MUST be sent to the Service Broker to prevent an orphan being
-created on the Service Broker.
+created on the Service Broker. However, while the platform will attempt
+to send an unbind request, Service Brokers MAY automatically delete
+any resources associated with the failed bind request on their own.
 
 #### Body
 


duglin:
On hold until we resolve #353

spec.md
```diff
@@ -718,6 +718,12 @@ For success responses, the following fields are valid.
 }
 ```
 
+If the successful response includes a `state` of `failed` then the Platform
+MUST send a deprovision request to the Service Broker to prevent an orphan
+being created on the Service Broker.  However, while the platform will attempt
```
mattmcneeney: I'd capitalise Platform (multiple occurences in this PR) now to save conflicts when #356 is merged

spec.md
```diff
@@ -1290,9 +1308,9 @@ $ curl 'http://username:password@service-broker-url/v2/service_instances/:instan
 | 410 Gone | MUST be returned if the binding does not exist. The expected response body is `{}`. |
 
 Responses with any other status code will be interpreted as a failure and the
-binding will remain in the marketplace database. Service brokers can include a
-user-facing message in the `description` field; for details see [Service
-Broker Errors](#service-broker-errors).
+binding MUST remain in the marketplace database.
```
mattmcneeney: Is this the first time we've used "marketplace database"? Can we rephrase this to something like:
```
Responses with any other status code will be interpreted as a failure and the Platform MUST continue to remember the Service Binding.
```

spec.md

spec.md
duglin: its not the only time we talk about DBs for the marketplace. Let me see if I can sneak in that change....

duglin: actually, let me do that in a new PR since its more than just a one word change and might require some thinking by the reviewers.

duglin: See: https://github.com/openservicebrokerapi/servicebroker/pull/362

mattmcneeney: Thank you! :)

mattmcneeney:
LGTM

angarg12:
@duglin this PR needs rebase now. I will give my LG after it :)

duglin:
rebased

angarg12:
LGTM

duglin:
I know there was some concern about this text:
```
However, while the Platform will attempt
to send a deprovision request, Service Brokers MAY automatically delete
any resources associated with the failed bind request on their own.
```
so, if those concerns are still there would using this text (someplace in the spec) help?
```
When the creation of a resource fails, or when a resource is deleted, the Service Broker
SHOULD eventually clean up all artifacts related to that resource. The exact timing of when
that occurs is an implementation detail. However, from the perspective of the OSB APIs
the clean up SHOULD be immediate.
```
IMO this leaves it open for the resource to live on for a while if the broker has a good reason
too - but per the OSB API it really needs to appear like it was deleted, otherwise the API is
just lying to the platform.

Thoughts?

mattmcneeney:
Regarding this text:
```
However, while the Platform will attempt
to send a deprovision request, Service Brokers MAY automatically delete
any resources associated with the failed bind request on their own.
```

As a platform author, my interpretation of that is:
* I should send a deprovision request when something fails
* But in reality, that deprovision request may well fail as the broker may have already done some cleanup
* I'm fairly confident that this means that the cleanup has completed, but if that deprovision request I make fails, I should probably check that it failed because the cleanup has already happened, not because something has gone wrong

duglin:
I'm ok with that interpretation :-)
I would think most platforms would treat a 404/410 from a DELETE as a good thing and move on.

spec.md
```diff
@@ -962,7 +966,12 @@ $ curl http://username:password@service-broker-url/v2/service_instances/:instanc
...
 | 422 Unprocessable entity | MUST be returned if the requested change is not supported or if the request cannot currently be fulfilled due to the state of the Service Instance (e.g. Service Instance utilization is over the quota of the requested plan). Service Brokers SHOULD include a user-facing message in the body; for details see [Service Broker Errors](#service-broker-errors). Additionally, a `422 Unprocessable Entity` can also be returned if the Service Broker only supports asynchronous update for the requested plan and the request did not include `?accepts_incomplete=true`; in this case the expected response body is: `{ "error": "AsyncRequired", "description": "This Service Plan requires client support for asynchronous service operations." }` (see [Service Broker Errors](#service-broker-errors). |
 
-Responses with any other status code will be interpreted as a failure. Service Brokers can include a user-facing message in the `description` field; for details see [Service Broker Errors](#service-broker-errors).
+Responses with any other status code will be interpreted as a failure and
+the Service Broker MUST NOT apply any of the requested changes to the
```
n3wscott: How does the service broker deal with a long running, multi step update that fails near the end? It had to modify things to keep track of state in the migration, but was unable to complete the update. Do we require that the broker rolls all the steps back? 

Perhaps this MUST NOT should be a SHOULD NOT ?

duglin: My concern with SHOULD NOT is that it then becomes indeterminate and we can't know what state the broker left things. Which, sort of, means we can't use the instance upon failure and will be forced to delete it.

avade: I have a concern about the case where the broker returns a 500, as the resource could have been put into an unknown state. The same goes for timeout scenarios.

For 4xx I agree that we should state MUST NOT.

duglin: per our last call, I made the MUST NOT linked to a 4xx status code. I'll open a new issue to discuss if we can say anything about 5xx's.

n3wscott:
LGTM

Note this PR will conflict with https://github.com/openservicebrokerapi/servicebroker/pull/409.

angarg12:
Which one do you think will be easier to merge? This one last?

duglin:
I'm ok with #409 going first since it has more edits

duglin:
rebased

angarg12:
Travis is failing

> tools/../spec.md - 726: should only have a single space after a period

@duglin the check is working as intended :smile:

duglin:
:-)
fixed

duglin:
reviews needed

mattmcneeney:
LGTM

avade:
Happy with these changes.

LGTM

duglin:
2 more needed. ping @fmui @angarg12

duglin:
2 more needed

fmui:
LGTM

vaikas-google:
LGTM

PR #355, opened 2017-11-01
Remove terms from profile, link to spec terminology

n3wscott:
We had talked about making a new file for terms and linking from both, I think they should be in the spec like this. Thoughts?


duglin:
Even though I liked having a copy of them in here, I can live with just the ref.
LGTM

avade:
LGTM

fmui:
LGTM

mattmcneeney:
LGTM

Issue #361, opened 2017-11-06
Feature Request: Diagram


cholick:
Old documentation for the spec used to include an overview of the API calls (happy path) in a sequence diagram:
https://docs.pivotal.io/pivotalcf/1-8/services/api.html#api-overview

The short sequence diagram with an instance and binding's lifecycle was a really clear way to get a handle on what was going on quickly.

Could the spec repo include a similar/updated diagram?

duglin:
I'm sure we could probably abstract it so its not CF specific - sounds like a good idea. Anything to help clarify things for people....

mattmcneeney:
We have a diagram now, so closing this issue.

PR #364, opened 2017-11-08
Add diagram of workflows

mattmcneeney:
Following on from #361 and the discussion last night's call, here is a PR that adds a diagram of the platform to service broker workflows to the repository.

The [diagram](https://docs.google.com/drawings/d/1LDGED4dm-ets5Z_xQLzadnR42hgjyNMJj8-CpXO5fnc/edit) is a Google Drawing hosted in the [OSBAPI](https://drive.google.com/drive/u/0/folders/0B427Up4C9IE0VmM0ZlhHTG1Rc0E) Google Drive folder which all PMC members should be able to edit. The new `diagram.md` file automatically [renders the live version](https://github.com/mattmcneeney/servicebroker/blob/91104e5de1ff1e18f143e7f08789aab588540882/diagram.md) of the linked drawing. I'm hoping this is a sensible approach as it means nobody has to learn any new 'drawing languages'.

I also updated `CONTRIBUTING.md` to ensure the workflow diagram is updated on a spec release (which I presume is the most sensible time that we will do this).

Feedback is welcome!

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,8 @@
 
 [Read the profile](https://github.com/openservicebrokerapi/servicebroker/blob/v2.13/profile.md)
 
+[View the workflows](diagram.md)
+
 [Release Notes for past versions](https://github.com/openservicebrokerapi/servicebroker/blob/v2.13/release-notes.md)
 
 ## Working Changes
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -133,6 +133,9 @@ creating a new PR with the title of the proposed release. For example,
   `X-Broker-API-Version` headers) updated. Do not update the header
   `Open Service Broker API (master - might contain changes that are not yet released)`
   (this will be done if and when the release proposal is approved).
+  * Update [diagram.md](diagram.md) to show the new version number in the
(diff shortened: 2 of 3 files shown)


avade:
you could use a relative URL for the diagram so that Travis doesn't fail?

angarg12:
LGTM

avade:
LGTM

README.md
```diff
@@ -7,6 +7,8 @@
 
 [Read the profile](https://github.com/openservicebrokerapi/servicebroker/blob/v2.13/profile.md)
 
+[View the workflows](diagram.md)
```
duglin: I wonder if we should have the appropriate part of the workflow added to the right section of the spec too? Or at least a pointer from the spec to the entire thing. Right now if you just read the spec you wouldn't even know the picture existed.

duglin:
LGTM

README.md
avade: I would +1 having the diagram at the bottom and linked to from the index

avade:
LGTM

fmui:
LGTM

vaikas-google:
LGTM

PR #367, opened 2017-11-15
remove marketplace and use correct case for Platform

mattmcneeney:
Because I like causing myself pain... but really because I think a "marketplace" is a Platform implementation detail, and we shouldn't have it as part of the specification.

I also took the opportunity to change all occurrences of `platform` to `Platform` as it is a defined term in the spec.

### DIFF
--- a/profile.md
+++ b/profile.md
@@ -122,7 +122,7 @@ Platform and the Service Brokers.
 
 The list of properties within the Context Object can vary depending on
 which Service Broker API is being invoked and which Platform is being used.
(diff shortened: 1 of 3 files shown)


mattmcneeney:
@n3wscott you should like this one 😄

duglin:
Overall +1 but CI failed

angarg12:
LGTM

vaikas-google:
LGTM

duglin:
LGTM

pmorie:
LGTM

Issue #360, opened 2017-11-06
Orphan mitigation when provisioning asynchronously?


kibbles-n-bytes:
The orphan mitigation section of the spec only details synchronous situations. However, it seems to me like the platform giving up on an async provisioning operation because it hasn't received a `"succeeded"` or `"failed"` state should also trigger orphan mitigation.

duglin:
The https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#polling-interval-and-duration section talks about it going into a "failed" state, but then stops. If we added text in there about the failure triggering the orphan mitigation logic then I think that would be good.

avade:
SGTM

mattmcneeney:
SGTM too. @kibbles-n-bytes are you able to open a PR for this?

kibbles-n-bytes:
@mattmcneeney Yes, I will have one out today!

mattmcneeney:
I believe this is done!

PR #368, opened 2017-11-16
clarify orphan mitigation situations when provisioning asynchronously

kibbles-n-bytes:
Explicitly mention in both the polling and orphan mitigation sections what situations in which the platform should initiate orphan mitigation during an asynchronous provisioning operation.

Closes: #360

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -726,7 +726,9 @@ being created on the Service Broker.
 
 The frequency and maximum duration of polling MAY vary by platform client. If
 a platform has a max polling duration and this limit is reached, the platform
-MUST cease polling and the operation state MUST be considered `failed`.
+MUST cease polling, the operation state MUST be considered `failed`, and the
+Platform SHOULD send a deprovision request to the Service Broker to prevent an
+orphan being created.
 
 ## Provisioning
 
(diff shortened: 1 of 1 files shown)


mattmcneeney:
LGTM !

kibbles-n-bytes:
Seems Travis was complaining because it didn't understand "should" meaning "if" vs "should" meaning "SHOULD". :stuck_out_tongue: Fixed.

However, while re-reading this I noticed some inconsistencies with how we're using MUST vs SHOULD, so I'm not 100% about which should be used in this circumstance. I opened issue #370 to discuss this.

angarg12:
FYI I kept the needs work tag until we clarify the issue.

spec.md
```diff
@@ -1472,6 +1474,10 @@ Platforms SHOULD initiate orphan mitigation in the following scenarios:
 | 5xx | Service Broker error | Yes |
 | Timeout | Failure | Yes |
 
+In addition, during asynchronous provisioning, if the Platform has a max polling
+duration and this limit is reached, the Platform SHOULD initiate orphan
+mitigation.
```
duglin: Can we get a link to the orphan section added here?

spec.md
```diff
@@ -726,7 +726,9 @@ being created on the Service Broker.
 
 The frequency and maximum duration of polling MAY vary by platform client. If
 a platform has a max polling duration and this limit is reached, the platform
-MUST cease polling and the operation state MUST be considered `failed`.
+MUST cease polling, the operation state MUST be considered `failed`, and the
+Platform SHOULD send a deprovision request to the Service Broker to prevent an
```
duglin: I'd prefer if we do what you did down below and just say "SHOULD initiate orphan mitigation" with a link to that section. Then we're not duplicating what orphan mitigation means in 2 places.

kibbles-n-bytes:
Closing in favor of new language clarifying the situation around all stray resources.

Issue #160, opened 2017-04-04
Own the swagger definition for the Service Broker API


avade:
### Proposal

I propose that the OSABPI WG own the [Swagger](http://swagger.io/} definition for the service broker API that is currently in the [Cloud Foundry Incubator project](https://github.com/cloudfoundry-incubator/cf-swagger).

I have [raised this idea with the project lead Dr Max](https://github.com/cloudfoundry-incubator/cf-swagger/issues/6#issuecomment-291285111) and he thought this was a *good idea*™

#### Why Swagger? 

> Swagger makes it possible for API providers to create interactive API documentation where developers can not only learn specifics about an API, but also experiment with, test and debug API calls. Swagger can be implemented in many different ways including Scala, Java, HTML5, JavaScript, Ruby and PHP. The Swagger core is available for download at GitHub.
>
> Automated API documentation that is consistent, easy to read and provides the ability to make live API calls is beneficial to both API providers and API consumers. Thorough, interactive API documentation helps developers better understand and easily consume APIs and also aids in API discovery. For API providers, offering APIs that are easy to understand and consume helps in the promotion of their APIs and can increase the adoption and consumption of their APIs by developers.

https://www.programmableweb.com/news/swagger-interactive-api-documentation-framework-benefits-both-api-providers-and-consumers/2013/02/19

### Questions

* Would we maintain the current spec markdown or move to swagger generated markdown?

angarg12:
What would happen if we don't transfer it? Would it be kept and maintain by CF, or dropped?

Seems reasonable to me that we own the Swagger definition if we own the spec itself.

avade:
**What would happen if we don't transfer it? Would it be kept and maintain by CF, or dropped? **

I would imagine that the CF team would continue to maintain it.

pmorie:
I would support a swagger spec for this API, but it is important to preserve the context and textual description present in the current spec.md.  I think of swagger/openapi as replacing the markdown tables in the current spec documents.

angarg12:
To follow on @pmorie comment, I would like to avoid duplicate information to reduce overhead and the danger of divergence. Would it be an option to move the API definition to Swagger and keep md for the textual spec definition?

Also, if we give it a go, how would the transition look like? We will need to rebase it to the latest API version, I can take that task myself.

mattmcneeney:
We could link from the swagger spec to a contextual document using the `External Documentation Object` (from [here](http://swagger.io/specification/)).

[...]

PR #389, opened 2017-12-02
Add Open API document for OSB API

n3wscott:
Relates to https://github.com/openservicebrokerapi/servicebroker/issues/160

Ready.


n3wscott:
@rhodie27 Yes, there is an open bug with swagger editor not pulling in parameters from components correctly. https://github.com/swagger-api/swagger-ui/issues/3976  https://github.com/swagger-api/swagger-ui/issues/3558

But I think what is in this PR is correct based on the Open API spec. The swagger UI works correctly.

I will double check the spec... I think I am correct and the tools have not caught up to the new stuff in 3.0.0. https://swagger.io/docs/specification/components/

mattmcneeney:
Rendered version: http://petstore.swagger.io/?url=https://raw.githubusercontent.com/n3wscott/servicebroker/9bc40f42ecaa3ab89fe632404a3eb56f15fb03b8/openapi.yaml

fmui:
Here are some random nit-picking comments.

* In the spec we several places with wording like this: "MUST be a non-empty string". "minLength: 1" should cover this.
* In the schema definition Service -> properties -> requires, items should be an array of enums (syslog_drain, route_forwarding, volume_mount).
* The schema of the status code '409' of 'PUT /v2/service_instances/{instance_id}' could be an Error, not an Object.
* The status code '400' of 'PUT /v2/service_instances/{instance_id}/service_bindings/{binding_id}' is lacking a content definition.
* The definitions of "route" and "*_url" properties and could be enhanced by "format: url".
* The conventions for service and plan metadata are not in the spec, but we link to them from the spec. Should we do something similar here?

n3wscott:
@fmui 

> In the spec we several places with wording like this: "MUST be a non-empty string". "minLength: 1" should cover this.

I would vote for no validation in the first version of the swagger spec. It might be useful to not have the validation in to allow folks to poke at bad brokers. If we still want validation, could it be a followup exercise?

> The schema of the status code '409' of 'PUT /v2/service_instances/{instance_id}' could be an Error, not an Object.

The spec says MAY for an error and expected to be `{}`:

```
| 409 Conflict | MUST be returned if a Service Instance with the same id already exists but with different attributes. The expected response body is `{}`, though the description field MAY be used to return a user-facing error message, as described in [Service Broker Errors](#service-broker-errors).|
 ```

> The conventions for service and plan metadata are not in the spec, but we link to them from the spec. Should we do something similar here?

I opted to only use the spec.


I have updated based on feedback, adding the missing return types and enums, but I am hesitating to add more validation to the doc just because it is hard to know when to stop. I think it could be a valuable addition to the Open API document, but it is still possible to do as a followup and we could do it one resources

File: openapi.yaml [hunk 1]
```diff
@@ -0,0 +1,668 @@
...
+            type: boolean
+      requestBody:
+        description: parameters for the requested service instance update
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/ServiceInstanceUpdateRequest'
+      responses:
+        '200':
```
avade: This is missing the `operation` field that is valid in the response of PATCH on Service Instances

openapi.yaml
```diff
@@ -0,0 +1,668 @@
...
+          schema:
+            type: string
+        - name: accepts_incomplete
+          in: query
+          description: asynchronous deprovision supported
+          schema:
+            type: boolean
+      responses:
+        '200':
+          description: OK
```
avade: This is missing the operation field that is valid in the response of DELETE on Service Instances

fmui:
So, what's the meaning of this excerpt:
> | 409 Conflict | MUST be returned if a Service Instance with the same id already exists but with different attributes. The expected response body is `{}`, though the description field MAY be used to return a user-facing error message, as described in [Service Broker Errors](#service-broker-errors).|

1. You can add a description, but you better shouldn't. The platform doesn't expect it.
2. This a broker error, but you don't have to provide a description.

n3wscott:
@fmui it means that 409 must at least return `{}` and optionally return a service broker error.

openapi.yaml
n3wscott: ah yes, true. Ref: https://github.com/openservicebrokerapi/servicebroker/blob/v2.13/spec.md#body-5

openapi.yaml
n3wscott: yes yes: https://github.com/openservicebrokerapi/servicebroker/blob/v2.13/spec.md#body-9

n3wscott:
@avade thanks for the review! I will fix those missing return params up when I next get a chance.

openapi.yaml
n3wscott: done.

openapi.yaml
n3wscott: done.

openapi.yaml
```diff
@@ -0,0 +1,675 @@
+# This is the Open API interface for Open Service Broker API. Every attempt will
+# be made to make the Open API version of OSB API accurately represent the
+# written specification. If the spec and this document conflict, the spec is
+# the authroity.
+
+openapi: "3.0.0"
+
+info:
+  title: Open Service Broker API
+  description: The Service Broker API defines an HTTP(S) interface between Platforms and Service Brokers.
```
mattmcneeney: The **Open** Service Broker API...

avade:
LGTM

openapi.yaml
```diff
@@ -0,0 +1,675 @@
...
+servers:
+  - url: http://example.com
+    description: Broker Endpoint
+  - url: http://localhost:3000
+    description: OSB Checker - MockOSB
+
+paths:
+  /v2/catalog:
+    get:
+      summary: get the service catalog
```
mattmcneeney: Suggestion: "get the catalog of services that the service broker offers"

openapi.yaml
```diff
@@ -0,0 +1,675 @@
...
+          application/json:
+            schema:
+              $ref: '#/components/schemas/ServiceInstanceProvisionRequest'
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ServiceInstanceProvision'
```
mattmcneeney: HTTP 200 responses shouldn't return the `operation` field

openapi.yaml
```diff
@@ -0,0 +1,675 @@
...
+        '201':
+          description: Created
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ServiceInstanceProvision'
```
mattmcneeney: HTTP 201 responses also shouldn't return the `operation` field

openapi.yaml
```diff
@@ -0,0 +1,675 @@
...
+        '202':
+          description: Accepted
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/AsyncOperation'
```
mattmcneeney: HTTP 202 responses should include the `dashboard_url` as this is the only time it can be returned in the current spec

openapi.yaml
```diff
@@ -0,0 +1,675 @@
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Object'
+        '409':
+          description: Conflict
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Object'
```
mattmcneeney: HTTP 409 responses MAY contain a `description` field

openapi.yaml
```diff
@@ -0,0 +1,675 @@
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/AsyncOperation'
+        '400':
+          description: Bad Request
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Object'
```
mattmcneeney: I'm not 100% sure, but I think `error` and `description` field should be allowed to be returned with a 400

openapi.yaml
```diff
@@ -0,0 +1,675 @@
...
+        '202': 
+          description: Accepted
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/AsyncOperation'
+        '400':
```
mattmcneeney: I'm not 100% sure, but I think `error` and `description` field should be allowed to be returned with a 400

openapi.yaml
```diff
@@ -0,0 +1,675 @@
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Object'
+        '202':
+          description: Accepted
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/AsyncOperation'
+        '400':
```
mattmcneeney: I'm not 100% sure, but I think `error` and `description` field should be allowed to be returned with a 400

openapi.yaml
```diff
@@ -0,0 +1,675 @@
...
+        - $ref: '#/components/parameters/APIVersion'
+        - name: instance_id
+          in: path
+          description: instance id of instance to find last operation applied to it
+          required: true
+          schema:
+            type: string
+        - name: service_id
+          in: query
+          description: id of the service being used
```
mattmcneeney: Suggestion: `id of the service associated with the instance`

openapi.yaml
```diff
@@ -0,0 +1,675 @@
...
+          schema:
+            type: string
+        - name: plan_id
+          in: query
+          description: id of the plan being used
```
mattmcneeney: Suggestion: `id of the plan associated with the instance`

openapi.yaml
```diff
@@ -0,0 +1,675 @@
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ServiceInstanceLastOperation'
+        '400':
+          description: Bad Request
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Object'
```
mattmcneeney: I'm not 100% sure, but I think `error` and `description` field should be allowed to be returned with a 400

openapi.yaml
```diff
@@ -0,0 +1,675 @@
...
+      summary: generation of a service binding
+      tags:
+        - ServiceBindings
+      operationId: serviceBinding.binding
+      parameters:
+        - $ref: '#/components/parameters/APIVersion'
+        - $ref: '#/components/parameters/OriginatingIdentity'
+        - name: instance_id
+          in: path
+          description: instance id being bound
```
mattmcneeney: Suggestion: `instance id of instance to create a binding on`

openapi.yaml
```diff
@@ -0,0 +1,675 @@
...
+          required: true
+          schema:
+            type: string
+        - name: binding_id
+          in: path
+          description: binding id being bound
```
mattmcneeney: Suggestion: `binding id of binding to create` (or something else - I'm not sure the current description makes sense)

openapi.yaml
```diff
@@ -0,0 +1,675 @@
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ServiceBinding'
+        '400':
+          description: Bad Request
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Object'
```
mattmcneeney: I'm not 100% sure, but I think `error` and `description` field should be allowed to be returned with a 400

openapi.yaml
```diff
@@ -0,0 +1,675 @@
...
+        '409':
+          description: Conflict
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
```
mattmcneeney: The spec suggests that a 409 can only return the `description` field (not the `error` field)

openapi.yaml
```diff
@@ -0,0 +1,675 @@
...
+            type: string
+        - name: binding_id
+          in: path
+          description: id of the binding being deleted
+          required: true
+          schema:
+            type: string
+        - name: service_id
+          in: query
+          description: id of the service associated with the binding being deleted
```
mattmcneeney: Suggestion: `id of the service associated with the instance for which a binding is being deleted`

openapi.yaml
```diff
@@ -0,0 +1,675 @@
...
+          required: true
+          schema:
+            type: string
+        - name: plan_id
+          in: query
+          description: id of the plan associated with the binding being deleted
```
mattmcneeney: Suggestion: `id of the plan associated with the instance for which a binding is being deleted`

openapi.yaml
```diff
@@ -0,0 +1,675 @@
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Object'
+        '400':
+          description: Bad Request
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Object'
```
mattmcneeney: I'm not 100% sure, but I think `error` and `description` field should be allowed to be returned with a 400

openapi.yaml
```diff
@@ -0,0 +1,675 @@
...
+          type: array
+          items:
+            type: string
+            enum:
+              - syslog_drain
+              - route_forwarding
+              - volume_mount
+        bindable:
+          type: boolean
+        metadata:
```
mattmcneeney: Is there an easy way to link to the [Service Metadata Conventions](https://github.com/openservicebrokerapi/servicebroker/blob/master/profile.md#service-metadata) here? Maybe via a `description`?

openapi.yaml
```diff
@@ -0,0 +1,675 @@
...
+        - plan_id
+        - organization_guid
+        - space_guid
+      properties:
+        service_id:
+          type: string
+        plan_id:
+          type: string
+        context:
+          $ref: '#/components/schemas/Object'
```
mattmcneeney: Is there an easy way to link to the [Context Conventions](https://github.com/openservicebrokerapi/servicebroker/blob/master/profile.md#context-object) here? Maybe via a `description`?

openapi.yaml
```diff
@@ -0,0 +1,675 @@
...
+          type: string
+        operation:
+          type: string # could be a link object to last operation
+
+    ServiceInstanceUpdateRequest:
+      required:
+        - service_id
+      properties:
+        context:
+          $ref: '#/components/schemas/Object'
```
mattmcneeney: Is there an easy way to link to the [Context Conventions](https://github.com/openservicebrokerapi/servicebroker/blob/master/profile.md#context-object) here? Maybe via a `description`?

openapi.yaml
```diff
@@ -0,0 +1,675 @@
...
+        description:
+          type: string
+
+    ServiceBindingRequest:
+      required:
+        - service_id
+        - plan_id
+      properties:
+        context:
+          $ref: '#/components/schemas/Object'
```
mattmcneeney: Is there an easy way to link to the [Context Conventions](https://github.com/openservicebrokerapi/servicebroker/blob/master/profile.md#context-object) here? Maybe via a `description`?

openapi.yaml
n3wscott: This was a mistake, AsyncOperation was suppose to be used for only the update, not the create.

openapi.yaml
n3wscott: This is true. My argument is it might not hurt anything to allow an optional field be defined in the Open API doc for the 201 case. the 200 and 202 cases are the same object.

openapi.yaml
n3wscott: The 200 could apply to the sync or async cases.

openapi.yaml
n3wscott: This is what the spec says: 

> 400 -> MUST be returned if the request is malformed or missing mandatory data.

And then

> Service Brokers can include a user-facing message in the description field;

I struggled with this TBH, the spec says some responses should `{}` some should be in the `{ "error":"" , "description":""}` form and some it does not say. The line below [the table](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#response-2) says the quote above but it is not clear is that applies to all or some.

So I opted to define the response as Object at least, and where the spec calls it out, define the response as Error. 

Perhaps this is something we update after the doc has some usage if it is not working for folks or is unclear.

openapi.yaml
n3wscott: Looking at the spec, for PATCH it says this:

>  200 --> MUST be returned if the request's changes have been applied. The expected response body is {}.

So the spec does not say the `operation` field is allowed.

openapi.yaml
n3wscott: I think if the spec intends Error to be optionally returned for any error case it needs to be reworked. This is the language:

> Service Brokers can include a user-facing message in the description field;

There is no spec language in that sentence, like MAY, and it is unclear is the broker author is allowed to return the description field for any responses or just for responses that call out they return errors.


I made an [issue](https://github.com/openservicebrokerapi/servicebroker/issues/403) to clear this up.

openapi.yaml
n3wscott: I made an [issue](https://github.com/openservicebrokerapi/servicebroker/issues/403) to clear this up.

openapi.yaml
n3wscott: Thanks for these suggestions. I was not in a wordsmithing state when writing the doc and was hoping someone would help me out here.

openapi.yaml
n3wscott: I made an [issue](https://github.com/openservicebrokerapi/servicebroker/issues/403) to clear this up.

openapi.yaml
n3wscott: I made an [issue](https://github.com/openservicebrokerapi/servicebroker/issues/403) to clear this up.

openapi.yaml
n3wscott: This is not how I read the spec.

I made an [issue](https://github.com/openservicebrokerapi/servicebroker/issues/403) to clear this up. 

The Error object has both fields optional.

openapi.yaml
n3wscott: Not a bad idea. Added.

openapi.yaml
n3wscott: done.

openapi.yaml
n3wscott: done

openapi.yaml
mattmcneeney: Ah. I hadn't considered returning the id of a completed async operation (when a service instance has successfully been provisioned async).

openapi.yaml
mattmcneeney: Yeah, but that could make platform authors write logic to handle polling when they get a 201 which _should_ never be used.

openapi.yaml
mattmcneeney: 👍

openapi.yaml
mattmcneeney: Sounds good. If we can use this PR as an opportunity to clear up ambiguities in the spec, then I'd love for any error to be able to return the defined error object :)

openapi.yaml
mattmcneeney: @n3wscott I agree; `operation` shouldn't be returned here as a 200 implies the update has completed synchronously.

openapi.yaml
mattmcneeney: Awesome, thanks!

openapi.yaml
mattmcneeney: 👍

openapi.yaml
mattmcneeney: 👍

openapi.yaml
mattmcneeney: Tracking in #403

openapi.yaml
mattmcneeney: Did you spot this one too @n3wscott ?

openapi.yaml
```diff
@@ -0,0 +1,681 @@
...
+  schemas:
+    Catalog:
+      type: object
+      properties:
+        services:
+          type: array
+          items:
+            $ref: '#/components/schemas/Service'
+
+    Service:
```
luksa: I believe this and all others need to include the `type: object` attribute, otherwise you're saying they can also be arrays or primitives. At least that's how it was in OpenAPI 2.0.

openapi.yaml
```diff
@@ -0,0 +1,681 @@
+# This is the Open API interface for Open Service Broker API. Every attempt will
+# be made to make the Open API version of OSB API accurately represent the
+# written specification. If the spec and this document conflict, the spec is
+# the authroity.
```
luksa: s/authroity/authority/

openapi.yaml
n3wscott: Nice catch!

avade:
I propose that this is now merged and further refinements are new PRs 

LGTM

angarg12:
Would it make sense to put a link to the rendered swagger on the repo README so that it is easily accessible by people?

angarg12:
LGTM

n3wscott:
> Would it make sense to put a link to the rendered swagger on the repo README so that it is easily accessible by people?

Sounds like a good follow-up. Maybe even a symlink to a git tag when next release happens.

openapi.yaml
duglin: if I'm following this thread, upon an async provision a 202 is returned and it'll include the "operation". I don't think a 200 is ever returned for an async provision.  Or, if you're talking about the case of a resent provision and they asked for async, and its already done, then yes we'll get a 200 but w/o "operation" in the response body.

openapi.yaml
duglin: I agree with @mattmcneeney. While technically we could have the open api doc include lots of random properties because there's an extensibility point that allows them, I think people will treat this doc as a pseudo-spec and so we should try to be as precise as possible. Which means only including spec-specified properties.

openapi.yaml
```diff
@@ -0,0 +1,700 @@
...
+servers:
+  - url: http://example.com
+    description: Broker Endpoint
+  - url: http://localhost:3000
+    description: OSB Checker - MockOSB
+
+paths:
+  /v2/catalog:
+    get:
+      summary: get the catalog of services that the service broker offers
```
duglin: Can we start these with capital letters?  Get the catalog...
It looks better to me when generated.

duglin:
I'm new to OAPI but is there a way in the "Example Value" that it generates to have it do something like add a "?" when a field is optional?  Showing that example is really helpful, but I'm missing the required/optional aspect of each field.

duglin:
@dr.max ^^

duglin:
@maximilien ^^

n3wscott:
> is there a way in the "Example Value" that it generates to have it do something like add a "?" when a field is optional

@duglin This is a quirk of the default swagger ui, click the tab next to "example value" called "Schema" and you will see the object in table format with the required fields with descriptions if the model has them.

openapi.yaml
n3wscott: We can in a follow-up!

duglin:
Ah, I didn't realize the "Model" thing was a link - nice

vaikas-google:
LGTM

duglin:
@n3wscott did you want to fix the two spots @mattmcneeney commented on?

n3wscott:
> did you want to fix the two spots @mattmcneeney commented on?

@duglin the error pr is not in yet I think

duglin:
Sorry, not sure what you mean by "error pr"

duglin:
Matt are your issues resolved? github shows you're still requesting changes

duglin:
@n3wscott I think there are a few edits that people want to see before we merge this - any chance of getting those in?

mattmcneeney:
LGTM

[Approved with PullApprove](https://pullapprove.com/openservicebrokerapi/servicebroker/pull-request/389/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=openservicebrokerapi/servicebroker)

PR #391, opened 2017-12-04
Define all objects the same way.

n3wscott:
Update instance defines `previous_values`, and then defines the properties of `previous_values` in the same table. This does not follow the pattern in the spec. 

Making this consistent.


angarg12:
LGTM

avade:
LGTM

mattmcneeney:
LGTM

duglin:
LGTM

Issue #116, opened 2016-12-15
Ability to declare schema of binding credential output


bmelville:
Binding credentials are a generic map of data, and as such are difficult to reason about or program against in a meaningful way. Similar to issue #59, we would like to support the ability to declare the schema of credential outputs from a binding.

arschles:
@bmelville is there any documentation on how the schema works that I can look at?

angarg12:
Since binding is a generic operation itself, the output is arbitrary.

Therefore if we schematise the whole output, we will cover this use case (credentials) plus any other possible scenario.

bmelville:
@arschles if you take a look at issue #59 and its accompanying proposal doc, that should be a good start. Let me know if it doesn't satisfy your questions though.

bmelville:
I have an initial proposal for this: https://docs.google.com/document/d/1JbsJgqgNtqthcfYwK_KbS6C8sjElrZNgoLhu40dUPAs/edit?usp=sharing

duglin:
From 11/12 call, @n3wscott will write-up a proposal

pmorie:
Should this issue be in validating-through-implementation now that #392 is in that state?

mattmcneeney:
@pmorie I think so. Unless people were interested in having one schema for the API response to a create binding call, and another schema for the GET binding endpoint being implemented by #159.

mattmcneeney:
We spoke about this on today's OSBAPI weekly call. We think that a proposal similar to the `kinds` discussion we had a the last face-to-face is more powerful than this.

A proposal like this would provide a couple of benefits:
1. App developers could swap services easily if they knew that their application was going to work with services from a number of different providers (i.e. if there was a standard for what a MySQL service binding looked like)
1. Having standard binding structures would make it easier for developers to build libraries to consume service bindings (similar to the above), making it easier to build applications

We are waiting for an owner to volunteer to pick this one up, as we are currently tied up with the [Extensions](https://docs.google.com/document/d/1OKlsQKUab3yvOyQjvh8HFFcIfkILyFMaq0xngTdTcek/edit?usp=sharing) work.

hsiliev:
Our group in SAP is currently working with CF and k8s brokers. Some of the brokers we use provide IaaS native AWS, GCP and Azure services. Credentials for the same backing service (PostgreSQL for instance) might be slightly different: `uri` vs `url`, `user` vs `username`, etc.

[...]

PR #392, opened 2017-12-05
Allow Broker authors to define the JSON Schema for request output.

n3wscott:
Fixes https://github.com/openservicebrokerapi/servicebroker/issues/116

We found it useful to define the full output of responses vs defining just the credentials or disk info sections. This gives more flexibility to Broker authors.


spec.md
```diff
@@ -377,24 +377,31 @@ how Platforms might expose these values to their users.
 
 | Response field | Type | Description |
 | --- | --- | --- |
-| [create](#input-parameters-object) | object | The schema definition for creating a Service Instance. |
-| [update](#input-parameters-object) | object | The schema definition for updating a Service Instance. |
+| [create](#json-schema-objects) | object | The schema definition for creating a Service Instance. |
+| [update](#json-schema-objectst) | object | The schema definition for updating a Service Instance. |
```
angarg12: objectst -> typo

spec.md
```diff
@@ -377,24 +377,31 @@ how Platforms might expose these values to their users.
...
+| [create](#json-schema-objects) | object | The schema definition for creating a Service Binding. |
 
 
-##### Input Parameters Object
+##### JSON Schema Objects
+
+It is useful for the Platform for the Broker to define exactly what is expected
+when making requests inside opaque input request objects and unspecified output
+responses, that mechanism is JSON Schema Object definitions. This allows Broker
+authors to programaticly define inputs and outputs for applicable requests.
```
angarg12: programmatically ?

angarg12:
I like this :slightly_smiling_face: 

One typo is breaking the build so I'll wait until it's fixed.

spec.md
n3wscott: derp! thanks :D

spec.md
n3wscott: Thanks!

angarg12:
LGTM

spec.md
```diff
@@ -377,24 +377,31 @@ how Platforms might expose these values to their users.
...
-| [create](#input-parameters-object) | object | The schema definition for creating a Service Binding. |
+| [create](#json-schema-objects) | object | The schema definition for creating a Service Binding. |
 
 
-##### Input Parameters Object
+##### JSON Schema Objects
+
+It is useful for the Platform for the Broker to define exactly what is expected
+when making requests inside opaque input request objects and unspecified output
+responses, that mechanism is JSON Schema Object definitions. This allows Broker
```
khipkin: you separate two sentences with a comma here.

spec.md
```diff
@@ -377,24 +377,31 @@ how Platforms might expose these values to their users.
...
+authors to programmatically define inputs and outputs for applicable requests.
 
 | Response field | Type | Description |
 | --- | --- | --- |
-| parameters | JSON schema object | The schema definition for the input parameters. Each input parameter is expressed as a property within a JSON object. |
+| parameters | JSON schema object | The schema definition for the input `parameters`. Each input parameter is expressed as a property within a JSON object. |
+| response | JSON schema object | The schema definition for the output response. |
+
+The following rules apply if `parameters` or `response` included anywhere in the catalog:
```
khipkin: "The following rules apply if parameters or response **IS** included"...

n3wscott: Thanks! Using "are". Ex: If thing1 or thing2 are included. vs If thing1 or thing2 is included.

angarg12:
LGTM

spec.md
```diff
@@ -377,24 +377,32 @@ how Platforms might expose these values to their users.
...
 | Response field | Type | Description |
 | --- | --- | --- |
-| [create](#input-parameters-object) | object | The schema definition for creating a Service Binding. |
+| [create](#json-schema-objects) | object | The schema definition for creating a Service Binding. |
 
 
-##### Input Parameters Object
+##### JSON Schema Objects
+
+It is useful for the Platform for the Broker to define exactly what is expected
```
duglin: Is "Platform" here correct? I don't think you really mean "Platform" do you?  Perhaps just "It is useful for the Broker to define..."

n3wscott: It is not useful for the broker. I mean controller in the k8s side, but I think in the OSB spec we call that the `Platform`. So, yes I think I do mean `Platform` here. I want to say it is useful for the caller, the consumer of the broker API.

duglin: ok, can I suggest a rewording then since the "for... for..." confused me.
How about: `It is useful for the Platform if the Broker defines exactly what is expected...`  ?

n3wscott: how about `A Platform will find it useful for the Broker to define`

mattmcneeney:
@n3wscott I think we need a proper proposal for this one, outlining what the benefits are to both service broker authors and platform authors, guidance of how platforms could use this schema, etc.

n3wscott:
>outlining what the benefits are to both service broker authors and platform authors, guidance of how platforms could use this schema, etc.

There is a number of things we would like to do with this mechanism at google, and it all revolves around understanding the components that are returned from instance creation and binding. Binding more so. Because brokers are allowed to return anything at all from a Binding, we require a way to programmatically understand what the contents of the binding is.

It comes down to simply understanding what the response body is from the broker.

n3wscott:
A simple Design Doc for this proposal: https://docs.google.com/document/d/1pSLntUlkcQ7mqYE4BnGH3saFg4IbSGkYHlu_2-AgTmg/edit?usp=sharing

spec.md
```diff
@@ -377,24 +377,32 @@ how Platforms might expose these values to their users.
...
+A Platform will find it useful for the Broker to define exactly what is expected
+when making requests inside opaque input request objects and unspecified output
+responses, that mechanism is JSON Schema Object definitions. This allows Broker
+authors to programmatically define inputs and outputs for applicable requests.
 
 | Response field | Type | Description |
 | --- | --- | --- |
-| parameters | JSON schema object | The schema definition for the input parameters. Each input parameter is expressed as a property within a JSON object. |
+| parameters | JSON schema object | The schema definition for the input `parameters`. Each input parameter is expressed as a property within a JSON object. |
+| response | JSON schema object | The schema definition for the output response. |
```
mattmcneeney: Where do you think we may eventually add conventions for output responses? Ideally we would be able to encourage consistency across brokers for the common things platforms need to do when creating a binding i.e. changing networking/firewall policies.

n3wscott: >add conventions for output responses

I think that process should be the same process for agreement on generic actions. It is very much the same kind of community lead standardization. So... we just have to agree on how to get the generic actions out ! :D

spec.md
```diff
@@ -377,24 +377,32 @@ how Platforms might expose these values to their users.
...
 | Response field | Type | Description |
 | --- | --- | --- |
-| [create](#input-parameters-object) | object | The schema definition for creating a Service Binding. |
+| [create](#json-schema-objects) | object | The schema definition for creating a Service Binding. |
 
 
-##### Input Parameters Object
+##### JSON Schema Objects
+
+A Platform will find it useful for the Broker to define exactly what is expected
```
duglin: s/will/might/
s/Broker/Service Broker/

spec.md
```diff
@@ -377,24 +377,32 @@ how Platforms might expose these values to their users.
...
+when making requests inside opaque input request objects and unspecified output
+responses, that mechanism is JSON Schema Object definitions. This allows Broker
```
duglin: s/Broker/Service Broker/

spec.md
```diff
@@ -377,24 +377,32 @@ how Platforms might expose these values to their users.
...
 | --- | --- | --- |
-| [create](#input-parameters-object) | object | The schema definition for creating a Service Binding. |
+| [create](#json-schema-objects) | object | The schema definition for creating a Service Binding. |
 
 
-##### Input Parameters Object
+##### JSON Schema Objects
+
+A Platform will find it useful for the Broker to define exactly what is expected
+when making requests inside opaque input request objects and unspecified output
```
mattmcneeney: This paragraph reads funny to me. How about:
```
Platforms MAY be able to provide a better user experience if Service Brokers can programmatically define the inputs and outputs of requests using JSON Schema Object definitions.
```

mattmcneeney:
A few changes needed @n3wscott but other than that, I think the [proposal/design doc] and this PR look good!

Do you have a plan to add this to a specific Service Broker?

n3wscott:
> a specific Service Broker 

Yes, this is blocking progress on some of the features for the GCP Broker.

spec.md
n3wscott: Done.

spec.md
n3wscott: I like it. Added.

spec.md
n3wscott: done.

n3wscott:
Any objection for moving this PR to the next phase? `2 - validating through implementation`

spec.md
```diff
@@ -377,24 +377,31 @@ how Platforms might expose these values to their users.
...
 | Response field | Type | Description |
 | --- | --- | --- |
-| [create](#input-parameters-object) | object | The schema definition for creating a Service Binding. |
+| [create](#json-schema-objects) | object | The schema definition for creating a Service Binding. |
 
 
-##### Input Parameters Object
+##### JSON Schema Objects
+
+Platforms MAY be able to provide a better user experience if Service Brokers can
```
duglin: I think this "MAY" should be "might" since this isn't really something controlled by the spec - so its non-normative IMO.

duglin:
One minor request, otherwise looks good

pmorie:
cc @luksa

btbd:
Who is going this for validation?

duglin:
Who will implement this for validation?

duglin:
rebase needed

avade:
Implementation for service catalog will be driven by @luksa.

luksa:
We may want to specify whether the platform should check if the credentials returned by the broker match the response schema and what it should do if they don't.

spec.md
```diff
@@ -495,7 +504,26 @@ schema being used.
                   "type": "string"
                 }
               }
-            }
+            },
+            "response": {
+              "$schema": "http://json-schema.org/draft-04/schema#",
+              "type": "object",
+              "properties": {
+                "credentials" : {
```
pmorie: It would be extremely useful to have a way to indicate that a particular part of the bind response is secret vs. non-secret.  In k8s, we would be able to use this to differentiate storage for these keys into secret and configmap, respectively.

duglin:
rebase needed

spec.md
```diff
@@ -377,24 +377,31 @@ how Platforms might expose these values to their users.
...
+
+Platforms MAY be able to provide a better user experience if Service Brokers can
+programmatically define the inputs and outputs of requests using JSON Schema
+Object definitions.
 
 | Response field | Type | Description |
 | --- | --- | --- |
-| parameters | JSON schema object | The schema definition for the input parameters. Each input parameter is expressed as a property within a JSON object. |
+| parameters | JSON schema object | The schema definition for the input `parameters`. Each input parameter is expressed as a property within a JSON object. |
+| response | JSON schema object | The schema definition for the output response. |
```
duglin: As part of the /v2/catalog metadata we include the schema for the services and plans. Does it make more sense to include this schema as part of that response too? Additionally, including it here leaves open the possibility that it might change for each message, where if its part of the catalog metadata then its required to be static, at least until the entire catalog is refreshed.

luksa: Not sure what you mean by "here"? This _is_ the /v2/catalog metadata (specifically `services.plans.schemas`).

duglin: oh darn, sorry, I scrolled up and say "Service Binding Response" and thought this appeared in the Bind() response json - sorry - ignore my comment.

luksa:
I have two concerns:
- The spec implies that you can add the `response` schema to `service_instance.create` and `service_instance.update` in addition to `service_binding.create`, which was the only place where we actually need the response schema. The spec does not say whether a broker author can/should/must add the response schema in the two places I mentioned. Specifying the schema for those responses obviously shouldn't be required, since the OSBAPI spec itself defines what the response should be. 
- On a related note, the response schema defines the schema for the whole response, not just the returned `credentials`. This means there's overlap between what the OSBAPI spec prescribes and the broker-returned schema, so the question arises whether a broker that returns both `credentials` and e.g. `syslog_drain_url` must return the complete schema or just the schema for the `credentials` part? Technically, the schema in the [reference catalog response](https://github.com/n3wscott/servicebroker/blob/95910853f5287b7b647fce28e4165f42d3a5b715/spec.md#reference-catalog-response) in the OSBAPI spec states that the response will _not_ contain anything other than a `credentials` field (this is implied by the `type:object` property and the lack of either a `properties.syslog_drain_url` or an `additionalProperties:true` entry:
```
"response": {
  "type": "object",
  "properties": {
    "credentials": {
      "type": "object",
      "properties": {
        "token": {
          "description": "A token to be used to make fake API calls to the fake server.",
          "type": "string"
        },
        "uri": {
          "description": "The fake URI to use to make fake API calls against.",
          "type": "string"
        }
      }
    }
  }
}
```

spec.md
```diff
@@ -377,24 +377,31 @@ how Platforms might expose these values to their users.
...
 ##### Service Binding Object
 
 | Response field | Type | Description |
 | --- | --- | --- |
-| [create](#input-parameters-object) | object | The schema definition for creating a Service Binding. |
+| [create](#json-schema-objects) | object | The schema definition for creating a Service Binding. |
 
 
-##### Input Parameters Object
+##### JSON Schema Objects
```
luksa: This name is confusing, because it's in plural, but it actually denotes a single object containing multiple schemas. It should be renamed to _JSON Schemas Object_ or something along the lines of _Input/Output Schema Object_.

luksa:
I've just noticed that this PR doesn't include the necessary changes to https://github.com/openservicebrokerapi/servicebroker/blob/master/openapi.yaml

n3wscott:
> I've just noticed that this PR doesn't include the necessary changes to the OpenAPI schema 

The OpenAPI doc did not exist when this was created. I will update.

n3wscott:
@luksa 

> The spec implies that you can add the response schema to service_instance.create and service_instance.update in addition to service_binding.create, which was the only place where we actually need the response schema. The spec does not say whether a broker author can/should/must add the response schema in the two places I mentioned. Specifying the schema for those responses obviously shouldn't be required, since the OSBAPI spec itself defines what the response should be.

It was my explicit intention to allow service_instance.update and  service_binding.create to have defined responses. At the moment, current implementations do not use the response for update, but that will not be true in the future. None of this should be required, but it is my intention to allow some advanced behavior from platform/broker interactions with these hooks. 

> On a related note, the response schema defines the schema for the whole response, not just the returned credentials. This means there's overlap between what the OSBAPI spec prescribes and the broker-returned schema, so the question arises whether a broker that returns both credentials and e.g. syslog_drain_url must return the complete schema or just the schema for the credentials part? 

It was also intended to encompass the entire response, not just credentials. Bindings do more than just credentials and it would be a mistake to allow defining only the sup-resources for response.credentials. 

>Technically, the schema in the reference catalog response in the OSBAPI spec states that the response will not contain anything other than a credentials field (this is implied by the type:object property and the lack of either a properties.syslog_drain_url or an additionalProperties:true

This was also intentional, it is an example response that does not have any other fields.

spec.md
n3wscott: json schema allows you to add your own custom properties. We could add a field like `{... "secret":true, ...}` that the platform will have to be coded to look for and do this. I would want an implementer to try it and then feed that functionality back into the spec vs starting with it. It would not break forwards compatibility if it was added later.

spec.md
n3wscott: The rational was because the json schema may contain multiple object definitions... But yeah calling it just `Json Schema Object` is better.

spec.md
```diff
@@ -417,22 +417,30 @@ how Platforms might expose these values to their users.
...
 ##### Service Binding Object
 
 | Response Field | Type | Description |
 | --- | --- | --- |
-| [create](#input-parameters-object) | object | The schema definition for creating a Service Binding. |
+| [create](#json-schema-object) | object | The schema definition for creating a Service Binding. |
 
-##### Input Parameters Object
+
+##### JSON Schema Object
```
luksa: This needs to be called something else. In my previous comment I suggested JSON Schema**s** Object, but I now see that the 's' might be too easy to miss. 

It definitely can't be called "JSON Schema Object", because if you look in the table under this heading, you'll see the Type column refer to "JSON schema object" which makes it appear as though there's a recursion there.

A few ideas for the name (but none of them fits perfectly):
- InputOutputSchema (or InputOutputSchemas)
- RequestResponseSchema (or RequestResponseSchemas)
- JSONSchema**s**

n3wscott: This is a good catch. It does make it unclear if this is defining the form for Json Schema Object (not what I wanted) or something else. I renamed it from Input Parameters Object to not restrict it to just inout parameters...

Maybe:

* Schemas Definition Object
* Expected Schemas Object
* Defined Schemas
* Provided Schemas (or Broker Provided Schemas?)

pmorie: Looking at this, I think there are actually two different types here:

- The type for service instances, which should allow the broker to report parameter schemas for create and update
- The type for bindings, which should allow the broker to report parameter schema for create and an output schema for credentials

Seems like we ought to differentiate the types as:

- Service instance schema object
- Service binding schema object

WDYT?

n3wscott: I agree there are two different uses for the objects defined, but they are the same type of object. There is nothing special about the JSON Schema object, the special part has to do with how the platform interprets it (as the form of an expected input or expected output). Regardless of usage, the JSON Schema definition is the same.

n3wscott: How do people feel about something along the lines of using "dialect", used to mean the custom way this service will be communicating? @pmorie @luksa

n3wscott:
@luksa I decided on RequestResponseSchemaObject. 👍 ?

Also updated a little wording.

PTAL

spec.md
```diff
@@ -427,21 +427,19 @@ how Platforms might expose these values to their users.
 
 | Response Field | Type | Description |
 | --- | --- | --- |
-| [create](#json-schema-object) | object | The schema definition for creating a Service Instance. |
-| [update](#json-schema-object) | object | The schema definition for updating a Service Instance. |
+| create | [Request/ResponseSchemaObject](#request-response-json-schema-object) | The Service Broker schema definition for creating a Service Instance and the associated response. |
```
luksa: The "Request/ResponseSchemaObject" link text is slightly inconsistent with the rest of the document:
- there should be no forward slash in the name (all other links use the actual type name from the OpenAPI spec, not the "human-focused" name)
- there should be no "Object" suffix in the link text

n3wscott: I like it. Thanks for the feedback. Updated.

spec.md
```diff
@@ -427,22 +427,28 @@ how Platforms might expose these values to their users.
 
 | Response Field | Type | Description |
 | --- | --- | --- |
-| create | [InputParametersSchema](#input-parameters-schema-object) | The schema definition for creating a Service Instance. |
-| update | [InputParametersSchema](#input-parameters-schema-object) | The schema definition for updating a Service Instance. |
+| create | [RequestResponseSchema](#request-response-schema-object) | The Service Broker schema definition for creating a Service Instance and the associated response. |
```
pmorie: I don't think the type should change here - the response part doesn't seem like it has meaning for creating a service instance.

pmorie:
> It was my explicit intention to allow service_instance.update and service_binding.create to have defined responses. At the moment, current implementations do not use the response for update, but that will not be true in the future. None of this should be required, but it is my intention to allow some advanced behavior from platform/broker interactions with these hooks.

I don't think that we should add response to service instance create or update. They don't have meaning in the spec currently, there are no use cases presented for them, etc. If you feel strongly they should be part of the spec, I would suggest splitting them out of this PR and making the case for them separately. We have agreement on the other changes in this PR, as far as I can tell, so we should be able to get this merged once those changes are moved to another PR.

spec.md
n3wscott: I think there is value in allowing the JSON Schema requests for all responses, this allows a broker to define what it means (give some context text to the UI?) for it's response for any OSBAPI defined fields at the very least.

n3wscott:
>I don't think that we should add response to service instance create or update. They don't have meaning in the spec currently, there are no use cases presented for them, etc. If you feel strongly they should be part of the spec, I would suggest splitting them out of this PR and making the case for them separately. We have agreement on the other changes in this PR, as far as I can tell, so we should be able to get this merged once those changes are moved to another PR.

We have the simple use case of making responses programmatically understandable, even for the simple use-case of provision, it is interesting to be able to add broker provided context (read: descriptions and instructions) for `dashboard_url`, etc. It is no more work to make the change for all types of responses and it to not provide the opportunity for the broker to define the response for all types is a inconsistency I would like to avoid in the protocol of the broker.

Yes the main use-case is for binding response, but we see value in decorating the responses for all verbs to make usage of the brokers responses absolutely clear to the platform/user.

PR #393, opened 2017-12-05
[HOLD] Add a mechanism for broker authors to reference root schema definitions.

n3wscott:


vaikas-google:
LGTM

spec.md
```diff
@@ -400,10 +407,45 @@ The following rules apply if `parameters` is included anywhere in the catalog:
...
 schema being used.
-* Schemas MUST NOT contain any external references.
+* Schemas MUST NOT contain any external references that are resolved to a URI
+  that requires additional fetching.
 * Schemas MUST NOT be larger than 64kB.
 
 
+###### $ref Usage
+
+It is likely that a Broker will have redundent definitions of `parameters` in
```
angarg12: redundant

spec.md
```diff
@@ -400,10 +407,45 @@ The following rules apply if `parameters` is included anywhere in the catalog:
...
+the catalog. To simplfy this, the broker MAY provide a `root_schemas` entry. The
+Platform will need to process the catalog's `root_schemas` to fully qualify
+`$ref` object schmea definitions.
```
angarg12: schema

n3wscott:
@angarg12 thanks again. Any comment on the concept?

spec.md
```diff
@@ -523,8 +547,75 @@ schema being used.
         "bullets": [
           "40 concurrent connections"
         ]
+      },
+      "schemas": {
+        "service_instance": {
+          "create": {
+            "parameters": {
+              "$schema": "http://json-schema.org/draft-04/schema#",
+              "$ref": "/v2/catalog/root/#BillingAccountParameters"
```
aljim: Does root_schemas needs to be an array?
Would it be equivalent to have a map and keep using the standard [$ref syntax](https://spacetelescope.github.io/understanding-json-schema/structuring.html) 

definitions:
  BillingAccountParameters:
bindingDefinitions
  BillingAccountParameters

So that you can use

$ref: #/definitions/BillingAccountParameters
$ref: #/bindingDefinitions/BillingAccountParameters

angarg12:
@n3wscott no problem :slightly_smiling_face: looks like a good way to reduce redundancy, particularly in long catalogs.

spec.md
n3wscott: It is less complicated to have the root_schemas object as an array for aggregation/proxy brokers.

And you can't use `#/yada/yada` in the catalog schemas because those reference the local document schema for that catalog plan. It needs the full path that comes from the `$id` of the root schema, and the `$ref` has to be an external reference.

n3wscott: the root schema $id is meant to be a full uri but is not required. So I am taking advantage of the external $ref definition from the json schema spec, but putting a constraint that you need to provide that root schema in the catalog. In future versions of the OSB I would like to revisit this limitation, and allow fully external schemas. But At the F2F it was decided that fully external schemas was too much risk at the moment. So we have this half step.

To be clear, I am not inventing anything, I am just using json schemas mechanism for root schema $id links. I am adding a convention that platforms should process the root schemas array first to be able to know about links used in the plans schemas object.

angarg12:
LGTM

spec.md
```diff
@@ -400,10 +407,45 @@ The following rules apply if `parameters` is included anywhere in the catalog:
...
+* Schemas MUST NOT contain any external references that are resolved to a URI
+  that requires additional fetching.
 * Schemas MUST NOT be larger than 64kB.
 
 
+###### $ref Usage
+
+It is likely that a Broker will have redundant definitions of `parameters` in
+the catalog. To simplfy this, the broker MAY provide a `root_schemas` entry. The
+Platform will need to process the catalog's `root_schemas` to fully qualify
```
duglin: is this sentence necessary?

spec.md
```diff
@@ -400,10 +407,45 @@ The following rules apply if `parameters` is included anywhere in the catalog:
...
+`$ref` object schema definitions.
+
+
+##### Root Schema Objects
+
+| Response field | Type | Description |
+| --- | --- | --- |
+| $schema* | string | The JSON Schema declaring the version of JSON |
+schema being used. |
```
duglin: Needs to be on previous line.

spec.md
```diff
@@ -400,10 +407,45 @@ The following rules apply if `parameters` is included anywhere in the catalog:
...
+| $id* | string | The JSON Schema id to be referenced from elsewhere in the catalog. |
+| definitions* | Dictonary-of-JSON Schema Object | A definition of a JSON Schema Object. |
+
+The following rules apply if `root_schemas` is included anywhere in the catalog:
+* Platforms MUST support at least
+[JSON Schema draft v4](http://json-schema.org/).
+* Platforms SHOULD be prepared to support later versions of JSON schema.
+* The `$schema` key MUST be present in the schema declaring the version of JSON
+schema being used.
+* Schemas MUST NOT contain any external references that are resolved to a URI
```
duglin: We say this above - let's just say it once. I think having it here might be nicer, so I'd opt to remove it from above.

spec.md
n3wscott: Yes, because we insist that the catalog must contain the entire json schema, it breaks json normal json schema parsing tools. Our choices here will cause people to need to rewrite their parsing tools slightly.

spec.md
```diff
@@ -400,10 +407,45 @@ The following rules apply if `parameters` is included anywhere in the catalog:
...
+  that requires additional fetching.
+* `$id` MUST be a unique URI in the root schemas array.
+* All `definitions` MUST define a unique `$id` for the Root Schema Object in
```
duglin: I'm not sure what this mean given the previous bullet.

duglin: nvm

duglin: unique within that definition or across all?

spec.md
n3wscott: I was treating the two object types differently. I removed the size restriction from the root_schemas object.

spec.md
```diff
@@ -462,39 +504,21 @@ schema being used.
...
             "parameters": {
               "$schema": "http://json-schema.org/draft-04/schema#",
-              "type": "object",
-              "properties": {
-                "billing-account": {
-                  "description": "Billing account number used to charge use of shared fake server.",
-                  "type": "string"
-                }
-              }
+              "$ref": "http://www.openservicebrokerapi.org/v2/catalog/root/#BillingAccountParameters"
```
duglin: I think this URI is just an example, right? If so, i'd prefer to not use the OSBAPI URI, otherwise people might think its an ok thing to do, rather than use one specific to their service

n3wscott: It kind of is ok. It does no harm, we say you can't have external URI loads.

spec.md
n3wscott: just the definition. 

the first point applies to the `root_schema`, the second point applies to the `definitions` object, `root_schemas` can have `definitions` that conflict as long as their parent `$id`s are unique.

duglin: let's mention the scope in the doc.

spec.md
duglin: its not a question of harm as much as its about ownership. Technically, we own our URI, examples that suggest other people should use our URI for their own data is bad precedent.

mattmcneeney:
Thanks for putting this together @n3wscott. Thinking about this further:
* I don't think it will be trivial for client tooling to develop against this. They will have to pull schemas from two places in the catalog data and cleverly merge them together.
* It feels clunky for broker authors to have to split their schemas up in one big JSON object with somehwhat "out-of-json-schema-spec" referencing across the catalog.

We discussed the ideal solution (possibly for v3) whereby instead of embedding schemas in the catalog, service brokers could host these and serve them up on new endpoints. I'd be keen to explore this further, in a backwards compatible way (maybe having `schema_url` at the same level as `schemas`, for example).

We know that some platforms can operate in internetless/locked-down environments, so we should be aware of that and would likely have to specify that external schemas (hosted elsewhere) cannot be referenced.

spec.md
n3wscott: the OSB spec says $schema is required :(

skaegi: Right -- in the face of $ref usage patterns we'll need to re-think this a bit.
If the $ref referenced a root schema document with a $schema I think we would be fine. Unfortunately I think the more common usage pattern will be to reference a part of a schema document via pointer or $id.   
One possibility is to infer the $schema from the referenced document if expecting a root schema. Another alternative is to use $ref one-level down so we don't overwrite the root schema documents $schema.

n3wscott:
Given the feedback and our current awkward usage of json schema, how about defining a new /v2/schema endpoint?

kibbles-n-bytes:
It does seem a bit awkward to have the root schema(s) be a tag-along resource in the `v2/catalog` payload, but the benefit would be to not require a second broker call to fetch the schema.

I was brainstorming with Scott offline, and I think we might want to go with an endpoint like `/v2/catalog/schemas/` that hosts schemas for that catalog. To reference this endpoint within our existing model, we'd define a custom URI scheme that refers to this broker's host without having to embed that explicit link. So something like `osb://v2/catalog/schemas/`. Broker authors can put all sorts of wacky subresources in this base URI as long as they're routable.

Since the schema could update between reading the catalog and fetching the schemas. We'd likely want to tell broker authors that they must be static, and recommend versioning their schemas. So they'd have something like `osb://v2/catalog/schemas/my-schema-v1-0-0#`.

In a platform ecosystem where the UI and controller are separate, it'd likely be the responsibility of the controller to fetch, cache, and parse this information so that the UI could just make one request. So for us K8S folk, we'd want to make this easy on the Service Catalog side of things.

mattmcneeney:
@n3wscott @kibbles-n-bytes I'd very much welcome a proposal along those lines that you outlined. I think this would be much easier for both platform and service broker authors to implement and manage going forward (I can imagine the split-schema in the catalog approach would lead to a number of errors as custom tooling would have to be built in order to ensure that the schemas do not contain bugs).

I also agree that we should encourage service broker authors should version schemas so that platforms can do clever things like caching. We would probably want to add some guidance about how/how often service brokers can update their schemas if they move to a different endpoint.

In CF today, we cache the schema embedded in the catalog at registration time and presume this never changes (unless a service broker update is explicitly requested).

n3wscott:
Closed in favor of https://github.com/openservicebrokerapi/servicebroker/pull/402

spec.md
skaegi: Thinking about this more If $ref is using JSON Schema "delegation" everything works out just fine meaning...
```
"parameters": {
    "$ref": "http://www.openservicebrokerapi.org/v2/catalog/root/#BillingAccountParameters"
 }
```
 ... is acceptable.

PR #394, opened 2017-12-12
Adding a sample service broker for Azure support

arschles:
This patch adds the [Open Service Broker for Azure](https://github.com/Azure/open-service-broker-azure) to the list of sample brokers. This is per the suggestion at https://docs.google.com/document/d/1OqY5DoJuRKnlyF1w1XrhxlW9PjymEStU3IIG0qQLXfg/edit?disco=AAAABb4e6os


avade:
LGTM

We should refactor this page to split out samples from actual products.

mattmcneeney:
Woohoo! LGTM

angarg12:
LGTM

Would it make sense to move samples to the Wiki so we don't need PRs to introduce new links?

duglin:
LGTM

PR #398, opened 2017-12-19
add opaque Bearer Token to approved authentication methods

### LINKED ISSUE #381: Add support for opaque bearer token auth to spec
In the November 29th F2F meeting we decided to add support for opaque bearer token auth to the spec.

Some details:

- Agnostic of any particular auth scheme
- No specification for how the platform obtains the token (ie, no wording around refresh tokens, for example)

jboyd01:
fixes #381 

per https://tools.ietf.org/html/rfc6750 I explicitly called out that when using bearer tokens you must be using TLS.

### DIFF
--- a/profile.md
+++ b/profile.md
@@ -232,6 +232,12 @@ part of a Kubernetes API call:
   }
   ```
 
+## Authentication
+
+The [specification](spec.md) requires that brokers implement HTTP Basic authentication, but also allows for other mechanisms if the platform and the broker agree upon them outside of the specification. This section contains a list of supported mechanisms that platforms and brokers support. The list will be growing as time goes on.
+
+- The [Kubernetes Service Catalog](https://github.com/kubernetes-incubator/service-catalog) supports arbitrary HTTP Bearer tokens over TLS as an authentication mechanism
+
 ## Service Metadata
 
 While the [specification](spec.md) does not mandate the property names used
--- a/profile.md
+++ b/profile.md
@@ -18,7 +18,6 @@ or REQUIRED level requirements defined herein.
   - [Context Object Properties](#context-object-properties)
   - [Cloud Foundry](#cloud-foundry)
   - [Kubernetes](#kubernetes)
-- [Platform to Service Broker Authentication](#platform-to-service-broker-authentication)
 - [Service Metadata](#service-metadata)
   - [Cloud Foundry Service Metadata](#cloud-foundry-service-metadata)
 
@@ -232,12 +231,6 @@ part of a Kubernetes API call:
     "namespace": "development"
   }
   ```
-
-## Platform to Service Broker Authentication
-
-The [specification](spec.md) requires that brokers implement HTTP Basic authentication, but also allows for other mechanisms if the platform and the broker agree upon them outside of the specification. This section contains a list of supported mechanisms that platforms and brokers support. The list will be growing as time goes on.
-
-- The [Kubernetes Service Catalog](https://github.com/kubernetes-incubator/service-catalog) supports arbitrary HTTP Bearer tokens over TLS as an authentication mechanism
 
 ## Service Metadata
 
--- a/profile.md
+++ b/profile.md
@@ -18,6 +18,7 @@ or REQUIRED level requirements defined herein.
   - [Context Object Properties](#context-object-properties)
   - [Cloud Foundry](#cloud-foundry)
   - [Kubernetes](#kubernetes)
+- [Platform to Service Broker Authentication](#platform-to-service-broker-authentication)
 - [Service Metadata](#service-metadata)
   - [Cloud Foundry Service Metadata](#cloud-foundry-service-metadata)
 
--- a/profile.md
+++ b/profile.md
@@ -232,7 +232,7 @@ part of a Kubernetes API call:
   }
   ```
 
-## Authentication
+## Platform to Service Broker Authentication
 
 The [specification](spec.md) requires that brokers implement HTTP Basic authentication, but also allows for other mechanisms if the platform and the broker agree upon them outside of the specification. This section contains a list of supported mechanisms that platforms and brokers support. The list will be growing as time goes on.
 
--- a/spec.md
+++ b/spec.md
@@ -170,9 +170,12 @@ specification. Please see the
 for details on these mechanisms.
 
 If authentication is used, the Service Broker MUST authenticate the request
-using the predetermined authentication mechanism, securing communications
-via TLS, and MUST return a `401 Unauthorized` response if the authentication
-fails.
+using the predetermined authentication mechanism, and MUST return a `401 Unauthorized`
+response if the authentication fails. 
+
+Additionally, the Service Broker MUST secure communucations with TLS. The Platform
+and Service Broker SHOULD agree whether the Service Broker should use a root-signed
+certificate or a self-signed certificate.
 
 Note: Using an authentication mechanism that is agreed to via out of band
 communications could lead to interoperability issues with other Platforms.
--- a/spec.md
+++ b/spec.md
@@ -157,11 +157,14 @@ While the communication between a Platform and Service Broker MAY be unsecure,
 it is RECOMMENDED that all communications between a Platform and a Service
 Broker are secured via TLS and authenticated.
 
-Unless there is some out of band communication and agreement between a
-Platform and a Service Broker, the Platform MUST authenticate with the
-Service Broker using HTTP basic authentication (the `Authorization:` header)
-on every request. This specification does not specify how Platform and Service
-Brokers agree on other methods of authentication.
+Unless there is some out of band communication and agreement between a Platform
+and a Service Broker, the Platform MUST authenticate with the Service Broker
+using HTTP basic authentication (the `Authorization:` header) or a bearer token
+(the `Authorization: Bearer` header) on every request. If using bearer tokens,
+TLS MUST be used used.  This specification does not specify how Platform and
+Service Brokers agree on other methods of authentication nor does it specify how
+the bearer token is issued or the how token lifecycle including token timeout
+and token renewal is handled.
 
 If authentication is used, the Service Broker MUST authenticate the request
 using the predetermined authentication mechanism and MUST return a `401
--- a/spec.md
+++ b/spec.md
@@ -171,10 +171,10 @@ for details on these mechanisms.
 
 If authentication is used, the Service Broker MUST authenticate the request
 using the predetermined authentication mechanism, and MUST return a `401 Unauthorized`
-response if the authentication fails. 
+response if the authentication fails.
 
 Additionally, the Service Broker MUST secure communucations with TLS. The Platform
-and Service Broker SHOULD agree whether the Service Broker should use a root-signed
+and Service Broker SHOULD agree whether the Service Broker will use a root-signed
 certificate or a self-signed certificate.
 
 Note: Using an authentication mechanism that is agreed to via out of band
--- a/spec.md
+++ b/spec.md
@@ -173,7 +173,7 @@ If authentication is used, the Service Broker MUST authenticate the request
 using the predetermined authentication mechanism, and MUST return a `401 Unauthorized`
 response if the authentication fails.
 
-Additionally, the Service Broker MUST secure communucations with TLS. The Platform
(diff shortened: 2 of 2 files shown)


duglin:
I'm not sure we can do it this way w/o breaking backwards compat.  Rather I think we can get the same net results by adding a subsection that basically talks about how "if" bearer tokens are used then this is how you do it.

arschles:
+1 to @duglin. It would be nice to keep the `MUST` language under the "If bearer tokens are used" subsection though.

jboyd01:
@duglin , @arschles  I've attempted a rework on this, happy to go a different direction if you think there is something better.

mattmcneeney:
This looks good to me too.

As a side note @jboyd01, do you have any documentation outlining how you see this working generally?

spec.md
```diff
@@ -165,6 +165,13 @@ Service Broker using HTTP basic authentication (the `Authorization:` header)
 on every request. This specification does not specify how Platform and Service
 Brokers agree on other methods of authentication.
 
+An alternative to carrying basic authentication on every request is to utilize a
+bearer token via the `Authorization: Bearer` header.  When a bearer token is
+used additional processes are often required to deal with token expiration and
+renewal.  These are details not covered by this specification and must be worked
```
mattmcneeney: Travis will complain about lowercase must

spec.md
```diff
@@ -165,6 +165,13 @@ Service Broker using HTTP basic authentication (the `Authorization:` header)
 on every request. This specification does not specify how Platform and Service
 Brokers agree on other methods of authentication.
 
+An alternative to carrying basic authentication on every request is to utilize a
+bearer token via the `Authorization: Bearer` header.  When a bearer token is
+used additional processes are often required to deal with token expiration and
```
mattmcneeney: Travis will complain about lowercase required

jboyd01:
@mattmcneeney, thanks for the review.  regarding your question above about "any documentation outlining how you see this working generally" - I don't think we have any doc, but as a concrete example:

Service Catalog uses opaque tokens on Open Shift for authenticating with our service brokers such as the Ansible Service Broker.  When a new Broker is created, the broker authentication information specifies to use a specific secret whose content contains the necessary bearer token.  The bearer token secret is created external to Catalog by the Broker installation and the token does not expire.

If you desire token expiration, you would need a mechanism to update the K8s secret.  The updated secret would automatically be picked up by Service Catalog the next time it is referenced.

duglin:
The text itself looks good. Just one higher-level question... why are we calling out this one specific security mechanism? Will this open the door to people wanted to add guidance for others as well? For now I guess its ok, but I'm not keen on duplicating what should really be part of the "out of bands agreement(spec)" that we say the platform and broker decide on.  Its kind of like we're saying "its none of our business, but we're gonna tell you want to do anyway"  :-)

spec.md
```diff
@@ -165,6 +165,13 @@ Service Broker using HTTP basic authentication (the `Authorization:` header)
 on every request. This specification does not specify how Platform and Service
 Brokers agree on other methods of authentication.
 
+An alternative to carrying basic authentication on every request is to utilize a
+bearer token via the `Authorization: Bearer` header.  When a bearer token is
+used additional processes are often needed to deal with token expiration and
+renewal.  These are details not covered by this specification and need to be
+worked out by the Platform.  If bearer tokens are used communication with the
+Broker MUST be secured over TLS.
```
n3wscott: `Broker` -> `Service Broker`

duglin:
Per our last call, I believe Jay, Aaron and Alex are going to talk off-line about rewording this. I think to mainly focus on how using TLS is good when any auth mechanism is used.

jboyd01:
Alex, Aaron and I are doing some additional rework on this removing the specifics  about opaque token and clarifying some aspects around authentication.  Marked as Work In Progress for now, not ready for review.

mattmcneeney:
Hey @jboyd01 You may want to rebase this PR on top of master as I believe it's showing many changes that aren't part of this change!

jboyd01:
@mattmcneeney  Yes, it was a rebase gone bad, got it straightened out now.  Thanks

spec.md
```diff
@@ -163,9 +163,15 @@ Service Broker using HTTP basic authentication (the `Authorization:` header)
 on every request. This specification does not specify how Platform and Service
 Brokers agree on other methods of authentication.
 
+Platforms and brokers may agree on an authentication mechanism other than
```
duglin: s/brokers/Service Brokers/
s/may/MAY/

spec.md
```diff
@@ -163,9 +163,15 @@ Service Broker using HTTP basic authentication (the `Authorization:` header)
...
+basic authentication, but the specific agreements are not covered by this
+specification. Please see the platform.md documentation for a description
```
duglin: s/platform.md/[profile](profile.md)/

spec.md
```diff
@@ -163,9 +163,15 @@ Service Broker using HTTP basic authentication (the `Authorization:` header)
...
+of each mechanism and platform support.
+
 If authentication is used, the Service Broker MUST authenticate the request
-using the predetermined authentication mechanism and MUST return a `401
-Unauthorized` response if the authentication fails.
+using the predetermined authentication mechanism, securing communications
+via TLS, and MUST return a 401 Unauthorized response if the authentication
```
duglin: add tick marks around `401 Unauthorized`

duglin:
sgtm just minor typos

jboyd01:
ready for review, simplified approach indicating Basic Auth is defacto auth mechanism and referring to the profile.md for specifics on Kube & CF.   Hopefully we'll update Service Catalog soon with more doc around platform to broker authentication as well.

profile.md
```diff
@@ -232,6 +233,12 @@ part of a Kubernetes API call:
   }
   ```
 
+## Platform to Service Broker Authentication
+
+The [specification](spec.md) requires that brokers implement HTTP Basic authentication, but also allows for other mechanisms if the platform and the broker agree upon them outside of the specification. This section contains a list of supported mechanisms that platforms and brokers support. The list will be growing as time goes on.
```
duglin: can you wrap these at 80 cols to match the rest of the spec?

profile.md
```diff
@@ -232,6 +233,12 @@ part of a Kubernetes API call:
...
+
+- The [Kubernetes Service Catalog](https://github.com/kubernetes-incubator/service-catalog) supports arbitrary HTTP Bearer tokens over TLS as an authentication mechanism
```
duglin: I'm not sure we want to add this to the spec - it feels kind of awkward since each time an impl changes what it  supports we'd need to publish a new version of the spec. Putting it in our wiki might be ok though.

duglin: If we wanted to add something to the Profile about auth, then I think having auth-specific rules or guidance would be appropriate - stuff that all impls should adhere too.

profile.md
mattmcneeney: s/brokers/Service Brokers

profile.md
mattmcneeney: As a service author, where can I go to learn how to build a broker against this? The service-catalog readme doesn't direct me anywhere to learn this

duglin: This is why the info here shouldn't really be platform specific, rather, if anything, it would be auth specific. E.g. I don't think svc-cat's use of HTTP bearer tokens is unique, or at least I hope not, otherwise we wouldn't have interop. 

I kind of wonder what the purpose of this text is. I don't think we should just reiterate what each platform's docs say w.r.t. which auth mechanisms it supports. That's not only duplicate info but requires us to keep things in sync. I'd prefer for this section to be more about any OSB API specific aspects of the various auth mechanisms Platforms support that people will need to know about if they want to write code - that way the odds of having interop go up.

spec.md
```diff
@@ -163,9 +163,15 @@ Service Broker using HTTP basic authentication (the `Authorization:` header)
...
+Platforms and Service Brokers MAY agree on an authentication mechanism other
+than basic authentication, but the specific agreements are not covered by this
+specification. Please see the [profile](profile.md) documentation for a
+description of each mechanism and platform support.
+
 If authentication is used, the Service Broker MUST authenticate the request
-using the predetermined authentication mechanism and MUST return a `401
-Unauthorized` response if the authentication fails.
+using the predetermined authentication mechanism, securing communications
+via TLS, and MUST return a `401 Unauthorized` response if the authentication
```
fmui: A dependency on TLS makes service broker development more difficult. For production TLS is a must, for development not necessarily. If we insist on TLS, do we allow self-signed certificates?

profile.md
jboyd01: re [Matt's comment](https://github.com/openservicebrokerapi/servicebroker/pull/398#discussion_r164432525)
>As a service author, where can I go to learn how to build a broker against this? The service-catalog readme doesn't direct me anywhere to learn this

We've got a [general documentation issue](https://github.com/kubernetes-incubator/service-catalog/issues/1678) we hope to address for the next major release which will help, but you are right, presently there isn't much doc on this

duglin:
rebase needed

profile.md
duglin: @jboyd01 ping

profile.md
arschles: +1 to ^^ - we don't have much to put here yet, but it would be nice to set the precedent. if we don't want to put specific systems in here, I would be ok with something like a "coming soon" in this section

spec.md
arschles: I'd be ok with self-signed certs for non-production deployments, or honestly even not requiring TLS in those cases

profile.md
arschles: Per offline conversation with @duglin, @jboyd01 and myself, we will:

- move this section into the spec.md, replacing it with the link to profile.md
- link to a wiki page which @duglin will create
- the wiki page will be non-normative and contain at least a non-exhaustive list of known platforms (and possibly brokers in the future?) that implement other auth. mechanisms

spec.md
arschles: Per offline conversation with @duglin, @jboyd01 and myself, we will change this language to require TLS only in production, and say that it's a decision between the broker and the platform whether they use root or self signed certs

duglin:
there are a ton of other changes in here, did you mean to make those?

arschles:
sorry @duglin @jboyd01 - my markdown linter went nuts in https://github.com/openservicebrokerapi/servicebroker/pull/398/commits/f50a287253b971a159e2b2d6ac26b11ae3909ee8. I'm gonna back that out and this should be ready to go

arschles:
@duglin @jboyd01 I have removed the lame commit I made that linted all the markdown. This should be good to go now

spec.md
```diff
@@ -163,9 +163,19 @@ Service Broker using HTTP basic authentication (the `Authorization:` header)
 on every request. This specification does not specify how Platform and Service
 Brokers agree on other methods of authentication.
 
+Platforms and Service Brokers MAY agree on an authentication mechanism other
+than basic authentication, but the specific agreements are not covered by this
+specification. Please see the 
+[Platform Features authentication mechanisms wiki document](https://github.com/openservicebrokerapi/servicebroker/wiki/Platform-Features)
```
pmorie: I don't understand why this isn't in the spec, but I'm happy with this for now.

pmorie:
LGTM

jboyd01:
@duglin this should be all set, PTAL.

duglin:
LGTM!

fmui:
LGTM

spec.md
```diff
@@ -163,9 +163,19 @@ Service Broker using HTTP basic authentication (the `Authorization:` header)
...
+for details on these mechanisms.
+
 If authentication is used, the Service Broker MUST authenticate the request
-using the predetermined authentication mechanism and MUST return a `401
-Unauthorized` response if the authentication fails.
+using the predetermined authentication mechanism, and MUST return a `401 Unauthorized`
+response if the authentication fails.
+
+Additionally, the Service Broker MUST secure communucations with TLS. The Platform
```
mattmcneeney: communucations -> communications

duglin:
LGTM
3 more needed

fmui:
LGTM

mattmcneeney:
LGTM

n3wscott:
LGTM

PR #399, opened 2017-12-19
Fix it so we don't stop after the first error/file

duglin:
Also made the verbose setting more consistent across the tools

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1470,7 +1470,7 @@ Platforms SHOULD initiate orphan mitigation in the following scenarios:
 | 201 with malformed response | Failure | Yes |
 | All other 2xx | Failure | Yes |
 | 408 | Timeout failure | Yes |
-| All other 4xx | Service Broker rejected request | No |
+| All other 4xx | Request rejected | No |
 | 5xx | Service Broker error | Yes |
 | Timeout | Failure | Yes |
 
--- a/tools/verify-phrases.sh
+++ b/tools/verify-phrases.sh
@@ -55,6 +55,13 @@ verbose=""
 debug=""
 stop=""
 
+# Error file processing
+err=tmpCC-$RANDOM
+trap clean EXIT
+function clean {
+  rm -f ${err}*
+}
+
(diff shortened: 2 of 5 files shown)


spec.md
```diff
@@ -1470,7 +1470,7 @@ Platforms SHOULD initiate orphan mitigation in the following scenarios:
 | 201 with malformed response | Failure | Yes |
 | All other 2xx | Failure | Yes |
 | 408 | Timeout failure | Yes |
-| All other 4xx | Service Broker rejected request | No |
+| All other 4xx | Request rejected | No |
```
duglin: I made this change just so the checker doesn't need to look for "5 word" columns to ignore, that's getting kind of long. Plus saying "Service Broker" isn't really necessary, its clear who is doing the rejection.

avade:
LGTM

tools/verify-phrases.sh
```diff
@@ -164,18 +168,11 @@ function checkFile {
     for phrase in "${bannedPhrases[@]}"; do
       if [[ "${words[@]^^} " == "${phrase^^} "* ]]; then
       ll=${words[*]}
-      echo line ${lines[0]}: \'${ll:0:${#phrase}}\' is banned | tee -a $err
+      echo line ${lines[0]}: \'${ll:0:${#phrase}}\' is banne
```
mattmcneeney: missing a `d`

duglin: fixed!

duglin:
Actually, I've purposely tried to keep the scripts separate in order to keep them maintainable. One large script that tries to do it all might make it harder for people to understand what's going on - or just scare people. Most of the scripts have the same boiler-plate logic for flag parsing, etc... and so its really just the "checkFile" type of function people need to focus on.

angarg12:
LGTM

mattmcneeney:
LGTM (@avade need yours again btw)

duglin:
2 more reviews needed

fmui:
LGTM

avade:
LGTM

PR #402, opened 2017-12-20
Add JSON Schema Endpoint.

n3wscott:
The [previous discussion](https://docs.google.com/document/d/1-IKI-PwPnhtbK0su1UzWA_UoyaCjL8S2d6ky3Xmuwvg/edit#heading=h.ihmogawcexny) went over a lot of alternatives and there was much debate. What we ended up with was "method attached schemas". The current implementation and integration of JSON Schema per plan is proving to be less than ideal for large catalogs. For simple brokers the current usage works just fine, but when a broker's catalog grows to more than a few services, each with a couple of plans, and all use a similar object as parameters, the catalog grows with duplicated data. There is no allowance in the specification to allow for external references and the implementation prevents internal references to be shared between methods or plans. This PR implements the ideas outlined in the [new design document](https://docs.google.com/document/d/1EURbGG2UzELmKyZrbNnn4kbEHkwNyJSmLyQDH7kcQjo/edit#) on how to solve reusable json schema definitions.

As a pivot on the idea formed in https://github.com/openservicebrokerapi/servicebroker/pull/393


mattmcneeney:
@n3wscott I think this is looking good, thanks for putting it together. Question: if a broker author simply wants to use the same reference for every schema they offer for create/update/bind, then is there a shorter way for them to do this than the below for each one:
```
"service_binding": {
  "create": {
    "parameters": {
      "$ref": "osb.v2:///fakeservice/v1/bindings#AccountParameters"
    }
  }
}
```

n3wscott:
>  a shorter way

Hm like a default? It seems very tricky to get that right but this change does not preclude that idea down the road.

Perhaps an enhancement that could be interesting later would be to allow the schemas parameter for the service rather than the plan? Schemas on service + a defaults might give the smallest amount of catalog metadata, but it dramatically increases the amount of logic needed to understand the catalog.

duglin:
@n3wscott can you modify the top comment with the history behind this PR (what problems its trying to solve, etc...) so that when I share with other folks they can understand why this is needed?

n3wscott:
> (what problems its trying to solve, etc...)

@duglin I gave it a shot, does that help?

duglin:
Thanks @n3wscott that'll help.

duglin:
Been having some talks internally about this and this approach raises some concerns. This makes it harder for UIs/CLIs to get their job done. When the schema is in-lined then the results of `GET /v2/catalog` can be passed on to CLIs directly. For example, CF's `GET /v2/services` returns the list of services and in there will be all of the schema needed for the CLI to produce a pretty UX. If we pull out shared/common schema into a separate endpoint then its not quite as nice.

We would either need:
1) to allow CLIs to talk directly to the schema endpoint (which may not be possible - e.g. due to it not having the knowledge, access, or creds to talk directly to the broker)
2) the platform would need to act as a proxy of some kind to support retrieving just the $ref's available to the one CLI doing a request
3) perform some kind of merging of the schema (and its refs) into the `GET /v2/services` response stream so that we avoid the need for either of the first two options

@jwforres - I'd be interested in knowing if you guys have run into this concern with your UI work

n3wscott:
Thanks for the feedback Doug! 

My general comment is the OSB tends to take the stand for simple brokers and complex controllers. JSON Schema Endpoints does make the broker slightly more complicated and makes the controller much more complicated for scheme processing. But the goal here is to solve the redundant data issue that current catalogs are suffering from. We are finding our catalog payloads are growing by about 1mb per 4 services with a reasonable amount of plans provided, so that is where our motivation for wanting this kind of solution comes from. The current usage of schema definitions in the catalog is unusable in the long term.

Let me respond to your points:

>1. to allow CLIs to talk directly to the schema endpoint (which may not be possible - e.g. due to it not having the knowledge, access, or creds to talk directly to the broker)

I don't think the CLI can ever assume it is allowed to talk to the broker directly. It will be the job of the controller or platform to help the CLI get the required schemas. 

>2. the platform would need to act as a proxy of some kind to support retrieving just the $ref's available to the one CLI doing a request

I would suggest the platform acting as a cache vs a proxy. Just as the platform should act as a cache for the catalog.

> 3. perform some kind of merging of the schema (and its refs) into the GET /v2/services response stream so that we avoid the need for either of the first two options

That could be a choice for your platform. You can replace a $ref with the targeted document, but this approach has some serious downsides for complicated JSON Schema in regards to dealing with reference cycles and loops. Hence the advice for the link style processing. I bet it would work for 
99% of the broker cases though. The cache of the catalog in the platform could be written do undo the $ref usage as a processing step when it is importing the catalog. This will remove any catalog payload savings.

Also, using a JSON Schema Endpoint is optional. If you don't plan to import 3rd party brokers, and you don't intend to implement a the Schemas endpoint, your CLI does not need to care.

jwforres:
We also have concerns about the current size (from the client -> platform perspective) when fetching the list of available services. It can and will become a significantly large request, which creates a crappy experience waiting for it to be loaded. We don't actually need or want the schema until the point where you choose a particular service to provision, or an instance to update.

As mentioned above, the clients (browser / CLI) can not be expected to talk directly to the broker.  So yes the burden is on the platform to make the information available in some way via its APIs.

Speaking purely from the client -> platform API level, I dont want listing a service catalog to return the schemas. I would be ok with getting a specific service class (or plan if they are split out) including the schemas. Alternatively having the platform have a separate API for getting a specific schema is not terrible either.

cc @spadgett since he's most familiar with our editors and how we use the schemas

mattmcneeney:
@jwforres Do you have any idea how easy it would be for platforms to completed unroll all `$ref` links in schemas (potentially many links nested in each other) to create one top-level schema? For the reasons @duglin mentioned, this is what we'd want to do in CF and then expose APIs to collect create/update/bind schemas for a specific service ID and plan ID.

duglin:
I want to be clear that the concerns raised aren't about whether this PR can technically work or not, its about whether the use of $ref's is worth the added complexity/risks it introduces.  In other words, is asking each platform to manage the $ref's (unroll, dereference, merge, duplicate, whatever... the data) rather than just blindly return whatever json is provided by the broker (without any additional processing that adds the potential for errors) worth it. Yes the data returned by the broker could be large, but if they're only asking for it (say) once a week (or less) then we need to weigh the pros/cons.  But we're still thinking too...

duglin:
@n3wscott and I had lots of good chats where he explained some of the really interesting usecases they want to support as a result of this PR. Some of the more interesting ones revolved around the Platform being able to cache some of the results from the Broker - mainly due to the size of the data (schema) involved, and the time required to process that data. They have a system where they re-list their Broker's catalogs often, very often, like every minute! There's nothing in the spec to prevent this, or even recommend against it, so that's ok.

For this reason they want to have the schema separate from the normal service metadata - then they can do things like only GET new/unknown $ref's w/o the rest of the schema - on every re-list.

However, the spec doesn't say anything about the "static-ness" of the data. For example, while it doesn't come right out and say it, I think there is a strong implication that the Platform will do a diff of the `/v2/catalog` results against the previous results to see what, if anything, has changed. The spec does make it clear that the only static thing across re-lists will be the ID of the services/plans - everything else is mutable:
```
When determining what, if anything, has changed on a Service Broker, the Platform will use
the id of the resources (services or plans) as the only immutable property and MUST use
that to locate the same resource as was returned from a previous query.
```

If we head down the path this PR suggests, then I think more text/requirements will be needed in order to address these usecases. For example, the spec would need to mandate that schema never changes - which is different from how we deal with service metadata today (where its mutable and the Platform needs to diff it each time).  I think we may also need to be clear about the lifetime of the schema at this new endpoint. For example, is there any requirement for the Broker to have schema for any old service/plan? I'm assuming not, but then we need to make it clear to the Platform that they really can't lazy-load the schema because its possible that between the time they GET `/v2/catalog` and the time they GET the schema the schema may have changed due to the catalog changing the list of its services/plans. This would then obviously put a requirement on the Broker to ensure consistency between the two endpoints - and we may need to be explicit about the lifetime of that consistency and the implications for (ever popular :-) ) stateless brokers to support it.

spec.md
```diff
@@ -523,13 +518,295 @@ schema being used.
...
+The root JSON Schema document can be found by getting the JSON Schemas
+endpoint without a `:component_path`.
+
+If the `osb_v2:` scheme is used by the broker, then the Platform MAY fetch the
+component JSON Schema directly using the following convention:
+
+Given: `osb_v2:///{:component_path}`, fetch `/v2/catalog/schemas/:component_path`
+and then the Platform is responsible to continue fetching any unknown
+`:component_path` URIs. Care ought to be taken by the platform to not fetch
+duplicate definitions.
```
duglin: Not sure what this means by fetching "duplicate definitions".

spec.md
```diff
@@ -523,13 +518,295 @@ schema being used.
...
+Platform but increases the potential payload size of the JSON Schemas.
+
+A Platform MUST fetch the root Schema document for any other `$ref` URIs found
+that do not match the `osb_v2:` scheme. If this happens the Platform has no use
+for the previously fetched sub-schemas and can just use the root schema. The
+Broker MUST provide the JSON Schema Object for the referenced URI in a subschema
+returned.
+
+Component Paths that use the scheme `osb_v2:` SHOULD be versioned, and the
+document behind the URI MUST NOT change after the catalog has been served to a
```
duglin: I missed this MUST NOT before.  Does this include never deleting it? I hope not.

n3wscott: Delete is fine unless it is still referenced by the catalog.

spec.md
n3wscott: Fetching the same `:component_path`. So then you end up with a list containing two copies of a schema with the same id.

spec.md
n3wscott: I think this line addresses your concern with:

>However, the spec doesn't say anything about the "static-ness" of the data. For example, while it doesn't come right out and say it, I think there is a strong implication that the Platform will do a diff of the /v2/catalog results against the previous results to see what, if anything, has changed.

Correct?

spec.md
```diff
@@ -523,13 +518,295 @@ schema being used.
...
 }
 ```
 
 
+## JSON Schemas
+
+The JSON Schemas endpoint is used to define reusable definitions of JSON Schema
+Objects referenced throughout the [Catalog](#catalog-management). It will be the
+responsibility of the Platform to examine the response from the catalog to
+determine if the Broker is using `$ref` JSON Schemas. The Broker can signal to
```
duglin: s/Broker/Service Broker/ 
twice

spec.md
```diff
@@ -523,13 +518,295 @@ schema being used.
...
   }]
 }
 ```
 
 
+## JSON Schemas
+
+The JSON Schemas endpoint is used to define reusable definitions of JSON Schema
+Objects referenced throughout the [Catalog](#catalog-management). It will be the
+responsibility of the Platform to examine the response from the catalog to
```
duglin: I believe the Service Broker will be control over whether this feature is used or not. Meaning, Platforms must look for $ref, support osbapi:// URLs AND support retrieving things from a separate JSON Schema Endpoint if their Service Brokers choose to use these features and the Platform does schema checking. This basically makes it a requirement on Platforms that want to deal with schema at all.

I think this means that we should probably add some MUSTs about what the Platform needs to support *if* it wants to do schema validation. Although, I guess it could reject the request if it doesn't support these things - but then it must be at least aware enough to even look for things like osbapi:// to reject the request. Anyway, I think we need to say something about this to ensure some kind of interop and validation during catalog processing, because I think it will cause issues if the Broker returns something that the Platform doesn't know about (or support) and no one detects this mismatch.

spec.md
```diff
@@ -523,13 +518,295 @@ schema being used.
...
+determine if the Broker is using `$ref` JSON Schemas. The Broker can signal to
+the Platform that JSON Schema components are supported by using the `osb_v2:`
+scheme. Components are a way to allow Platforms to only fetch sub-schemas.
+
+The root JSON Schema document can be found by getting the JSON Schemas
+endpoint without a `:component_path`.
+
+If the `osb_v2:` scheme is used by the broker, then the Platform MAY fetch the
```
duglin: s/broker/Service Broker/

spec.md
duglin: I'm still not following. Are you saying that if `osb_v2:///foo` appears twice in the schema, then the Platform only needs to retrieve it once? If so, yes I agree, but the "Care ought to be taken..." sentence makes it sound like something bad will happen if they do retrieve it more than once.  I would suggest that rather than making it sound like there's something people need to be worried about, we phrase it more like a good thing, an optimization. For example, talk about how this feature gives Platforms a more optimized schema processing and retrieval behavior by allowing portions of the schema that are used in multiple places to only be retrieved once. Or something like that.

spec.md
```diff
@@ -523,13 +518,295 @@ schema being used.
...
+component JSON Schema directly using the following convention:
+
+Given: `osb_v2:///{:component_path}`, fetch `/v2/catalog/schemas/:component_path`
+and then the Platform is responsible to continue fetching any unknown
+`:component_path` URIs. Care ought to be taken by the platform to not fetch
+duplicate definitions.
+
+A Platform MAY ignore the `osb_v2:` scheme and just fetch `/v2/catalog/schemas`
```
duglin: Does this mean that a Broker MUST return ALL of the schema via a single `GET /v2/catalog/schemas` even if it also supports retrieving a single component schema via `GET /v2/catalog/schemas/foo`? If so, can we get a MUST in here stating that?

spec.md
```diff
@@ -523,13 +518,295 @@ schema being used.
...
+which will contain all sub-Schema documents. This reduces the complexity of the
+Platform but increases the potential payload size of the JSON Schemas.
+
+A Platform MUST fetch the root Schema document for any other `$ref` URIs found
```
duglin: I don't think we can use MUST here in this way since we can't mandate they retrieve any of this. I think the last sentence in this paragraph where we mandate that the broker MUST have the info available is probably the best we can do.

spec.md
```diff
@@ -523,13 +518,295 @@ schema being used.
...
+that do not match the `osb_v2:` scheme. If this happens the Platform has no use
+for the previously fetched sub-schemas and can just use the root schema. The
```
duglin: Which "previously fetched sub-schemas" does this refer to?

spec.md
duglin: Saying component paths SHOULD be versioned feels awkward to me. It seems like it should be a MUST instead if the data can never be changed. But, do we really need to talk about versioning at all? That feels more like an impl detail/choice. What if we just said:
When the `osb_v2:` scheme is used, the data referenced by a component path MUST NOT change and can only be deleted when there are no longer any references to it in the Service Broker's schema. Additionally, once deleted, that component path MUST NOT ever be used again by this Service Broker.

spec.md
```diff
@@ -523,13 +518,295 @@ schema being used.
...
+Broker MUST provide the JSON Schema Object for the referenced URI in a subschema
+returned.
+
+Component Paths that use the scheme `osb_v2:` SHOULD be versioned, and the
+document behind the URI MUST NOT change after the catalog has been served to a
+Platform. 
+
+Platforms are able to leverage this JSON Schema by using or creating a library
```
duglin: I'd remove this - it sounds like an impl detail to me.

spec.md
```diff
@@ -523,13 +518,295 @@ schema being used.
...
+that supports pre-fetched JSON Schemas.
+
+Brokers MUST NOT allow duplicate JSON Pointers to be formed in any JSON Schema.
```
duglin: Can you elaborate on this? I'm not sure what a "duplicate JSON pointer" is.

spec.md
```diff
@@ -523,13 +518,295 @@ schema being used.
...
+libraries that fully support the [JSON Schema](http://json-schema.org/) spec in
+the usage of URI for `$ref`. Only URL was supported in libraries. So as a
+compromise the `osb_v2:` scheme SHOULD be followed by three forward slashes
+(`///`). This indicates the authority is implied (authority is ''). 
+
+
+#### $ref Usage
+
+Brokers MAY leverage `$ref` and external JSON Schemas only if the JSON Schema
+references can be found within the [JSON Schemas](#json-schemas) API endpoint. $ref
```
duglin: I think the $ref should be in back-ticks

spec.md
```diff
@@ -523,13 +518,295 @@ schema being used.
...
+usage is assumed to be delegation (linking), not inclusion (replacement).
+Libraries that implement `$ref` content replacement result in adverse behaviors.
+
+For Brokers that support `$ref`:
```
duglin: add ", they" before the colon

spec.md
```diff
@@ -523,13 +518,295 @@ schema being used.
...
+| Status Code | Description |
+| --- | --- |
+| 200 OK | MUST be returned upon successful processing of this request. The expected response body is below. |
+
+
+#### Body
+
+The root JSON Schema document is a complete collection of JSON Schemas
+leveraged by the Broker's catalog. The Platform will process this
+document before catalog `$ref` URIs map to definitions. The body MUST contain
```
duglin: I don't know what: "The Platform will process this document before catalog `$ref` URIs map to definitions."
means. Can you elaborate?

spec.md
```diff
@@ -523,13 +518,295 @@ schema being used.
...
+all JSON Schema object definitions found in the catalog and referenced
+sub-schemas.
+
+| Response field | Type | Description |
```
duglin: s/field/Field/

spec.md
```diff
@@ -523,13 +518,295 @@ schema being used.
...
+| --- | --- | --- |
+| $schema* | string | The JSON Schema declaring the version of JSON schema being used. |
+| $id* | string | The top level JSON Schema ID for this document. |
+| definitions* | Dictionary-of-JSON Schema Object | A collection of JSON sub-Schema Objects. |
+
+Please refer to the [JSON Schema](http://json-schema.org/) specification for
+more details on JSON Schema.
+
+Note the same restriction on `$ref` applies to `/schemas`.
```
duglin: Which restrictions? Perhaps mention `/schema` where ever those restrictions are, or repeat them here.

spec.md
```diff
@@ -523,13 +518,295 @@ schema being used.
...
+the Platform that JSON Schema components are supported by using the `osb_v2:`
+scheme. Components are a way to allow Platforms to only fetch sub-schemas.
+
+The root JSON Schema document can be found by getting the JSON Schemas
+endpoint without a `:component_path`.
+
+If the `osb_v2:` scheme is used by the broker, then the Platform MAY fetch the
+component JSON Schema directly using the following convention:
+
+Given: `osb_v2:///{:component_path}`, fetch `/v2/catalog/schemas/:component_path`
```
duglin: just wondering... is there any way to achieve the same net results w/o introducing a new URL scheme? E.g. "foo" means its in the full schema doc but "/foo" means its available at /v2/catalog/schemas/foo".  Might not be a big enough distinction, but I'd like to brainstorm a bit to see if there are other options rather than introducing something that feels like a bit of a hack.

spec.md
```diff
@@ -523,13 +518,295 @@ schema being used.
...
 
 
+## JSON Schemas
+
+The JSON Schemas endpoint is used to define reusable definitions of JSON Schema
+Objects referenced throughout the [Catalog](#catalog-management). It will be the
+responsibility of the Platform to examine the response from the catalog to
+determine if the Broker is using `$ref` JSON Schemas. The Broker can signal to
+the Platform that JSON Schema components are supported by using the `osb_v2:`
+scheme. Components are a way to allow Platforms to only fetch sub-schemas.
```
duglin: We need some text around the lifecycle of this data. For example, let's say we have the following flow:
- platform get the catalog, which includes the schema and it has osb_v2:/// refs in there
- hours go by and the platform doesn't retrieve any sub-component schemas
- broker updates its catalog and removes the references sub-component schema - perhaps because the plan was removed
- platform then tries to retrieve the sub-component schema because someone tries to use that service/plan. They'll get a 404

While the user will eventually get some kind of error when they try to talk to the broker, but in the meantime the platform has the wrong info and is presenting it to the user. I don't think the platform can assume that any 404 for a schema means it should re-get the catalog, plus it may not want to update all its services/plans at that time.

So, I'm not sure what we can say about this yet so I'd like to brainstorm a bit, but I think we need to say something about the requirements we're placing on the broker w.r.t. how these endpoints need to be in-sync

spec.md
n3wscott: Maybe you can help me wordsmith this, the care comes into play if the platform chooses to implement the external schemas as a dumb loop that downloads and appends the schema to a list to be given to the pre-fetched schema list. If that document collection has objects that collide in id, the processing library is unable to process any ids. That is the warning I am trying to put here. 

Maybe ... "Platforms MUST read the JSON Schema spec." ?

spec.md
n3wscott: Yes, it is already there in the bullets for service brokers in the `$ref Usage` section.

spec.md
n3wscott: Change it to SHOULD?

spec.md
n3wscott: In the case where it is processing a catalog and the service broker has done a bad thing and used a mix of pure external ids and osb_v2: external ids.

spec.md
n3wscott: I like it. Replaced.

spec.md
n3wscott: Gone.

spec.md
n3wscott: I don't remember. I think it was the broker should not make duplicate ids that collide but I think it is cleaner to just not state this, the broker would never work if they do this anyway.

spec.md
n3wscott: done.

spec.md
n3wscott: It means this only works with pre-fetched json schema support which is a more advanced Json Schema client feature.

spec.md
n3wscott: Done.

spec.md
n3wscott: I was trying to say go read the `$ref Usage` section, but it looks like that section applies here too so I am goign to just leave this line out of the spec.

n3wscott:
Thanks for the review @duglin. I have updated the PR and left a couple questions for you.

spec.md
```diff
@@ -560,13 +549,294 @@ schema being used.
...
   }]
 }
 ```
 
 
+## JSON Schemas
+
+The JSON Schemas endpoint is used to define reusable definitions of JSON Schema
+Objects referenced throughout the [Catalog](#catalog-management). If a Platform
+intends to support JSON Schemas, it MUST examine the response from the catalog
```
duglin: does it have a choice in the matter?

duglin: By that I mean, the Platform could choose to support schema but it can't choose to not support the schema endpoint if it does support schema - correct?

n3wscott: That is correct. If the platform supports JSON Schema with `$ref`, then it must support the schemas endpoint.

spec.md
n3wscott: I have updated the language here to try to make it more clear.

spec.md
n3wscott: I realized that the platform does not have to process this. They could just pass the response along to the client UI and have that process it. So I reworked this to loosen it up.

spec.md
```diff
@@ -570,13 +559,297 @@ schema being used.
...
+to determine if the Service Broker is using `$ref` JSON Schemas. The Service
+Broker can signal to the Platform that JSON Schema components are supported by
+using the `osb_v2:` scheme. Components are a way to allow Platforms to only
```
fmui: `obs_v2` or `obs.v2`? The example above use `obs.v2`.

spec.md
```diff
@@ -570,13 +559,297 @@ schema being used.
...
+| Header | Type | Description |
+| --- | --- | --- |
+| X-Broker-API-Version* | string | See [API Version Header](#api-version-header). |
+
+\* Headers with an asterisk are REQUIRED.
+
+#### cURL
+
+```
+$ curl http://username:password@service-broker-url/v2/catalog/schemas -H "X-Broker-API-Version: 2.13"
```
fmui: The broker API version should be 2.14 in this example because this endpoint has not been defined in 2.13.

spec.md
n3wscott: oh shoot, nice catch!!! Thanks

spec.md
martinmaly: This requirement may affect adoption of schemas in general. because schemas are already in spec, this addition will result in making platforms that are compliant (and support schemas as of v2.12) non-compliant.

Could versioning, or some other signal, be used for platform to communicate to the broker: "I like schemas, but can't yet deal with references and schema endpoint", and broker can either inline schemas, or use $refs to its schema endpoint?

spec.md
```diff
@@ -570,13 +559,297 @@ schema being used.
...
 
 
+## JSON Schemas
+
+The JSON Schemas endpoint is used to define reusable definitions of JSON Schema
+Objects referenced throughout the [Catalog](#catalog-management). If a Platform
+intends to support JSON Schemas, it MUST examine the response from the catalog
+to determine if the Service Broker is using `$ref` JSON Schemas. The Service
+Broker can signal to the Platform that JSON Schema components are supported by
+using the `osb_v2:` scheme. Components are a way to allow Platforms to only
```
martinmaly: is the osb_v2 schema necessary to satisfy the requirements? 
I wonder if it is equivalent to http:///v2/catalog/schemas/

(basically a relative link).

spec.md
```diff
@@ -570,13 +559,297 @@ schema being used.
...
+Note: If the Service Broker provides a mix of `osb_v2:` in addition to other
+schemes that result in fetching both `/v2/catalog/schemas/:component_path` and
+`/v2/catalog/schemas`, the Platform will have colliding JSON Schema `$ids`. In
+this case the Platform SHOULD ignore the component paths and use the root
+schema. Service Brokers SHOULD avoid mixing `osb_v2:` and other schemes.
+
+When the osb_v2: scheme is used, the data referenced by a component path MUST
+NOT change and can only be deleted when there are no longer any references to
+it in the Service Broker's schema. Additionally, once deleted, that component
+path MUST NOT ever be used again by this Service Broker, unless the schema
```
martinmaly: which requirement leads to this strong constraint of infinite memory in the broker to never reuse schema component paths? Is there need for unlimited caching by the platforms?

spec.md
```diff
@@ -570,13 +559,297 @@ schema being used.
...
+using the `osb_v2:` scheme. Components are a way to allow Platforms to only
+fetch sub-schemas.
+
+The root JSON Schema document can be found by getting the JSON Schemas
+endpoint without a `:component_path`.
+
+If the `osb_v2:` scheme is used by the Service Broker, then the Platform MAY
+fetch the component JSON Schema directly using the following convention:
+
+ - Given: `osb_v2:///{:component_path}`, fetch
```
martinmaly: Suppose it were a goal to rely on 'pure json schema' facilities in terms of references; motivation for such goal may be using existing JSON schema libraries to fetch/resolve/interpret references. How would the proposal change with such goal in mind?

Reason I wonder is that OSB being an HTTP-based API, ideally there should be no reason why the schema support in OSB shouldn't rely on base JSON schema facilities, fetching schemas via HTTP and using combination of external and internal $refs to achieve its goals.j

Then, the only constraint we may choose to add is that all external refs (and those could be exclusively contained within the /v2/catalog response) point to that very broker's /v2/catalog/schemas endpoint.

That would lead to broker being transitively closed world of schemas.

Yet another way altogether to achieve this feature may be to drop the constraint which prohibits brokers to use external refs in their schemas, and allow brokers to serve schemas in variety of ways (similar to how dashbaord client is handled by the spec). This last option is not my favorite :) but I'd prefer solution which does less customization of JSON schema.

spec.md
n3wscott: that is technically not a relative link, it is a authority-less URL and I don't think it is enough to make assumptions on if we can capture the path and use it for the component id. The osb_v2 is the signal for a component path.

If we just went off `http:///` we might hijack someones valid implementation.

spec.md
n3wscott: The path can be fetched again, this requirement is for serving an object that does not match the previously served object from the same url. This is the spec way to say use versions without saying directly that the objects need to be versioned. for example, a one-off broker does not want to worry about versions, but their resources are essentially versioned from the fact that they are static. This gives the rule that they need to not serve a new object from the same url.

spec.md
n3wscott: This proposal works with existing json schema libraries as demonstrated in the design doc in the description. The only requirement we are really applying is that json schema libraries are not allowed to fetch schemas. They must be provided to the library, and the platform needs a mechanism to work out where to fetch these from. 

Then the broker author must make a choice, either they support `osb_v2` style uris and the fetching of sub-resources, or they don't and the broker expects the platform to get the schema objects from `/v2/schemas`

spec.md
n3wscott: The api version is already in every request. The broker could implement this if they choose to, but it is up to that broker.

spec.md
```diff
@@ -570,13 +559,297 @@ schema being used.
...
+
+The JSON Schemas endpoint is used to define reusable definitions of JSON Schema
+Objects referenced throughout the [Catalog](#catalog-management). If a Platform
+intends to support JSON Schemas, it MUST examine the response from the catalog
+to determine if the Service Broker is using `$ref` JSON Schemas. The Service
+Broker can signal to the Platform that JSON Schema components are supported by
+using the `osb_v2:` scheme. Components are a way to allow Platforms to only
+fetch sub-schemas.
+
+The root JSON Schema document can be found by getting the JSON Schemas
```
duglin: by "root" do you mean the "entire set of schema for the broker" ? If so, we can say that to be clear?

spec.md
```diff
@@ -570,13 +559,297 @@ schema being used.
...
+
+If the `osb_v2:` scheme is used by the Service Broker, then the Platform MAY
+fetch the component JSON Schema directly using the following convention:
+
+ - Given: `osb_v2:///{:component_path}`, fetch
+   `/v2/catalog/schemas/:component_path` and then the Platform is responsible
+   to continue fetching any unknown `:component_path` URIs. Care ought to be
+   taken by the platform to not fetch duplicate definitions.
+
+A Platform MAY ignore the `osb_v2:` scheme and just fetch `/v2/catalog/schemas`
```
duglin: I think we may need to add something about how the Broker MUST support this

spec.md
```diff
@@ -570,13 +559,297 @@ schema being used.
...
+
+A Platform SHOULD fetch the root Schema document for any other `$ref` URIs
+found that do not match the `osb_v2:` scheme. The Service Broker MUST provide
+the JSON Schema Object for the referenced URI in a subschema returned.
+
+Note: If the Service Broker provides a mix of `osb_v2:` in addition to other
+schemes that result in fetching both `/v2/catalog/schemas/:component_path` and
+`/v2/catalog/schemas`, the Platform will have colliding JSON Schema `$ids`. In
+this case the Platform SHOULD ignore the component paths and use the root
+schema. Service Brokers SHOULD avoid mixing `osb_v2:` and other schemes.
```
duglin: re: https://github.com/openservicebrokerapi/servicebroker/pull/402#pullrequestreview-92616345

I agree its hard but we need to say something about this. If a Broker is changing its service definitions often it has no requirement to keep the old schemas around. Which means that if the interval that a Platform has for re-listing the services is longer than the interval at which a Broker updates (and removes) its schema then the Platform will end up missing data.

Perhaps for now the best we can do is say that if the Broker returns a 404/410? from a schema GET then it SHOULD re-list since it probably means the Broker's service list and schema have changed.  Although, an existing service usage is now in trouble.  Maybe we need to say that AND that a Broker can't remove a schema as long as there is an existing instance?

Also, we should be clear on what a Broker returns when a schema is removed - 404 or 410? Should probably be a 410 to be consistent with the rest of the spec.

And we should probably say that an existing schema reference can only be updated as long as its done so in a backward compatible way.

spec.md
n3wscott: root document is a JSON Schema term.

spec.md
n3wscott: It is already in there: https://github.com/openservicebrokerapi/servicebroker/pull/402/files#diff-958e7270f96f5407d7d980f500805b1bR658

spec.md
n3wscott: >AND that a Broker can't remove a schema as long as there is an existing instance?

I can see a service class add a parameter that would be in support of the next minor version of that class, it is optional and therefor not breaking of the existing instance. The suggestion would block this behavior and that is not ideal.

> 410 to be consistent with the rest of the spec.

This means the broker needs to keep track of all schemas ever offered. I don't want to sign the broker authors up for this. 

I think we need some real implementations to test if the platform can get easily confused. I suspect the issues you are raising here are non-issues for versatile platforms. The issues at the root are pull-based APIs.

duglin:
Agreed to moving to validate thru impl on today's call

spec.md
martinmaly: can you please elaborate on the hijacking issue?

spec.md
```diff
@@ -570,13 +559,297 @@ schema being used.
...
+endpoint. `$ref` usage is assumed to be delegation (linking), not inclusion
+(replacement). Libraries that implement `$ref` content replacement result in
+adverse behaviors.
+
+For Service Brokers that support `$ref`, they:
+
+  * MUST respond with all JSON Schema documents for `/schemas` requests that are
+    referenced in the catalog.
+  * MAY provide `$ref`'s with the `osb_v2:` scheme, then:
+    * MUST respond to `/schemas/:component_id` requests.
```
martinmaly: this is the only occurrence of component_id ... I was looking for its definition (i.e. whether it is a simple identifier, or whether a path is allowed, to give broker more control).

Then I found out that elsewhere :component_path is used. Are they intended to be the same, or are they different (and the :component_id needs more clarification)?

spec.md
```diff
@@ -447,8 +449,9 @@ The following rules apply if `parameters` is included anywhere in the catalog:
 [JSON Schema draft v4](http://json-schema.org/).
 * Platforms SHOULD be prepared to support later versions of JSON schema.
 * The `$schema` key MUST be present in the schema declaring the version of JSON
-schema being used.
-* Schemas MUST NOT contain any external references.
+schema being used unless the schema is composed of only a `$ref`.
+* Schemas MUST NOT contain any external references except for schemas found
+  under `/schemas{/*}` (see [$ref Usage](#ref-usage) for further information).
```
martinmaly: is the correct path under which $refs can point: `/v2/catalog/schemas/...`  ?

the construct `{/*}` doesn't appear defined anywhere

spec.md
martinmaly: I'd recommend clarifying the construct {:component_path}, specifically the curly braces. Quick scan through the spec indicates there is no precedent and without specifying what it means it is open to interpretation and to confusion.

spec.md
```diff
@@ -570,13 +559,297 @@ schema being used.
...
+#### cURL
+```
+$ curl http://username:password@service-broker-url/v2/catalog/schemas/:component_path -H "X-Broker-API-Version: 2.13"
+```
+
+### Response
+
+| Status Code | Description |
+| --- | --- |
+| 200 OK | MUST be returned upon successful processing of this request. The expected response body is below. |
```
martinmaly: Should 304s be allowed? That would allow better controls over cache-ability of schemas.

The response codes seem written from perspective of the broker (broker returns ...), but there may be other responses that the platform may see, i.e. 304 if there is a caching proxy in-between. Would it be worth clarifying, or is it assumed?

spec.md
n3wscott: there is nothing that prevents an author of json schema to make their document `$id` to be `http:///something/` with sub-documents be appended. So if we redirect `http:///` to mean schema fetch from broker, then we might be smashing someones implementation. TBH there is nothing that prevents existing impls to use `osb_v2:///` today, it would just be a really strange choice for them.

spec.md
n3wscott: Tried to clear the wording up like this:

* Schemas MUST NOT contain any external references except for schemas found
  under `/v2/catalog/schemas` or `osb_v2:` style `$ref`s via
  `/v2/catalog/schemas/{:component_path}` (see [$ref Usage](#ref-usage) for
  further information).

My intent is to say `osb_v2` style external refs are ok, and standard `http://my.host.com/my/schema/doc.json#moreInfo` will work too as long as that can be found in `/v2/catalog/schemas`. It is subtle in the spec, and only people who understand JSON schema spec will understand this is allowed. But if you don't understand json schema spec you will get it wrong and break client side processing so I think this requirement is ok.

spec.md
n3wscott: You are correct! the spec just uses :yada to mean yada is a var, removed the `{}`.

spec.md
n3wscott: I think this is an interesting idea, and perhaps it can be a followup because knowing how to send a 304 requires some kind of etag or cache-control header to also be sent and there is nothing like that in the spec today. If this is a requirement, please make an issue and we can follow up with a spec amendment.

n3wscott:
Hello everyone, I have taken another shot at making `$ref` useful using the new `/v2/catalog/schemas` endpoint. I have realized that `/v2/catalog/schemas/:component_path` + `:component_path` == `''` is the same as `/v2/catalog/schemas`, so a empty component path IS the root document and flattened the two definitions. I hope this makes it more clear.

I also rewrote the language around what it means to use external URIs for `$ref` and am trying to make the root document optional. REALLY complex brokers find it hard to provide a aggregated master document, and would rather only support just the sub-documents via `:component_path`. 

Please take a look and let me know if you think this is more clear, I am working on a demo of this in the background. No eta for the demo yet.

n3wscott:
@mattmcneeney as an update to this question:

> Question: if a broker author simply wants to use the same reference for every schema they offer for create/update/bind, then is there a shorter way for them to do this than the below for each one:

```
"service_binding": {
  "create": {
    "parameters": {
      "$ref": "osb_v2:///fakeservice/v1/bindings#AccountParameters"
    }
  }
}
```

If the resource is really the same, the broker can have the following:

```
"service_binding": {
  "create": {
    "parameters": {
      "$ref": "osb_v2:///"
    }
  }
}
```

with some caveats that the schema should never change. Add a version to be safe:

```
"service_binding": {
  "create": {
    "parameters": {
      "$ref": "osb_v2:///v1"
    }
  }
}
```

mattmcneeney:
Closing as this is no longer needed according to @n3wscott

PR #407, opened 2018-01-10
deprecate orphan mitigation and add failure cleanup sections

### LINKED ISSUE #369: The term "orphan" has become ambiguous
With the addition of the text about platforms sending a deprovision request for every resource that has attempted to be provisioned, the term "orphan" has become quite muddied. The orphan mitigation section of the spec specifically highlights the situation where the platform is unsure whether a resource has been successfully created or not. In this context, an "orphan" is specifically an instance that has been provisioned *successfully*, but the platform doesn't know about it or has given up on it. However, with the latest changes, we have started using the term "orphan" to also encapsulate situations where both the broker and the platform have agreed that the resource has failed but there may be broken resources left over on the broker side.

I believe we should take a hard stance here, clarifying that an "orphan" only refers to successfully provisioned resources that the platform doesn't know about. This way, it's clear that these are two separate situations, with potentially slightly different semantics.

### LINKED ISSUE #370: SHOULD vs MUST with orphan mitigation
In the new text about platforms sending a deprovision request for every resource that has attempted to be provisioned, we specify that the platform MUST send a deletion request. However, the existing orphan mitigation section says that the platform SHOULD attempt a deletion for the specific orphan mitigation circumstances. As neither section specifies the timing as to when this deletion request will happen, the new text basically overwrites the entirety of the orphan mitigation section.

We should clarify our intent here. Should orphan mitigation happen immediately upon the platform deciding to mark the resource as failed? If so, we should specify that here. If not, then is orphan mitigation no longer relevant? We should then rework this section entirely to just talk about cleaning up of resources in general.

### LINKED ISSUE #397: Clarify wording around how to clean up stray resources when a provision/bind fails
We currently have slightly conflicting and unclear messaging around cleaning up failed provisions/binds spread throughout the spec. At certain sections we specify that the Platform MUST send a deprovision/unbind request at some time, but in other sections we also specify that the Platform SHOULD initiate orphan mitigation for a subset of these situations.

As a group, we have decided that we should clarify this wording, deconstructing the orphan mitigation section and applying it on a case-by-case basis to where in the spec it matters. I am currently working on a PR that fixes this.

kibbles-n-bytes:
This PR deprecates orphan mitigation and adds a failure cleanup section to the Provisioning and Binding sections.

Closes #397, closes #370, closes #369

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -715,9 +715,9 @@ For success responses, the following fields are valid.
 }
 ```
 
-If the successful response includes a `state` of `failed` then the Platform
-MUST send a deprovision request to the Service Broker to prevent an orphan
-being created on the Service Broker.
+If the successful response includes a `state` of `failed`, then the Platform
+MUST clean up the failed resource as per the [Provisioning Failure
+Cleanup](#failure-cleanup) section.
 
 ### Polling Interval and Duration
 
@@ -818,12 +818,10 @@ $ curl http://username:password@service-broker-url/v2/service_instances/:instanc
 | 409 Conflict | MUST be returned if a Service Instance with the same id already exists but with different attributes. The expected response body is `{}`, though the description field MAY be used to return a user-facing error message, as described in [Service Broker Errors](#service-broker-errors). |
 | 422 Unprocessable Entity | MUST be returned if the Service Broker only supports asynchronous provisioning for the requested plan and the request did not include `?accepts_incomplete=true`. The expected response body is: `{ "error": "AsyncRequired", "description": "This Service Plan requires client support for asynchronous service operations." }`, as described below (see [Service Broker Errors](#service-broker-errors). |
 
-Responses with any other status code will be interpreted as a failure and a
-deprovision request MUST be sent to the Service Broker to prevent an orphan
-being created on the Service Broker.
-
-Service Brokers can include a user-facing message in the `description` field;
-for details see [Service Broker Errors](#service-broker-errors).
+Request timeouts, responses with any other status code, or malformed response
+bodies will be interpreted as failures. Service Brokers can include a user-facing
+message in the `description` field; for details see [Service Broker
+Errors](#service-broker-errors).
 
 #### Body
 
@@ -848,6 +846,15 @@ error responses, see [Service Broker Errors](#service-broker-errors).
   "operation": "task_10"
 }
 ```
+
+### Failure Cleanup
+
+There could potentially be side effects on the Service Broker as a result of an
+unsuccessful provision attempt. If a provision request has been sent to a Service
+Broker, then the Platform MUST send a deprovision request at some point before
+cleaning up its internal representation of the Service Instance. If the
+deprovision attempt fails, the Platform SHOULD keep retrying until the Service
+Broker responds with a success or the Platform reaches its internal retry limit.
 
 ## Updating a Service Instance
 
@@ -1182,9 +1189,8 @@ $ curl http://username:password@service-broker-url/v2/service_instances/:instanc
 | 409 Conflict | MUST be returned if a Service Binding with the same id, for the same Service Instance, already exists but with different parameters. The expected response body is `{}`, though the description field MAY be used to return a user-facing error message, as described in [Service Broker Errors](#service-broker-errors). Additionally, if the Service Broker rejects the request due to a concurrent request to create a binding for the same Service Instance, then this error MUST be returned (see [Blocking Operations](#blocking-operations)). |
 | 422 Unprocessable Entity | MUST be returned if the Service Broker requires that `app_guid` be included in the request body. The expected response body is: `{ "error": "RequiresApp", "description": "This service supports generation of credentials through binding an application only." }` (see [Service Broker Errors](#service-broker-errors). |
 
-Responses with any other status code will be interpreted as a failure and an
-unbind request MUST be sent to the Service Broker to prevent an orphan being
-created on the Service Broker. Service Brokers can include a user-facing
+Request timeouts, responses with any other status code, or malformed response
+bodies will be interpreted as failures. Service Brokers can include a user-facing
 message in the `description` field; for details see [Service Broker
 Errors](#service-broker-errors).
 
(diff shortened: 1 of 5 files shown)


spec.md
```diff
@@ -1258,6 +1264,15 @@ can be mounted on all app instances simultaneously.
...
 ```
 
+### Failure Cleanup
+
+After an unsuccessful bind attempt, a Service Broker MAY have stray resources
+persisted as a side effect. In order to prevent resource leaks, if a bind request
+has been sent to a Service Broker, then the Platform MUST send an unbind request
+at some point before cleaning up its internal representation of the Service
+Binding. Should that unbind attempt fail, the Platform SHOULD keep retrying until
+the Service Broker responds with a success.
```
pmorie: Do we want to say 'retry up to its internal limit' or something similar?  We mention limits of polling elsewhere.

One thing that would be useful is the ability to list service instances and bindings.  However, I suggest that that is such a radical change, and so unlikely to be implemented correctly (broker authors will have to implement acl filtering, for example), that we should consider 'list' and 'watch' in the API to be v3 things.

spec.md
```diff
@@ -1460,6 +1475,10 @@ example, to specify a description in a different language.
 
 ## Orphans
 
+*Deprecated. Refer to the failure cleanup sections for
```
pmorie: It's not clear exactly what deprecation means here.  Are these for reference, or are we saying this is wholly replaced by the new language?

angarg12:
Travis is printing a couple of cacophonic errors


> line 856: 'Should' should be 'SHOULD'
> line 1273: 'Should' should be 'SHOULD'

spec.md
kibbles-n-bytes: I'm happy to add a mention of an internal limit. The original text didn't mention one in the orphan mitigation case, but I think it's a useful weakening, especially since polling already has the concept.

spec.md
```diff
@@ -849,6 +847,15 @@ error responses, see [Service Broker Errors](#service-broker-errors).
 }
 ```
 
+### Failure Cleanup
+
+After an unsuccessful provision attempt, a Service Broker MAY have stray
+resources persisted as a side effect. In order to prevent resource leaks, if a
```
n3wscott: the use of MAY is funny here, this reads as the Service broker is allowed to have stray resources. Is this intended? If it is, then MAY is the correct wording.

spec.md
n3wscott: Can this whole Orphans section be removed? I am also unclear on which part becomes deprecated.

kibbles-n-bytes: Since the new language encapsulates all the cases of the orphan mitigation section, the text is irrelevant now. I didn't rip it out entirely only because I thought we'd be more comfortable giving existing spec users some time to digest the change. This way they wouldn't be surprised when they go to the bottom of the spec expecting to see the orphan mitigation section and there's nothing there.

spec.md
kibbles-n-bytes: That is intended. From previous discussions on this topic, people were saying they wanted the option for Service Brokers to have broken remnants of resources left behind after a failure so that they can troubleshoot.

spec.md
n3wscott: > be more comfortable giving existing spec users some time to digest the change

I vote remove it and add a note in the release notes to address this concern.

kibbles-n-bytes: I'm happy with that as well if others feel that way!

spec.md
```diff
@@ -818,12 +818,10 @@ $ curl http://username:password@service-broker-url/v2/service_instances/:instanc
...
 | 422 Unprocessable Entity | MUST be returned if the Service Broker only supports asynchronous provisioning for the requested plan and the request did not include `?accepts_incomplete=true`. The expected response body is: `{ "error": "AsyncRequired", "description": "This Service Plan requires client support for asynchronous service operations." }`, as described below (see [Service Broker Errors](#service-broker-errors). |
 
-Responses with any other status code will be interpreted as a failure and a
-deprovision request MUST be sent to the Service Broker to prevent an orphan
-being created on the Service Broker.
-
-Service Brokers can include a user-facing message in the `description` field;
-for details see [Service Broker Errors](#service-broker-errors).
+Request timeouts, responses with any other status code, or malformed response
+bodies will be interpreted as failures. Service Brokers can include a user-facing
```
duglin: Anyone else bothered by the use of the word "will"?  I know its not new for this PR, but what do people think about making it a MUST to make it clear that there is no ambiguity about it, it IS A FAILURE. The MUST wouldn't place any new requirements on the Platform w.r.t. what it does about it (it could choose to ignore the failure), but at least from a spec perspective we're very clear on what *we* think of these types of responses. This may not be the only spot in the spec for this either - so if people agree I will open a new PR/issue for it.

spec.md
duglin: Toughie... but, the MAY there is kind of awkward to read because, as scott said, it almost encourage people to leave stuff around.  I wonder if we could reword it a little to get the same net result:
```
After an unsuccessful provision attempt, the Service Broker might not be able to,
or might not want to, immediately delete any resources persisted as a side effect
of the provision request. In order to prevent...
```
This way we're not encouraging, or discouraging, it - we're just stating our reason
for why the Platform needs to send the deprovision.

spec.md
duglin: +1 to pulling it out

spec.md
kibbles-n-bytes: Hmm... Yeah, I'd also prefer if we didn't sound like we recommend it, so I'll get rid of the "MAY". The wording here's a bit tricky because it's also trying to encapsulate the timeout situation where the Service Broker doesn't realize yet the Platform considered the request a failure. I'd prefer keeping the motivation as general as possible.

In fact, it's not really "stray resources" we should care about; it's more generally just any side effects, so how about:

`There could potentially be side effects on the Service Broker as a result of an unsuccessful provision attempt. If a provision request has been sent...`

spec.md
kibbles-n-bytes: That seems reasonable, let's open an issue for it!

spec.md
codevulture: Can we say that using failure cleanups mentioned above, we could be 100% sure that orphans are removed, for e.g: considering a case when `the Platform reaches its internal retry limit.` and resource is not yet cleaned up. In such a case we have no way to clean up the resource when we are considering internal timeouts . I am okay with the change, just wanted to double check.

kibbles-n-bytes: I don't think there's a way we can guarantee all orphaned resources are removed in every case. This should encompass all situations where a broker is at least somewhat responsive. If a broker just stops accepting deletion requests permanently or takes an inordinate amount of time to accept deletion of a resource, though, I think it's reasonable to say the Platform is not expected to keep sending deletion requests forever in the hopes that one day it'll succeed. Presumably the Platform would let the user know that the cleanup failed, so they can out-of-band figure out what happened.

kibbles-n-bytes:
@mattmcneeney Yup, working on that today!

n3wscott:
Kibbles asked to close this PR as irrelevant.

PR #409, opened 2018-01-10
Consistent error object and error codes throughout the specification

mattmcneeney:
This change moves the error codes listed throughout the spec into a single section and makes it clearer that any error response to return the same fields (`error` and `description`).

From me & @n3wscott


n3wscott:
The script for RFC SHOULD vars is sort of annoying. Are we not allowed to use the word should without it having a spec meaning? SHOULD is spec, should is english. I think the script is not useful. What do you think @duglin ?

LGTM

duglin:
the RFC doesn't distinguish between SHOULD and should - both are keywords. They are often upper-cased to help stand-out, but also so its clear that we meant to use it as normative keyword and didn't just use it by mistake.

avade:
LGTM

spec.md
```diff
@@ -245,6 +243,32 @@ querying of the Service Broker's catalog, the Platform might not have an
...
 originating identity header would not be included in those messages.
 
+## Service Broker Errors
+
+When a request to a Service Broker fails, the Service Broker MUST return an
+appropriate HTTP response code. If an appropriate response code is not defined
+within this specification, Service Broker authors SHOULD chose a code that
+accurately reflects the nature of the failure.
+
+For error responses, the following fields are valid:
```
fmui: Does this prohibit broker specific error fields?

spec.md
```diff
@@ -245,6 +243,32 @@ querying of the Service Broker's catalog, the Platform might not have an
...
+| description | string | A meaningful error message explaining why the request failed. | 
+
+### Error Codes
+
+There are failure scenarios described throughout this specification for which
+the `error` field MUST contain a specific string. Service Broker authors MUST
+use these error codes for the specified failure scenarios. For other scenarios,
+Service Broker authors SHOULD choose a suitable one-word error code.
+
+| Error | Reason | Expected Action |
```
fmui: The error codes are now defined twice in the spec. This may lead to inconsistencies in the future. I don't think we need them here.

spec.md
```diff
@@ -245,6 +243,32 @@ querying of the Service Broker's catalog, the Platform might not have an
...
+
+| Response Field | Type | Description |
+| --- | --- | --- |
+| error | string | A single word uniquely identifying the error condition. |
+| description | string | A meaningful error message explaining why the request failed. | 
+
+### Error Codes
+
+There are failure scenarios described throughout this specification for which
+the `error` field MUST contain a specific string. Service Broker authors MUST
```
fmui: This MUST could be a breaking change.

spec.md
n3wscott: Not at all, the broker may always add fields to the response, but it can't assume a platform will not ignore them.

spec.md
n3wscott: The MUST was already in use throughout the spec, we just consolidated the meaning of the error codes, and they continue to be called out in certain situations as required, for `AsyncRequired` for example, the broker must report that back to the platform if it requires the `accepts_incomplete=true` parameter. 

What we have done is centralized the definition of the error codes and removed some confusing language around calling out what exactly needs to be returned further down in the spec.

spec.md
n3wscott: the intent is now each time they are referenced later in the spec, they refer back to this section. If we missed a spot, that is an issue with the PR and we would like to catch and fix it.

n3wscott:
> the RFC doesn't distinguish between SHOULD and should - both are keywords.

fair enough. I will just bookmark http://www.thesaurus.com/ now. 😆

n3wscott:
LGTM

angarg12:
LGTM

spec.md
fmui: You are the native speaker. ;-)

But it wouldn't hurt to clarify this with a few sentences for people like me. Something along those lines: 
"Brokers MAY add more fields, which MUST be ignored by the platform.  Broker author SHOULD NOT use generic field names that may clash with future specification updates. It is recommended choose field names with a company prefix.”

spec.md
```diff
@@ -245,6 +243,32 @@ querying of the Service Broker's catalog, the Platform might not have an
 end-user with which to associate the request, therefore in those cases the
 originating identity header would not be included in those messages.
 
+## Service Broker Errors
+
+When a request to a Service Broker fails, the Service Broker MUST return an
+appropriate HTTP response code. If an appropriate response code is not defined
+within this specification, Service Broker authors SHOULD chose a code that
```
duglin: s/chose/choose/

spec.md
duglin: s/valid/defined/.

Then afterwards:
Service Brokers MAY include additional fields within the response. When adding new fields, it is SHOULD use a unique prefix on the field names to reduce the chances of conflict with future specification defined fields.

I don't think we can, or should , say what a Platform can do with this data since they can ignore all of it if they want.  Also, this "add a prefix" thing might be good for any other spot in the spec where we allow extension json fields.

spec.md
```diff
@@ -245,6 +243,32 @@ querying of the Service Broker's catalog, the Platform might not have an
...
+accurately reflects the nature of the failure.
+
+For error responses, the following fields are valid:
+
+| Response Field | Type | Description |
+| --- | --- | --- |
+| error | string | A single word uniquely identifying the error condition. |
```
duglin: We can do this in a follow-on PR if we want, but we may want to be more precises on this. For example, are underscores ok? Should we RECOMMEND camelCase since likely its not just a single word it multiple words w/o spaces. Perhaps even the word "word" isn't appropriate.

spec.md
```diff
@@ -245,6 +243,32 @@ querying of the Service Broker's catalog, the Platform might not have an
...
+| description | string | A meaningful error message explaining why the request failed. |
```
duglin: I think in other spots we talk about thing being "user-facing" to imply it should be human readable, perhaps we could use that wording since "meaningful" could be interpreted as "meaningful to a machine so its easily parse-able".

spec.md
```diff
@@ -245,6 +243,32 @@ querying of the Service Broker's catalog, the Platform might not have an
...
+| --- | --- | --- |
+| error | string | A single word uniquely identifying the error condition. |
+| description | string | A meaningful error message explaining why the request failed. | 
+
+### Error Codes
+
+There are failure scenarios described throughout this specification for which
+the `error` field MUST contain a specific string. Service Broker authors MUST
+use these error codes for the specified failure scenarios. For other scenarios,
+Service Broker authors SHOULD choose a suitable one-word error code.
```
duglin: Isn't this a dup of what's on line 250?

spec.md

spec.md

spec.md

spec.md
```diff
@@ -245,6 +243,32 @@ querying of the Service Broker's catalog, the Platform might not have an
...
+
+| Error | Reason | Expected Action |
+| --- | --- | --- |
+| AsyncRequired | This request requires client support for asynchronous service operations. | The query parameter `accepts_incomplete=true` MUST be included the request. |
+| ConcurrencyError | The Service Broker does not support concurrent requests that mutate the same resource. | Clients SHOULD wait until pending requests have completed for the specified resources. |
```
duglin: I wonder if this should be a MUST instead. I would think that for the same resource the broker wouldn't change whether it supports concurrency or not - unless someone can think of a case where the use of this error would vary based on which fields in the resource are being touched.

spec.md

spec.md
mattmcneeney: Yes - removed!

spec.md
mattmcneeney: I remember saying it's difficult to specify MUSTs for a client. Hmm.

spec.md
```diff
@@ -245,6 +243,32 @@ querying of the Service Broker's catalog, the Platform might not have an
...
+| RequiresApp | The Service Broker requires that `app_guid` be included in the request body. | Include `app_guid` in the request body. |
```
duglin: I like the use of the RFC2119 keywords in the other errors - would it make sense to say something like "The app_guid MUST be included in the request body" ?

mattmcneeney:
Updated, thanks @duglin for the feedback!

spec.md
```diff
@@ -600,45 +616,37 @@ operation.
...
 ## Polling Last Operation
 
 When a Service Broker returns status code `202 Accepted` for
 [Provision](#provisioning), [Update](#updating-a-service-instance), or
 [Deprovision](#deprovisioning), the Platform will begin polling the
 `/v2/service_instances/:instance_id/last_operation` endpoint to obtain the
-state of the last requested operation. The Service Broker response MUST
-contain the field `state` and MAY contain the field `description`.
+state of the last requested operation. The Service Broker response MUST contain
+the field `state` and MAY contain the field `description`.
```
duglin: Just curious - did you manually notice that "contain" could fit on the previous line or does your editor automatically move it up for you?  Unless there's some other change in there I'm not seeing.

spec.md
```diff
@@ -1179,25 +1158,16 @@ $ curl http://username:password@service-broker-url/v2/service_instances/:instanc
...
 
-All response bodies MUST be a valid JSON Object (`{}`). This is for future
-compatibility; it will be easier to add fields in the future if JSON is
-expected rather than to support the cases when a JSON body might or might not
-be returned.
-
-For success responses, the following fields are supported. Others will be
-ignored. For error responses, see [Service Broker
-Errors](#service-broker-errors).
+For success responses, the following fields are valid:
```
duglin: not a fan of 'valid', I prefer "defined" but this might be a global change - so its up to you.

spec.md
```diff
@@ -1389,80 +1353,32 @@ $ curl 'http://username:password@service-broker-url/v2/service_instances/:instan
...
-
 ## Orphans
 
-The Platform is the source of truth for Service Instances and bindings. Service
-Brokers are expected to have successfully provisioned all the Service Instances
-and bindings that the Platform knows about, and none that it doesn't.
+The Platform is the source of truth for Service Instances and Service Bindings.
+Service Brokers are expected to have successfully provisioned all of the Service
+Instances and Service Bindings that the Platform knows about, and none that it
+doesn't.
```
duglin: I know you just added the word "Service" in there so you didn't really change anything, but this paragraph is really kind of weird.  While there is no normative text in there so people can basically ignore it, putting requirements on one component based on what another component "knows about" just sounds really odd. Normally, this requirement would be backed up with normative text in the doc around our ops and thru people adhering to those ops we would implicitly get the correct net results. But just saying "make sure component A follow's component B's state" when its this spec's job to ensure it just reads oddly to me. No action required, just rambling... :-)

spec.md
duglin: We use MUST in the previous one, so we wouldn't be inconsistent.

I would need to think more about whether a MUST for a client is an issue or not in general, but I think in this case it makes a lot of sense. This table is basically telling the client why it got an error, so saying that it "MUST do xxx" to avoid the error seems appropriate to me. Saying it "MAY do xxx" could be interpreted as "there might be other things it could do", or "try it again with the same data and next time maybe it'll work" - neither of which is what we're trying to say. I think the "expected action" column saying "you MUST do this to avoid the error" is exactly what people are looking for - they want to know what to do to fix it.

spec.md
mattmcneeney: Manually noticed it :)

spec.md
mattmcneeney: I agree - changing

spec.md
mattmcneeney: SGTM, and this is consistent with line 269

mattmcneeney:
Another update (thanks @duglin)

n3wscott:
LGTM

[Approved with PullApprove](https://pullapprove.com/openservicebrokerapi/servicebroker/pull-request/409/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=openservicebrokerapi/servicebroker)

spec.md
duglin: @mattmcneeney what about this one?  MUST?

spec.md
```diff
@@ -245,6 +243,34 @@ querying of the Service Broker's catalog, the Platform might not have an
...
+
+There are failure scenarios described throughout this specification for which
+the `error` field MUST contain a specific string. Service Broker authors MUST
+use these error codes for the specified failure scenarios.
+
+| Error | Reason | Expected Action |
+| --- | --- | --- |
+| AsyncRequired | This request requires client support for asynchronous service operations. | The query parameter `accepts_incomplete=true` MUST be included the request. |
+| ConcurrencyError | The Service Broker does not support concurrent requests that mutate the same resource. | Clients SHOULD wait until pending requests have completed for the specified resources. |
+| RequiresApp | The `app_guid` MUST be included in the request body. | Include `app_guid` in the request body. |
```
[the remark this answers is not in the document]
n3wscott: The action is how a platform is expected to react.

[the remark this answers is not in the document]
mattmcneeney: I agree - updating now

spec.md
```diff
@@ -572,19 +597,12 @@ Note: Asynchronous operations are currently supported only for provision,
...
-```
-{
-  "error": "AsyncRequired",
-  "description": "This Service Plan requires client support for asynchronous service operations."
-}
-```
+`accepts_incomplete=true` MUST be included the request. If the parameter is not
+included or is set to `false`, and the Service Broker cannot fulfil the request
+synchronously (guaranteeing that the operation is complete on response), then
+the Service Broker SHOULD reject the request with the status code `422
```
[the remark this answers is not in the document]
mattmcneeney: Is this a breaking change though?

mattmcneeney:
PR updated! @duglin @n3wscott

avade:
LGTM

spec.md
```diff
@@ -245,6 +243,34 @@ querying of the Service Broker's catalog, the Platform might not have an
...
+
+For error responses, the following fields are defined. Service Brokers MAY
+include additional fields within the response. When adding new fields, Service
+Brokers SHOULD use a unique prefix for the field names to reduce the chances of
+conflict with future specification defined fields.
+
+| Response Field | Type | Description |
+| --- | --- | --- |
+| error | string | A single word in camel case that uniquely identifies the error condition. |
+| description | string | A user-facing error message explaining why the request failed. |
```
duglin: I think we may need to add the standard "If present, MUST be a non-empty string." to these fields.

spec.md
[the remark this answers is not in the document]
duglin: Created https://github.com/openservicebrokerapi/servicebroker/issues/412 as a reminder to do the investigation

duglin:
One minor tweak, otherwise it looks good.

mattmcneeney:
@duglin done!

Ready for reviews

cc @n3wscott @avade etc

duglin:
merci.
LGTM

n3wscott:
LGTM

angarg12:
@mattmcneeney we need rebase!

n3wscott:
LGTM

angarg12:
LGTM

btbd:
@mattmcneeney has a few todos from today's call

duglin:
From today's call, @mattmcneeney will make sure we don't lose the current suggested wording for the "description" properties.

n3wscott:
LGTM

spec.md
mattmcneeney: Changed to MUST

mattmcneeney:
This PR has been updated following yesterday's call. Where we used to specify the `description` field of the error response, we now provide a helpful error message that a broker MAY respond via the `description` field. Let me know what you think / ready for review!

spec.md
```diff
@@ -245,6 +243,34 @@ querying of the Service Broker's catalog, the Platform might not have an
 end-user with which to associate the request, therefore in those cases the
 originating identity header would not be included in those messages.
 
+## Service Broker Errors
+
+When a request to a Service Broker fails, the Service Broker MUST return an
+appropriate HTTP response code. If an appropriate response code is not defined
+within this specification, Service Broker authors SHOULD choose a code that
```
duglin: This SHOULD bothers me because of the MUST in the first sentence. SHOULD implies something is optional, so what's the alternative to "choosing a code that accurately reflects the nature of the failure"? Choosing a random value? The MUST prevents that.

Perhaps just:
- - - - 
When a request to a Service Broker fails, the Service Broker MUST return an
appropriate HTTP response code. When the specification defines the expected
response code that value MUST be used.
- - - - 


Not sure we need to say anything more.

duglin:
Just one minor thing - otherwise LGTM

spec.md
mattmcneeney: Good spot, thanks @duglin! Updated so need your LGTM again.

duglin:
thanks
LGTM

avade:
LGTM

fmui:
LGTM

[Approved with PullApprove](https://pullapprove.com/openservicebrokerapi/servicebroker/pull-request/409/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=openservicebrokerapi/servicebroker)

Issue #408, opened 2018-01-10
Blocking Operations and 409 Error Codes.


n3wscott:
The Blocking Operations section of the spec says "MUST return 422" but creating a binding and getting a conflict has the error language as:

`| 409 Conflict | MUST be returned if a Service Binding with the same id, for the same Service Instance, already exists but with different parameters. Additionally, if the Service Broker rejects the request due to a concurrent request to create a binding for the same Service Instance, then this error MUST be returned (see [Blocking Operations](#blocking-operations)). |`

This, ironically, conflicts. 

The spec needs to either remove the requirement for MUST 422 in the case of a ConcurrencyError, OR move the Blocking Operation text to the 422 for bindings. OR option 3?

mattmcneeney:
I'd vote for using 422 for any blocking operation. I'm guessing we can remove this MUST (breaking change) since it is a conflict in the spec.

@avade what does ODB return today when it tries to create a binding on an instance that has an operation in progress?

Any other broker authors please let us know too!

duglin:
+1 to a 422.
PR: https://github.com/openservicebrokerapi/servicebroker/pull/413

PR #413, opened 2018-01-16
Resolve 409/422 conflict on concurrent updates

duglin:
Closes: #408 

Moved the "Additionally..." sentence to the 422 section.


angarg12:
Would this be a breaking change since we add MUST to the error code?

duglin:
We have an inconsistency in the spec right now (saying MUST use 422 in one spot and MUST use 409 in another spot for the same condition), so yes something will technically break but we can't avoid it.

mattmcneeney:
This is going to conflict with #409... Let's get that one through first!

duglin:
rebased

angarg12:
LGTM

duglin:
3 needed?

vaikas-google:
LGTM

mattmcneeney:
LGTM

avade:
LGTM

zrob:
I went back to the [2.7 spec](http://docs.pivotal.io/pivotalcf/1-6/services/api.html#blocking) and noticed the text actually indicated that it should be a 400.  It seems we've previously introduced a breaking change to get into this ambiguous state of 422/409.

```
The Cloud Controller ensures that service brokers do not receive requests for an instance while an asynchronous operation is in progress. For example, if a broker is in the process of provisioning an instance asynchronously, the Cloud Controller will not allow any update, bind, unbind, or deprovision requests to be made through the platform. A user who attempts to perform one of these actions while an operation is already in progress will get an HTTP 400 with error message “Another operation for this service instance is in progress.”
```

Given the current state I'm in favor of moving forward to the now suggested 422, but it is concerning that our "refactor only" changes are resulting in breaking changes.

duglin:
@zrob the change was made here: https://github.com/openservicebrokerapi/servicebroker/pull/193/files
If I'm remembering correctly the change was made because the old text talked about two things:
1 - what the end user should see, not what the REST API from the Broker should return - which is technically out of scope for the spec
2 - how the platform needs to block concurrent requests to the Broker

The 2nd one is actually the critical piece here because it means that the spec was written with the assumption that concurrent requests would never make it to the broker, so the spec was basically silent on this point. Which means we didn't technically break anything other than to go from "undefined" to "well-defined semantics" in an error situation. I guess you could claim that the Broker was free to do anything before and now we're forcing a particular error and that's a breaking change, but I guess at the time we decided interop was more important.

I believe, at the same time, we decided that having Platform block concurrent requests was an unnecessary restriction since some brokers might be able to deal with it just fine, and it may not always be possible for a Platform to serialize all requests for the same instance if its a multi-threaded/process/component infrastructure.

PR #415, opened 2018-01-18
Add space_guid to Cloud Foundry bind_resource extension in profile.md

mattmcneeney:
Cloud Foundry sends an additional field in the `context` object when creating a service binding called `bind_resource.space_guid`, which indicates the Cloud Foundry space that the binding is associated with. This is not to be confused with the `space_guid` currently found in the `context` object, which indicates the space in which the service instance was/is going to be created.

### DIFF
--- a/profile.md
+++ b/profile.md
@@ -13,11 +13,17 @@ or REQUIRED level requirements defined herein.
 ## Table of Contents
 
 - [Notations and Terminology](#notations-and-terminology)
+  - [Notational Conventions](#notational-conventions)
+  - [Terminology](#terminology)
 - [Originating Identity Header](#originating-identity-header)
+  - [Cloud Foundry Originating Identity Header](#cloud-foundry-originating-identity-header)
+  - [Kubernetes Originating Identity Header](#kubernetes-originating-identity-header)
 - [Context Object](#context-object)
   - [Context Object Properties](#context-object-properties)
-  - [Cloud Foundry](#cloud-foundry)
-  - [Kubernetes](#kubernetes)
+  - [Cloud Foundry Context Object](#cloud-foundry-context-object)
+  - [Kubernetes Context Object](#kubernetes-context-object)
+- [Bind Resource Object](#bind-resource-object)
+  - [Cloud Foundry Bind Resource Object](#cloud-foundry-bind-resource-object)
 - [Service Metadata](#service-metadata)
   - [Cloud Foundry Service Metadata](#cloud-foundry-service-metadata)
 
@@ -50,7 +56,7 @@ Note that when both the originating identity HTTP Header and the Context
 object appear in the same message the `platform` value MUST be the same
 for both.
 
-### Cloud Foundry
+### Cloud Foundry Originating Identity Header
 
 *`platform` Value*: `cloudfoundry`
 
@@ -73,7 +79,7 @@ would appear in the HTTP Header as:
 X-Broker-API-Originating-Identity: cloudfoundry eyANCiAgInVzZXJfaWQiOiAiNjgzZWE3NDgtMzA5Mi00ZmY0LWI2NTYtMzljYWNjNGQ1MzYwIiwNCiAgInVzZXJfbmFtZSI6ICJqb2VAZXhhbXBsZS5jb20iDQp9
 ```
 
-### Kubernetes
+### Kubernetes Originating Identity Header
 
 *`platform` Value*: `kubernetes`
 
@@ -111,8 +117,6 @@ In the [Open Service Broker API specification](spec.md) there are certain
 message flows that include a `context` property. This property is defined
 as an opaque JSON object that is meant to contain contextual information
 about the environment in which the Platform or Application is executing.
-For example, it might include the organizational information (eg. a
-Cloud Foundry `organization` GUID) in which the Application is owned.
 
 While the `context` property is defined as an opaque JSON object, in practice,
 it is often useful and necessary for there to be an agreed upon set of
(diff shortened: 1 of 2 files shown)


duglin:
Is there any reason you didn't add this to the real bind_resource object? It says:
```
Platforms MAY choose to add additional ones as needed
```

mattmcneeney:
@duglin Mainly because that object already contains CF-specific fields that we will likely want to move out of that object in future versions. The `context` object is designed for platform-specific fields like this, and so we thought we should start setting this precedent. What do you think?

duglin:
Are you thinking that the bind_resource object will go away? I actually don't think "app_guid" or "route" aren't too CF specific (maybe just a little, but tolerable).  Anyway, the text for for bind_resource says:
```
The bind_resource object contains Platform specific information related to the context in
which the service will be used.
```
so it sounds like a direct match to me. But, if there's confusion around when to use "bind_resource" and "context" we should probably make that clearer in the spec.

mattmcneeney:
No I don't think `bind_resource` will go away. I'm happy to put the new field in the top-level `bind_resource` object and I agree that it sounds suitable, but we should discuss as a group where we want things like this to live moving forward. I've added this to the agenda for next week's call.

mattmcneeney:
I guess it also depends on whether or not we want multiple extension points for Platform specific information, or just one with `context`. Food for thought...

profile.md
```diff
@@ -142,17 +140,16 @@ Note that when both the originating identity HTTP Header and the Context
...
 
 The following properties are defined for usage within a Cloud Foundry
 deployment:
 
-- `organization_guid`<br>
-  The organization GUID as defined by the Cloud Foundry
-  specification/project. This property MUST be a non-empty string serialized
-  as follows:
+- `organization_guid`  
+  The GUID of the organization that a Service Instance is associated with.
```
duglin: The text here wasn't instance specific before, it was more generic. For now this is probably ok but we need to be careful that if "context" is ever used on messages that are not necessarily related to instances then we'll need to revert this change.

spec.md
```diff
@@ -1112,7 +1112,7 @@ might not be able to provide this information at the time of the binding
 request, therefore the `bind_resource` and its fields are OPTIONAL.
 
 Below are some common fields that MAY be used. Platforms MAY choose to add
-additional ones as needed.
+additional ones as needed (see [conventions](profile.md#bind-resource-object)).
```
duglin: s/conventions/Bind Resource Object/

profile.md
```diff
@@ -232,6 +229,37 @@ part of a Kubernetes API call:
...
+In the [Open Service Broker API specification](spec.md), requests to
+[create a Service Binding](spec.md#binding) can contain a `bind_resource`
+object in which Platforms MAY choose to add additional fields.
+
+#### Cloud Foundry Bind Resource Object
+
+The following properties are defined for usage within a Cloud Foundry
+deployment:
+
+- `space_guid`
```
duglin: Is this required or optional for CF?  We should be clear.

mattmcneeney: optional - will make this clearer

spec.md
mattmcneeney: done

profile.md
```diff
@@ -232,6 +229,37 @@ part of a Kubernetes API call:
...
+  this property MUST be a non-empty string serialized as follows:
+  ```
+  "space_guid": "space-guid-here"
+  ```
+  For example:
+  ```
+  "space_guid": "15823972-c216-4ba5-9f3f-e75b0b891297"
+  ```
+
+The following example shows a `bind_resource` object that might appear as part
```
avade: When "might" it appear?

mattmcneeney: When a CF deployment supports it, and when a binding is being created to an application

avade: I would suggest you replace might, with the exact example / CF deployment version supported from.

Is there any way for a client to know when to depend on it from the API header?

mattmcneeney: Currently we just version this file, so broker authors will know it is supported when CF sends `2.14` as `x-broker-api-version`. That's probably enough right?

profile.md
```diff
@@ -16,8 +16,10 @@ or REQUIRED level requirements defined herein.
 - [Originating Identity Header](#originating-identity-header)
 - [Context Object](#context-object)
   - [Context Object Properties](#context-object-properties)
-  - [Cloud Foundry](#cloud-foundry)
-  - [Kubernetes](#kubernetes)
+  - [Cloud Foundry Context Object](#cloud-foundry-context-object)
+  - [Kubernetes Context Object](#kubernetes-context-object)
+- [Bind Resource Object](#bind-resource-object)
+  - [Cloud Foundry Bind Resource Object](#cloud-foundry-bind-resource-object)
```
duglin: I really like that you added the extra words ("Context Object", "Bind Resource Object") to these titles. Just makes it easier to track what section you're in 
👍

duglin: Having said that, something is a bit off with this TOC. In some cases we list level 4 stuff but not level 3 stuff.  And I think the platform specific context object stuff might be one level too deep.  If you don't want to fix those in this PR that's fine, I'll do it in a follow-on one- just thought I mention it before I forget.

mattmcneeney: Yeah I noticed that. I'll just fix it in here whilst it's open as has no LGTXs.

mattmcneeney: OK, PR updated. Now all level 2 and 3 titles are in the TOC.

profile.md
```diff
@@ -232,14 +233,45 @@ part of a Kubernetes API call:
...
+In the [Open Service Broker API specification](spec.md), requests to
+[create a Service Binding](spec.md#binding) can contain a `bind_resource`
+object in which Platforms MAY choose to add additional fields.
+
+### Cloud Foundry Bind Resource Object
+
+The following properties are defined for usage within a Cloud Foundry
+deployment:
+
+- `space_guid` (_optional_)
```
duglin: I think the use of `_optional_` here is new and I'd prefer to not confuse people so perhaps we should just say:
`This OPTIONAL property is the GUID of a ....`

mattmcneeney: Done!

duglin:
LGTM

avade:
We can address "might" in another PR / issue. 

LGTM

angarg12:
LGTM

duglin:
just one more needed

fmui:
LGTM

PR #416, opened 2018-01-19
Add Cloud Foundry specific service metadata

mattmcneeney:
Cloud Foundry now allows Service Instances to be shared across orgs and spaces. This change adds this to a new **Cloud Foundry Service Metadata** section of profile.md

### DIFF
--- a/profile.md
+++ b/profile.md
@@ -19,6 +19,7 @@ or REQUIRED level requirements defined herein.
   - [Cloud Foundry](#cloud-foundry)
   - [Kubernetes](#kubernetes)
 - [Service Metadata](#service-metadata)
+  - [Cloud Foundry Service Metadata](#cloud-foundry-service-metadata)
 
 ## Notations and Terminology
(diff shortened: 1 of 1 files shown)


duglin:
I suspect we may have more platform-specific extensions to add over time too.
LGTM

angarg12:
Would it make sense to move platform-specific info, such as metadata, profiles and example brokers, to the wiki so that we don't need a PR and approval cycle every time we need to change them?

mattmcneeney:
@angarg12 I guess it depends if we want to tie changes to platform-specific info/context with versions of the OSBAPI spec or not. Something to discuss on tomorrow's call.

duglin:
Versioning the extensions would be good, but I would think that we'd want the text to be normative too.

avade:
LGTM

fmui:
LGTM

angarg12:
LGTM

PR #425, opened 2018-01-22
Provide some guidance for CLI-friendly names

### LINKED ISSUE #305: service/plan names can be "-"
we should probably add something about how "-" isn't a good choice for a name :-)

duglin:
Closes #305

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -338,6 +338,11 @@ defines a "CLI-friendly" string as a short string that MUST only use lowercase
 characters, numbers and hyphens, with no spaces. This will make it easier for
 users when they have to type it as an argument on the command line.
 
+For backwards compatibility reasons, this specification does not preclude
+the use of CLI-unfriendly strings that might cause problems for command line
+parsers (or that are not very meaningful to users), such as `-`.
+It is therefore RECOMMENDED that implementations avoid such strings.
+
 | Response field | Type | Description |
 | --- | --- | --- |
 | services* | array-of-service-objects | Schema of service objects defined below. MAY be empty. |


angarg12:
LGTM

mattmcneeney:
LGTM

spec.md
```diff
@@ -338,6 +338,11 @@ defines a "CLI-friendly" string as a short string that MUST only use lowercase
 characters, numbers and hyphens, with no spaces. This will make it easier for
 users when they have to type it as an argument on the command line.
 
+For backwards compatibility reasons, this specification does not preclude
```
pmorie: Should this read "CLI-unfriendly"?

duglin: I think it can go either way, but "unfriendly" does has kind of a funny tweak to it so I'll switch it.

duglin:
@angarg12 @mattmcneeney re-review needed

angarg12:
LGTM

avade:
LGTM

duglin:
just one more needed - @vaikas-google ? @pmorie ?

duglin:
@fmui ?

fmui:
LGTM

vaikas-google:
LGTM

avade:
LGTM

dotchev:
> For backwards compatibility reasons, this specification does not preclude the use of CLI-unfriendly strings that might cause problems for command line parsers (or that are not very meaningful to users), such as -. It is therefore RECOMMENDED that implementations avoid such strings.

This seems contradicting with

> A CLI-friendly name of the service. MUST only contain alphanumeric characters, periods, and hyphens (no spaces).

The issue we have is that we have many services in CF that use some other characters like `_`. Now we cannot register these in Service Catalog in K8S.
```
Status:   ErrorSyncingCatalog - Error syncing catalog from ClusterServiceBroker.Error reconciling ClusterServiceClass (K8S: "367188a4-2511-4038-9be7-9694b05214dc" ExternalName: "simple_service") (broker "simple-broker"): ClusterServiceClass.servicecatalog.k8s.io "367188a4-2511-4038-9be7-9694b05214dc" is invalid: spec.externalName: Invalid value: "simple_service": [-.a-zA-Z0-9]+ (regex used for validation is 'service-name-40d-0983-1b89') @ 2018-09-14 10:16:34 +0000 UTC  
```

duglin:
I was going to open up an issue for this exact situation (the underscore) because we're seeing it too. Apparently CF only verifies that its a string so I'm not sure why the spec places any restrictions at all.

As for those 2 spots in the spec... they're not exactly contradictory. The 2nd one tries to constrain the set of valid characters, the 1st one tries to tell people to be smart about what words they use while using those characters. I think the example given is a good one.... while "-" is a valid character, its really odd (and not friendly) to use just "-" as the service name.

Issue: https://github.com/openservicebrokerapi/servicebroker/issues/603

PR #430, opened 2018-01-24
Terminology: Use "Platform" in place of undefined term "Controller"

mihnjong:
"Controller" is not defined in the spec. Using pre-defined "Platform" is better than defining a new term.


spec.md
```diff
@@ -359,7 +359,7 @@ It is therefore RECOMMENDED that implementations avoid such strings.
...
 | requires | array-of-strings | A list of permissions that the user would have to give the service, if they provision it. The only permissions currently supported are `syslog_drain`, `route_forwarding` and `volume_mount`. |
 | bindable* | boolean | Specifies whether Service Instances of the service can be bound to applications. This specifies the default for all plans of this service. Plans can override this field (see [Plan Object](#plan-object)). |
-| metadata | object | An opaque object of metadata for a Service Offering. Controller treats this as a blob. Note that there are [conventions](profile.md#service-metadata) in existing Service Brokers and controllers for fields that aid in the display of catalog data. |
+| metadata | object | An opaque object of metadata for a Service Offering. A Platform treats this as a blob. Note that there are [conventions](profile.md#service-metadata) in existing Service Brokers and the Platform for fields that aid in the display of catalog data. |
```
duglin: s/A Platform treats/It is expected that Platforms will treat/
s/the Platform/Platforms/

and in the other spot too

mihnjong: Hello Doug, thanks for the prompt review. One question before making a further change: what does it mean by "treat as a blob"? Do we mean that Platform can ignore `metadata` field and skip to parse it? What's the intention of letting readers know of "blob"?

mihnjong: Follow-up: "it is expected that Platforms will treat" can be shortned as "Platforms SHOULD treat" ?

duglin: Sort of.  "treat as a blob" just means that the Platform doesn't need to try to interpret the json - it is supposed to just pass it along to the user. I'm not sure we want to say SHOULD in there because I don't think we can really place any requirements on the Platform either way, nor can we verify it.

mihnjong: What I am trying to achieve is to rewriting your suggested sentence with [RFC2119](https://tools.ietf.org/html/rfc2119). Now, it looks that we can use "MAY". What about this?

"Platforms MAY parse the metadata object".

duglin: The problem is that I'm not sure we can use RFC2119 keywords in this case because what, if anything, the Platform does with this data is our of scope of the spec.  I'd prefer if we just focused on using "Platform" instead of "Controller".

duglin:
I think you need to rebase, this includes commits that are not part of this PR

spec.md
mihnjong: Fixed as you suggested. Thanks!

mihnjong:
Sorry for the mess. I think it's good for now.

duglin:
thanks! LGTM

angarg12:
LGTM

fmui:
LGTM

avade:
LGTM

Issue #114, opened 2016-12-07
Generic Extensions (Actions) Proposal


avade:
Following the discussion in https://github.com/openservicebrokerapi/servicebroker/issues/107 this issue aims to create a proposal for "Service Broker Actions".

---

## Problem 

*As a service broker author, I want to be able to optionally offer operations beyond "provision, update, bind, unbind, delete" so that I can provide new features to users of my service via the service broker API*

## Examples

- Backup and restore functionality (previous discussion https://github.com/openservicebrokerapi/servicebroker/issues/107)
- Schedule things via the service broker

## Workflows

Service clients can:

- Discover the actions available from a service broker
- Trigger a specific action synchronously
- Trigger a specific action asynchronously
- Pass input parameters to the action
- Discover and validate input parameters (input parameter schema)

Service authors can:

- Optionally define actions at service broker registration time

## History

https://github.com/openservicebrokerapi/servicebroker/issues/107 
https://docs.google.com/document/d/1Y5vwWjvaUIwHI76XU63cAS8xEOJvN69-cNoCQRqLPqU/edit#heading=h.kzfeyua6md59

## Next steps

1. [Proposal document]( https://docs.google.com/document/d/1Dhs4RrvXxyDcmK8LdxbkHcEb_HrpiA4XqVwaIy2sTXg)
1. Review on OSBAPI WG call

gberche-orange:
Here is a [google docs placeholder](https://docs.google.com/document/d/1Dhs4RrvXxyDcmK8LdxbkHcEb_HrpiA4XqVwaIy2sTXg/edit). I'll try fleshing some ideas there in the coming days. Anyone feel free to start earlier if they have some bandwidth.

@avade do you have pointers to the WG calendar and dialin/minutes details ?

avade:
@gberche-orange Perfect thanks! - @shalako can add you to the invite.

shalako:
Sent. Guillaume, consider that dial-in info will change for next week's call. I'll be sending an update to the invite shortly.

gberche-orange:
thanks @shalako 

I've proposed a first draft into https://docs.google.com/document/d/1Dhs4RrvXxyDcmK8LdxbkHcEb_HrpiA4XqVwaIy2sTXg/edit 

I've moved the sharing mode to comment mode so that we can systematically get notifications from suggestions. Let me know the list of people to share with edit role as to accept/reject suggestions collaboratively.

gberche-orange:
[...]

PR #431, opened 2018-01-26
Generic Extensions (actions)

rhodie27:
See issue https://github.com/openservicebrokerapi/servicebroker/issues/114.

First draft of the generic actions (aka extensions) proposal on a per-service-instance basis.

A couple notes: 

1. In addition to updating **Provisioning**, I added a separate section for **Extensions** at the bottom b/c I suspect we will add actions/extensions at the service (and broker) level next. A lot of this can likely be handled in the same way so it made sense to me to separate it out now.
2. The OpenAPI, Auth & adheres_to sections obviously need more content/wording, but I wanted to get some quick feedback on the direction.
3. I haven't really looked at any changes required outside of spec.md yet.
4. I'm an awful speller today, so good luck.


rhodie27:
Reopen please, my org was private for some reason. I fixed the CI issues.

spec.md
```diff
@@ -841,11 +842,30 @@ For success responses, the following fields are defined:
...
 | operation | string | For asynchronous responses, Service Brokers MAY return an identifier representing the operation. The value of this field MUST be provided by the Platform with requests to the [Last Operation](#polling-last-operation) endpoint in a percent-encoded query parameter. If present, MUST be a non-empty string. |
+| extension_apis | array-of-objects | For extensions to the Service Broker API on a per Service Instance basis, Service Brokers MAY return one or more objects that describe additional API endpoints via an OpenAPI document. See [Extension APIs Object](#extension-apis-object) for more information.|
+
+##### Extension APIs Object
+
+The `extension_apis` object MAY be used to describe any additional endpoint needed related to a Service Instance. An examples of this might include backup and restore endpoints for a database. If present, MUST return a `discovery_url`. See [Extensions](#extensions) for more information.
```
duglin: Can you wrap non-table lines at 80 columns to be consistent with the rest of the doc?

duglin: if present, each item in the `extension_apis` array MUST include a `discovery_url` property.

spec.md
```diff
@@ -841,11 +842,30 @@ For success responses, the following fields are defined:
...
+
+| Response field | Type | Description |
+| --- | --- | --- |
+| discovery_url | string | A URI pointing to a valid OpenAPI 3.0+ document describing the API extension(s) to the Service Instance including server location, endpoints, parameters and any other detail required for invocation. The location of the API extension endpoint(s) can be local to the Service Broker or on a remote server. The Platform MUST execute these API extensions on behalf of the client. MUST be a valid URI. See [OpenAPI](#openapi-document) for more information.|
```
duglin: s/required for invocation/needed for invocation/

duglin: Should probably use the '*' pattern to indicate that its a required field.

spec.md
duglin: Although, I think this MUST is redundant with the REQUIRED aspect of the discovery_url you mention in the table below. So it might be best to remove here so we're not duplicating requirements.

spec.md
duglin: Rather than "The Platform MUST execute...", which sounds like the platform is required to invoke these even if it has no need to, I would suggest that we remove that sentence from here. Instead, since this is really more of a generic statement about the entire concept and not about this one field, let's add a sentence to the intro section at line 849 about this issue.  And perhaps there say something like:

These APIs are extensions to the Open Service Broker API. As such they are intended to be invoked by the Platform on behalf of its clients.

I tried to think of a way to make it normative but it felt a bit odd. I don't think we ever say that the normal OSBAPIs MUST be invoked by the Platform, although its implied. So I think we can probably live with the same non-normative-ness in these too.  WDYT?

spec.md
```diff
@@ -841,11 +842,30 @@ For success responses, the following fields are defined:
...
+| credentials | object | A Service Broker MAY return authentication details for running any of the extension API calls, especially for those running on remote servers. If not present, platforms can assume that standard broker authentication will work for the new endpoint(s). See [Extension Authentication](#extension-api-authentication) for more information.|
```
duglin: If not present, the same authentication mechanism used for the normal Open Service Broker APIs MUST work for the new endpoint(s).

spec.md
```diff
@@ -841,11 +842,30 @@ For success responses, the following fields are defined:
...
+| adheres_to | string | A URI pointing to a specification detailing the implementation guidelines for the OpenAPI document hosted at the `discovery_url`. If present, must be a valid URI.|
```
duglin: I wonder if instead of "pointing" (which sort of implies there has to be a server listening at the endpoint), we should say "referring". Almost the same but perhaps wishy-washy enough to then allow us to add:  While this property is a URI, there is no requirement for there to be an actual server listening at that endpoint. This value is meant to provide a unique identifier representing the set of extensions APIs supported.

spec.md
```diff
@@ -841,11 +842,30 @@ For success responses, the following fields are defined:
...
 ```
 {
   "dashboard_url": "http://example-dashboard.example.com/9189kdfsk0vfnku",
-  "operation": "task_10"
+  "operation": "task_10",
+  "extension_apis":[{
+      "discovery_url": "http://example-openapi-doc.example.com/extensions",
+      "credentials":[{
+        "tokenURL": "https://example.com/api/oauth/token"
+      }]
```
duglin: add comma

spec.md
```diff
@@ -1422,3 +1442,21 @@ If the Platform encounters an internal error provisioning a Service Instance or
...
+
+### OpenAPI Document
+
+To extend the API, the Service Broker MUST create an OpenAPI 3.0+ document describing the additional API endpoints and parameters. The returned OpenAPI document MUST be in json format. See the [OpenAPI Specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md) for more information.
+
+### Extension API on Remote Servers
+
+Additional API endpoints MAY be executed on a remote server, however the OpenAPI document MUST include the servers `url` so that the Platform will know how to invoke the endpoint(s). If the server `context` parameter of the OpenAPI document is set to local host the Platform can assume the extension API endpoints are to be invoked using the Service Broker host and port.
```
duglin: s/can assume.../MUST use.... 

I assume that its just the host (localhost) that is swapped out in favor of the original host/port, right? meaning any path stuff in the OpenAPI doc should still be used right?  We should probably say that if so.  You may want to also say that any port number specified will be replace too if "localhost" is used - if that's your intent.

spec.md
```diff
@@ -1422,3 +1442,21 @@ If the Platform encounters an internal error provisioning a Service Instance or
...
+
+### Extension API Authentication
+
+The Service Broker MAY use the the same broker authentication method for invocation of extension endpoints by not providing credentials as part of a `extensions_api` object. If no credentials are present in an `extensions_api` object, the Platform can assume that the default broker authentication is required.
```
duglin: This is a dup from what's above - I would suggest that we only say it once - and I think the one from above is good enough.  You may be able to just move all of this section up above - not sure its worthy enough for just one para. Actually, would it read better to just move this entire "Extension" section into what we have above since it kind of feels like a repeat of the same material and anything new could just be sprinkled up there.  Just a thought.

duglin:
@rhodie27 overall I think it looks good!  Just some editorial suggestions.

spec.md
rhodie27: Yeah, I wasn't sure. Like I wrote in PR, while writing this I separated it out b/c a lot of it seemed like it would be redundant if we added extensions on other resources down the line. But, obviously, we could always abstract it out at that point in time.

duglin:
@rhodie27 any chance of addressing the comments before tomorrow's call?

spec.md
rhodie27: I think I got these.

spec.md
rhodie27: I think it makes sense. I changed it.

spec.md
rhodie27: Done.

spec.md
rhodie27: Done. Wishy-washy it is!

spec.md
rhodie27: Yup. Got it.

spec.md
rhodie27: I did it your way, for now. If others chime in and want it separated out I can always revert.

spec.md
rhodie27: I'm talking specifically about the server open, https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#serverObject, url. It can or cannot include a port. I'm think we want to say if it just says http://localhost that signifies that it should run on the SB host/port. However, there is an issue if the OAPI doc provides more than one url. Perhaps we can say it provides a url with a description "Service Broker host".

I need to think about this more, so I didn't modify it. Can get to it tomorrow.

spec.md
```diff
@@ -841,11 +841,45 @@ For success responses, the following fields are defined:
...
+The Service Instance provisioning request will return a URI to an OpenAPI 3.0+
+document that the Platform can use to determine the new endpoint(s),
+parameter(s), authentication mechanism and server URL. The new APIs are 
+extensions to the Open Service Broker API. As such they are indended to be 
+invokved by the Platform on behalf of its clients.
+
+Extension API endpoints MAY be executed on a remote server, however the OpenAPI
+document MUST include the servers `url` so that the Platform will know how to
+invoke the endpoint(s). If the server `context` parameter of the OpenAPI
+document is set to localhost the Platform can assume the extension API endpoints
```
duglin: may want to backtick `localhost`

duglin: should we s/can assume/MUST assume/ ?

spec.md
```diff
@@ -841,11 +841,45 @@ For success responses, the following fields are defined:
...
+are to be invoked using the Service Broker host and port.
+
+| Response field | Type | Description |
+| --- | --- | --- |
+| discovery_url | string | A URI pointing to a valid OpenAPI 3.0+ document describing the API extension(s) to the Service Instance including server location, endpoints, parameters and any other detail the platform needs for invocation. The location of the API extension endpoint(s) can be local to the Service Broker or on a remote server. MUST be a valid URI. The returned OpenAPI document MUST be in json format. See the [OpenAPI Specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md) for more information. |
```
duglin: I think you need to add a `*` and the `* Fields with an asterisk are REQUIRED.` sentence after the description of the fields.

spec.md
```diff
@@ -841,11 +841,45 @@ For success responses, the following fields are defined:
...
+| credentials | object | A Service Broker MAY return authentication details for running any of the extension API calls, especially for those running on remote servers. If not present, the same authentication mechanism used for the normal Open Service Broker APIs MUST work for the new endpoint(s). If the Service Broker wants to use alternate methods of authentication, (e.g. on remote servers) it MUST provide details to that mechanism in the OpenAPI document, (e.g. an OAuth Flow Object), and the appropriate credential(s), (e.g. bearer token), as part of the `extensions_api` object. If credentials are present in an `extensions_api` object, the Platform will need to verify the authentication method from the OpenAPI document. |
```
duglin: s/as part of the `extensions_api` object/within this `credentials` field/  I think

rhodie27: field? object? I put field for now.

spec.md
rhodie27: Yup, I thought I did. Must've reverted. There now.

spec.md
rhodie27: Done. I think that makes sense. I can't think of a reason they wouldn't want to assume that.

duglin:
thanks!  Looks good!

spec.md
```diff
@@ -841,11 +841,46 @@ For success responses, the following fields are defined:
...
+extensions to the Open Service Broker API. As such they are indended to be 
+invokved by the Platform on behalf of its clients.
+
+Extension API endpoints MAY be executed on a remote server, however the OpenAPI
+document MUST include the servers `url` so that the Platform will know how to
+invoke the endpoint(s). If the server `context` parameter of the OpenAPI
+document is set to `localhost` the Platform MUST assume the extension API endpoints
+are to be invoked using the Service Broker host and port.
+
+| Response field | Type | Description |
```
duglin: s/field/Field/

spec.md
```diff
@@ -841,11 +841,47 @@ For success responses, the following fields are defined:
...
+| extension_apis | array-of-objects | For extensions to the Service Broker API on a per Service Instance basis, Service Brokers MAY return one or more objects that describe additional API endpoints via an OpenAPI document. See [Extension APIs Object](#extension-apis-object) for more information. |
+
+##### Extension APIs Object
+
+The `extension_apis` object MAY be used to describe any additional endpoint
+to the Open Service Broker API. An example of this could be lifecycle
+management of a Service Instance, (e.g. "Day Two Operations"), like Backup,
+Restore, Stop, Start, Restart and Pause.
+
+The Service Instance provisioning request will return a URI to an OpenAPI 3.0+
```
duglin: perhaps s/request will return/response will include/  since its not the request that contains the data.

spec.md
```diff
@@ -841,11 +841,47 @@ For success responses, the following fields are defined:
...
+Extension API endpoints MAY be executed on a remote server, however the OpenAPI
+document MUST include a Server Object with an accurate `url` parameter and a
+`description` field labeled, "Service Broker Extensions Server". If no Server
+Object `description` field contains, "Service Broker Extensions Server", the 
+Platform MUST assume the extension API endpoints are to be invoked using the
+Service Broker host and port. See [OpenAPI Server Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#serverObject) for more information.
+
+| Response Field | Type | Description |
+| --- | --- | --- |
+| discovery_url* | string | A URI pointing to a valid OpenAPI 3.0+ document describing the API extension(s) to the Open Service Broker API including server location, endpoints, parameters and any other detail the platform needs for invocation. The location of the API extension endpoint(s) can be local to the Service Broker or on a remote server. MUST be a valid URI. The returned OpenAPI document MUST be in json format. See the [OpenAPI Specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md) for more information. |
```
duglin: for the dashboardURL we say this "The URL MUST contain enough information for the dashboard to identify the resource being accessed". I think we need something similar here, and possible about the endpoints returned within the swagger doc, because it doesn't really make sense for two instances to be given the same set of URLs - otherwise how will the server know which instance is being operated on.

spec.md
vaikas-google: To me this reads that there's only one object returned: I was thinking that there is an object:
extension_api {
  discovery_url string
  credentials credType
  adheres_to string
}
and:
extension_apis []extension_api

So, there's the struct that rerepresents a single discovery_url (with creds) and there can be an array of them. Somehow reading this however, I think there's only one.

vaikas-google:
LGTM

rhodie27:
Check out https://github.com/rhodie27/osb-actions-poc for this in action, (pun intended). Note, the verbs are slightly different and is lacking credentials and adheres_to.

spec.md
```diff
@@ -841,11 +841,61 @@ For success responses, the following fields are defined:
...
+    }
+  ]
+}
+```
+
+See [OpenAPI Server Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#serverObject) for more information.
+
+| Response Field | Type | Description |
+| --- | --- | --- |
+| discovery_url* | string | A URI pointing to a valid OpenAPI 3.0+ document describing the API extension(s) to the Open Service Broker API including server location, endpoints, parameters and any other detail the platform needs for invocation. The location of the API extension endpoint(s) can be local to the Service Broker or on a remote server. MUST be a valid URI. The returned OpenAPI document MUST be in json format. See the [OpenAPI Specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md) for more information. |
```
duglin: For `dashboard_url` we include:
> The URL MUST contain enough information for the dashboard to identify the resource being accessed

I think we need something in this PR that allows tells people that something needs to be unique in all of the URLs/data being passed around, otherwise how does the generic action endpoint know which instance is being operated on.

spec.md
```diff
@@ -841,11 +841,61 @@ For success responses, the following fields are defined:
...
+clients.
+
+Extension API endpoints MAY be executed on a remote server, however the OpenAPI
+document MUST include a Server Object with an accurate `url` parameter and a
+`description` field labeled, "Service Broker Extensions Server". If no Server
+Object `description` field contains, "Service Broker Extensions Server", the 
+Platform MUST assume the extension API endpoints are to be invoked using the
+Service Broker host and port. The `url` for this Server Object MUST be an
+absolute URL.
+
```
fmui: I don't get this. If the description field is not "Service Broker Extensions Server", the platform uses the brokers host and port. Why need the URL be absolute in this case? Isn't it relative to the broker? What if the broker host/port don't match the URL?
It can be difficult for a broker to determine the right host name in this context. It would be easier for a broker to return the (absolute) path here.

rhodie27: You're 100% right. I think I addressed this in the latest change.

spec.md
rhodie27: I don't agree that we MUST do this, but I will bow to the group decision. An example I can think of is you want to return the same swagger for every service instance and pass the service instance details via a body parameter, (as defined by the swagger).

duglin: We can do that, IF we also define how someone can programmatically know which parameter in the swagger represents the instance ID. W/o that I wouldn't know how to write code to do it.

So, note I said "URL/data" - meaning I'm ok with a statement about instance-unique URLs, or other text talking about how to know which param to set - either is fine - we just need to be clear about how to do it.

rhodie27: I think the latest address your concerns.

spec.md
```diff
@@ -841,11 +841,46 @@ For success responses, the following fields are defined:
...
+Platform can use to determine the new endpoint(s), parameter(s) and
+authentication mechanism. The new APIs are extensions to the Open Service Broker
+API. As such they are indended to be invokved by the Platform on behalf of its
+clients.
+
+See [OpenAPI Server Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#serverObject) for more information.
+
+| Response Field | Type | Description |
+| --- | --- | --- |
+| discovery_url* | string | A URI pointing to a valid OpenAPI 3.0+ document describing the API extension(s) to the Open Service Broker API including server location, endpoints, parameters and any other detail the platform needs for invocation. The location of the API extension endpoint(s) can be local to the Service Broker or on a remote server. MUST be a valid URI. The returned OpenAPI document MUST be in json format. See the [OpenAPI Specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md) for more information. |
```
fmui: In case of a local URI, I assume this is a path. Is this path relative to the broker host oder relative to the URL the broker was registered with?
For example, the broker has been register with `https://mybroker/broker1` and the path is `/backup`. Is the final URL `https://mybroker/broker1/backup` or `https://mybroker/backup`?

spec.md
duglin: @rhodie27 thanks - need to think more about it, at at first glace yes I think it does.

if we have an aggregation service broker in the path between the platform and the real service broker, I'm assuming the aggregation SB might need to twiddle relative URLs provided so they're not relative to the aggregation SB but point to the real one. Is this something we should say in here or is it obvious and we can leave that as an exercise for the dev?

spec.md
rhodie27: Good question, I hadn't thought of that. I would assume the former b/c of /broker1 is the path you are executing the provisioning from. I could add something like, "If discovery_url is a path, the Platform can assume it is to be called from the basepath of the Service Broker."

spec.md
rhodie27: I'd lean towards "up to the dev". I could see the aggregation SB actually hosting the openapi doc itself instead of passing the request through. And it could have any combination of local and remote endpoints.

spec.md
fmui: Sounds good. I guess we need something similar also for the `server_url` field.

spec.md
```diff
@@ -841,11 +841,46 @@ For success responses, the following fields are defined:
...
+Platform can use to determine the new endpoint(s), parameter(s) and
+authentication mechanism. The new APIs are extensions to the Open Service Broker
+API. As such they are indended to be invokved by the Platform on behalf of its
+clients.
+
+See [OpenAPI Server Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#serverObject) for more information.
+
+| Response Field | Type | Description |
+| --- | --- | --- |
+| discovery_url* | string | A URI pointing to a valid OpenAPI 3.0+ document describing the API extension(s) to the Open Service Broker API including server location, endpoints, parameters and any other detail the platform needs for invocation. The location of the API extension endpoint(s) can be local to the Service Broker or on a remote server. MUST be a valid URI. If `discovery_url` is a path, the Platform can assume it is to be called from the basepath of the Service Broker. The returned OpenAPI document MUST be in json format. See the [OpenAPI Specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md) for more information. |
```
duglin: s/from the basepath/relative to the basepath/

spec.md
```diff
@@ -841,11 +841,46 @@ For success responses, the following fields are defined:
...
+| server_url | string | A URI pointing to a remote server where API extensions will run. This URI will be used as the basepath for the endpoint(s) described by the `discovery_url` OpenAPI document. If no `server_url` is present, the Platform MUST assume the extension API endpoint(s) are to be invoked using the Service Broker host and port. If present, MUST be a valid URI. If `server_url` is a path, the Platform can assume it is to be called from the basepath of the Service Broker. |
```
duglin: s/from the basepath/relative to the basepath/

duglin: The text here says that if server_url is absent then the broker's host/port MUST be used, but if the OAPI doc has full URLs is that really true? Or can the OAPI doc only have relative paths?

spec.md
```diff
@@ -841,11 +841,46 @@ For success responses, the following fields are defined:
...
+| credentials | object | A Service Broker MAY return authentication details for running any of the extension API calls, especially for those running on remote servers. If not present, the same authentication mechanism used for the normal Open Service Broker APIs MUST work for the new endpoint(s). If the Service Broker wants to use alternate methods of authentication, (e.g. on remote servers) it MUST provide details to that mechanism in the OpenAPI document, (e.g. an OAuth Flow Object), and the appropriate credential(s), (e.g. bearer token), as part of the `extension_api` object within the `credentials` field. If credentials are present in an `extension_api` object, the Platform will need to verify the authentication method from the OpenAPI document. |
```
duglin: Do we need to document the sub-fields of this to make sure the same field names are used by all implementations?

duglin: I don't think this is a blocking comment for moving to validate-thru-impl - but if people agree, then we'll need it before merging.

duglin:
overall it looks good and I'm ok with moving to validate-thru-impl with the understanding that we'll do minor updates before we merge - I don't think any of my comments should block forward motion.

spec.md
rhodie27: This approach is stating that we are ignoring any server object from the OAPI doc, and only paying attention to the path objects, (which are relative). If you have 'server_url' defined, you do server_url/+path_object1-x. If you don't have 'server_url' defined, you do servicebrokerhost:port/+path_object1-x.

spec.md
rhodie27: Doug, are you talking about sub-fields of the credentials object?

duglin: yes

spec.md
duglin: Programmatically I see how that works. From an OAPI perspective, is it normal to ignore the server object? Would we need to say something here about how Platforms will ignore server objects? Or should they reject OAPI docs that have server objects?

rhodie27: In my experience, it's common practice to ignore the server object. You can also have more than one server object, so how do we decide which one to use? See, https://github.com/openservicebrokerapi/servicebroker/blob/master/openapi.yaml, we have two there and they are both basically examples. You typically always have a 'localhost' for testing during local development. I don't think there's a reason to explain this.

spec.md
rhodie27: We can't restrict them. The field names could be anything based on the openapi spec. The platform needs to read the spec; determine the auth type + variables, and then match those using the credentials object.

duglin: oh, I didn't realize this field had to conform to a definition in the OAPI spec - should we say that and add a ref to where in the spec they should look? This wasn't obvious to me since I didn't assume that something outside of an OAPI doc would be controlled by the OAPI spec.

spec.md
duglin: I can't tell you how much that hurts my head  :-)

spec.md
rhodie27: I thought that's what it says. :) Let's talk about how to reword this to make that clear.

spec.md
rhodie27: mission accomplished.

spec.md
```diff
@@ -841,11 +841,46 @@ For success responses, the following fields are defined:
...
+
+The `extension_api` object MAY be used to describe any additional endpoint
+to the Open Service Broker API. An example of this could be lifecycle
+management of a Service Instance, (e.g. "Day Two Operations"), like Backup,
+Restore, Stop, Start, Restart and Pause.
+
+The `extension_api` MUST include a URI to an OpenAPI 3.0+ document that the
+Platform can use to determine the new endpoint(s), parameter(s) and
+authentication mechanism. The new APIs are extensions to the Open Service Broker
+API. As such they are indended to be invokved by the Platform on behalf of its
```
zrob: indended -> intended

zrob: invokved -> invoked

pmorie: typo: intended

spec.md
```diff
@@ -841,11 +841,46 @@ For success responses, the following fields are defined:
...
+| alt_instance_id | string | Refers to a parameter in the `discovery_url` OpenAPI document that maps directly to the Service Broker API `instance_id` parameter. If the extension API endpoint(s) use a different value to represent a Service Broker instance, then `alt_instance_id` MUST be present. If not present, the Platform can assume `instance_id` means Service Broker Instance in the OpenAPI document. |
+| adheres_to | string | A URI refering to a specification detailing the implementation guidelines for the OpenAPI document hosted at the `discovery_url`. While this property is a URI, there is no requirement for there to be an actual server listening at that endpoint. This value is meant to provide a unique identifier representing the set of extensions APIs supported. If present, MUST be a valid URI. |
```
pmorie: The semantics of this column could be clarified a little bit - what is the advantage of specifying a URI that points to a location where no server is running?

spec.md
```diff
@@ -841,11 +841,46 @@ For success responses, the following fields are defined:
...
+Platform can use to determine the new endpoint(s), parameter(s) and
+authentication mechanism. The new APIs are extensions to the Open Service Broker
+API. As such they are indended to be invokved by the Platform on behalf of its
+clients.
+
+See [OpenAPI Server Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#serverObject) for more information.
+
+| Response Field | Type | Description |
+| --- | --- | --- |
+| discovery_url* | string | A URI pointing to a valid OpenAPI 3.0+ document describing the API extension(s) to the Open Service Broker API including server location, endpoints, parameters and any other detail the platform needs for invocation. The location of the API extension endpoint(s) can be local to the Service Broker or on a remote server. MUST be a valid URI. If `discovery_url` is a path, the Platform can assume it is to be called relative to the basepath of the Service Broker. The returned OpenAPI document MUST be in json format. See the [OpenAPI Specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md) for more information. |
```
pmorie: Should this object have a name? how about an ID?

pmorie: We should also make clear:

- the mutability characteristics of this object
- what expected platform actions are in response to the presence of this object

spec.md
```diff
@@ -841,11 +841,46 @@ For success responses, the following fields are defined:
...
+| server_url | string | A URI pointing to a remote server where API extensions will run. This URI will be used as the basepath for the endpoint(s) described by the `discovery_url` OpenAPI document. If no `server_url` is present, the Platform MUST assume the extension API endpoint(s) are to be invoked using the Service Broker host and port. If present, MUST be a valid URI. If `server_url` is a path, the Platform can assume it is to be called relative to the basepath of the Service Broker. |
```
pmorie: One thing that's been raised before is having a list of the operation names present in this object to make it easier to create UIs without hitting the URI.

spec.md
zrob: from call:   what credentials does the platform use to access this url?

spec.md
```diff
@@ -841,11 +841,46 @@ For success responses, the following fields are defined:
...
+| credentials | object | A Service Broker MAY return authentication details for running any of the extension API calls, especially for those running on remote servers. If not present, the same authentication mechanism used for the normal Open Service Broker APIs MUST work for the new endpoint(s). If the Service Broker wants to use alternate methods of authentication, (e.g. on remote servers) it MUST provide details to that mechanism in the OpenAPI document, (e.g. an OAuth Flow Object), and the appropriate credential(s), (e.g. bearer token), as part of the `extension_api` object within the `credentials` field. If credentials are present in an `extension_api` object, the Platform will need to verify the authentication method from the OpenAPI document. |
```
zrob: from call:   try to clear up that the workflow is this
1) the Spec returned from discovery_url will include an auth mechanism spec
2) the value of credentials includes the actual credential values to fill into the necessary variables indicated in the auth mechanism spec

zrob: could we get an example of an auth spec and a matching credentials object?

rhodie27:
Generic Actions Proposal:
https://docs.google.com/document/d/1fVBzqi6OK-lyfjDKADSk35hZD4YUqM2AMtoLOgrDDCE/edit

Generic Actions Usecases:
https://docs.google.com/document/d/1fVBzqi6OK-lyfjDKADSk35hZD4YUqM2AMtoLOgrDDCE/edit#heading=h.b62bp6xehn25

spec.md
rhodie27: got it. thanks.

rhodie27: got it. thanks.

rhodie27: got it. thanks.

spec.md
rhodie27: Sorry, I don't follow. Which part are you referencing?

spec.md
rhodie27: Remote docs SHOULD be no auth. I added that in.

spec.md
rhodie27: I refactored the credentials and added some examples. Hopefully it doesn't make it worse. :)

spec.md
rhodie27: The exention_api object is defined and described above... or I'm not following your comment.

Will think about mutability & platform responses.

spec.md
rhodie27: I added a new object to try to cover this.

n3wscott:
LGTM

spec.md
```diff
@@ -841,13 +841,113 @@ For success responses, the following fields are defined:
...
+| alt_instance_id | string | Refers to a parameter in the `discovery_url` OpenAPI document that maps directly to the Service Broker API `instance_id` parameter. If the extension API endpoint(s) use a different value to represent a Service Broker instance, then `alt_instance_id` MUST be present. If not present, the Platform can assume `instance_id` means Service Broker Instance in the OpenAPI document. |
```
mattmcneeney: Are you investigating an alternative key name @rhodie27?

spec.md
```diff
@@ -841,13 +841,113 @@ For success responses, the following fields are defined:
...
+| operation_display_names | object | A Service Broker MAY match operations in the `discovery_url` OpenAPI document with user consumable display names. If present, each key in this object MUST match an [operationID](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#operationObject) from the `discovery_url` OpenAPI document, with each value representing a human readable string. Platforms can use this object to show end users operations that are available on their Service Instance prior to calling the `discovery_url` OpenAPI document. |
+| adheres_to | string | A URI refering to a specification detailing the implementation guidelines for the OpenAPI document hosted at the `discovery_url`. While this property is a URI, there is no requirement for there to be an actual server listening at that endpoint. This value is meant to provide a unique identifier representing the set of extensions APIs supported. If present, MUST be a valid URI. |
```
mattmcneeney: `While this property is a URI, there is no requirement for there to be an actual server listening at that endpoint.` and `If present, MUST be a valid URI.` confuse me a little. If it's not required to be a server, why do we enforce it to be a URI?

Am I write in thinking that this field is going to be used to allow clients to spot "common" operations that they could do, say, in bulk across multiple service instances?

duglin: Often people push for unique IDs to be URIs so that they avoid strings that too simplistic - like "backup". By making it a URI its trying to force people to namespace things by including some domain name in the URI.

While I've heard bulk ops as one reason for this, to me the bigger reason is that a platform (or its UI) can know when it comes across some actions that it knows about and may want to do some special processing for. E.g. perhaps it has a dedicated UI for them.  W/o some well defined URI to say "yes, these are the set of ops that's been defined by spec xxx" the platform can't really be sure what those ops do. Checking (strcmp) the names of the ops alone doesn't really guarantee anything.

spec.md
rhodie27: Yes, but none of them are great. Open to other suggestions.

* openapi_instance_id
* id
* openapi_id

mattmcneeney: More brainstorming:
* `openapi_instance_key`
* `extensions_instance_key`
* `discovery_instance_key`

I don't like `id` as that makes me think of the value of the key-value pair, rather than the key if that makes sense.

spec.md
rhodie27: As per Paul's suggestion I will add some text here about what the Platform can do with this information. Should clear this up, (mildly).

spec.md
```diff
@@ -841,13 +841,113 @@ For success responses, the following fields are defined:
 | --- | --- | --- |
 | dashboard_url | string | The URL of a web-based management user interface for the Service Instance; we refer to this as a service dashboard. The URL MUST contain enough information for the dashboard to identify the resource being accessed (`9189kdfsk0vfnku` in the example below). Note: a Service Broker that wishes to return `dashboard_url` for a Service Instance MUST return it with the initial response to the provision request, even if the service is provisioned asynchronously. If present, MUST be a non-empty string. |
 | operation | string | For asynchronous responses, Service Brokers MAY return an identifier representing the operation. The value of this field MUST be provided by the Platform with requests to the [Last Operation](#polling-last-operation) endpoint in a percent-encoded query parameter. If present, MUST be a non-empty string. |
+| extension_apis | array-of-objects | For extensions to the Service Broker API, Service Brokers MAY return one or more `extension_api` objects that describe additional API endpoints via an OpenAPI document. See [Extension API Object](#extension-api-object) for more information. |
```
duglin: If https://github.com/openservicebrokerapi/servicebroker/pull/459/files is adopted then the 2nd column will need to change.

spec.md
```diff
@@ -841,13 +841,113 @@ For success responses, the following fields are defined:
...
+authentication mechanism. The new APIs are extensions to the Open Service Broker
+API. As such they are intended to be invoked by the Platform on behalf of its
+clients.
+
+See [OpenAPI Server Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#serverObject) for more information.
+
+| Response Field | Type | Description |
+| --- | --- | --- |
+| discovery_url* | string | A URI pointing to a valid OpenAPI 3.0+ document describing the API extension(s) to the Open Service Broker API including, endpoints, parameters, authentication mechanism and any other detail the platform needs for invocation. The location of the API extension endpoint(s) can be local to the Service Broker or on a remote server. If located on a remote server acccessing the OpenAPI document MUST NOT require authentication. MUST be a valid URI. If `discovery_url` is a path, the Platform can assume it is to be called relative to the basepath of the Service Broker. The returned OpenAPI document MUST be in json format. See the [OpenAPI Specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md) for more information. |
```
duglin: I think you're implying that if the location is not on a remote server then the OpenAPI doc MUST use the same auth as normal OSBAPI calls, correct? If so, we should probably state that.

spec.md
```diff
@@ -841,13 +841,113 @@ For success responses, the following fields are defined:
...
+| server_url | string | A URI pointing to a remote server where API extensions will run. This URI will be used as the basepath for the endpoint(s) described by the `discovery_url` OpenAPI document. If no `server_url` is present, the Platform MUST assume the extension API endpoint(s) are to be invoked using the Service Broker host and port. If present, MUST be a valid URI. If `server_url` is a path, the Platform can assume it is to be called relative to the basepath of the Service Broker. |
```
duglin: Does this mean that in the OpenAPI doc they can't put the full URL for each operation? If not, perhaps add to the end of the 2nd sentence " when those endpoints are relative URLs".   maybe?

spec.md
```diff
@@ -841,13 +841,113 @@ For success responses, the following fields are defined:
...
+| alt_instance_id | string | Refers to a parameter in the `discovery_url` OpenAPI document that maps directly to the Service Broker API `instance_id` parameter. If the extension API endpoint(s) use a different value to represent a Service Broker instance, then `alt_instance_id` MUST be present. If not present, the Platform can assume `instance_id` means Service Broker Instance in the OpenAPI document. |
+| operation_display_names | object | A Service Broker MAY match operations in the `discovery_url` OpenAPI document with user consumable display names. If present, each key in this object MUST match an [operationID](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#operationObject) from the `discovery_url` OpenAPI document, with each value representing a human readable string. Platforms can use this object to show end users operations that are available on their Service Instance prior to calling the `discovery_url` OpenAPI document. |
```
duglin: s/calling/retrieving/

duglin: How does this play with the idea of more than one broker supporting the same set of APIs? In particular, what if each provided their own set of display names and they're all different? As long as the platform never tries to do any kind of bulk action I guess that's ok - but even so, seeing a different phrase for the same semantics might give a funky UX.

While I understand the desire for this, it sure would be nice if we could say something like "use the last phrase in the operation's URL/path".

spec.md
jmrodri: I agree with @mattmcneeney on this, something other than id would fit better. Any of the above *_key variants seems acceptable. I think I like `openapi_instance_key` or `extensions_instance_key` the best.

jmrodri:
Nice work overall.

spec.md
spadgett: Can `operation_display_names` be a subset of the list of operations in the OpenAPI document? And if so, should the platform only show the actions in this list, or all?

spec.md
```diff
@@ -841,13 +841,113 @@ For success responses, the following fields are defined:
...
 | operation | string | For asynchronous responses, Service Brokers MAY return an identifier representing the operation. The value of this field MUST be provided by the Platform with requests to the [Last Operation](#polling-last-operation) endpoint in a percent-encoded query parameter. If present, MUST be a non-empty string. |
+| extension_apis | array-of-objects | For extensions to the Service Broker API, Service Brokers MAY return one or more `extension_api` objects that describe additional API endpoints via an OpenAPI document. See [Extension API Object](#extension-api-object) for more information. |
+
+##### Extension API Object
+
+The `extension_api` object MAY be used to describe any additional endpoint
+to the Open Service Broker API. An example of this could be lifecycle
+management of a Service Instance, (e.g. "Day Two Operations"), like Backup,
+Restore, Stop, Start, Restart and Pause.
```
spadgett: Some of these example operations seem like they would be long-running. Is there any way for a broker to report status of a long running action back to the platform?

spadgett: Related to this, is there a way brokers can provide either success or failure messages for the platform to display? Ideally if the request fails, the UI can give a specific error about what went wrong instead of a generic failed error.

spec.md
```diff
@@ -841,13 +841,113 @@ For success responses, the following fields are defined:
...
+
+The `extension_api` MUST include a URI to an OpenAPI 3.0+ document that the
+Platform can use to determine the new endpoint(s), parameter(s) and
+authentication mechanism. The new APIs are extensions to the Open Service Broker
+API. As such they are intended to be invoked by the Platform on behalf of its
+clients.
```
pmorie: One thing that occurred to me reading back through this is that it's not clear how the extensions are expected be invoked. Does the platform invoke them? Do users invoke them? This is probably entangled with @spadgett's question about long-running operations.

spec.md
rhodie27: Async ops, if enabled, would be defined via the swagger file. The platform could detect and subscribe to ops based on the returned swagger.

spec.md
rhodie27: Correct, the openapi endpoints are relative to one or more defined server objects. Will likely change to "path objects" which implies they are relative paths.

spec.md
rhodie27: I think there are multiple implementation paths for platforms to accept the extensions api:

1. Handle everything dynamically; use 'operation_display_names' to show operations before generating UI forms from the openapi doc.
2. Have a platform specific definition for how operations and openapi documents are written. The platform can create pre-generated UIs based on known operation names and parameters.
3. Utilizes adheres_to to pre-generate UIs on a case by case basis according to some standard(s).
4. All of the above.

IMO, it's pick your poison and 'operation_display_names' gives some UI advantage for #1 in that you don't need to load up the OpenAPI doc prior to generating the form for the specific operation.

zrob: I don't understand why display names should come back in this response.  Why isn't display name included in the schema that comes from discovery_url?

spec.md
spadgett: > Async ops, if enabled, would be defined via the swagger file. The platform could detect and subscribe to ops based on the returned swagger.

I'm not sure how a platform can detect it only from the swagger. A human looking at it might notice an endpoint called `/backup/<id>/status` and guess that it's for checking status of an async operation. But how can the platform know programmatically to do that?

As an end user, I'd be worried if I couldn't check the status of an operation like "Restore" and know that it completed successfully.

I could see it if the platform recognized the `adheres_to` and had special handling for that particular standard.

rhodie27: I think the short answer is it's not going to be possible across the board for all APIs.

My thought was that the platform could recognize a callback object in the openapi doc, (https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#callbackObject) and other openapi patterns to figure out what it needs, but we won't know until someone tries.

It's the main issue with going the generic route on extensions.

spec.md
rhodie27: @zrob the request was to have a way to show something in a platform UI before having to call the openapi document. See the comment above this, https://github.com/openservicebrokerapi/servicebroker/pull/431/files#r171010501.

Also, OpenAPI doesn't have a consistent way to give operations display names. They have the operationid, description, summary, but they all serve other purposes. And I wanted to stay away from telling people how to define their openapi document, (as much as I could), so that existing ones could be used.

spec.md
n3wscott: I would like to push back on this field entirely and I apologize for not being in the original discussions for it. 

I think alt_instance_id is an attempt at serving pre-processed open api information and the correct way to solve this issue is with vendor extensions in the open api doc. You could add a key, say `"x-osb-instance-id": "instance_name"` on the definition of the resource and it will allow you to do the same thing without polluting the response from provisioning.

spec.md
n3wscott: I would also push back on this field. `operation_display_names` could be accomplished in other ways, like a lookup service that takes the `adhears_to` (plus more data?) url and returns the same list, maybe hosted by the broker.

I think we need to draw a line on the data that is returned from a provision is data that is specific to the instance that was created, and when instances of the same type start to return the exact same data is a red flag for me.

Are we going to need the platform to remember the exact `operation_display_names` per instance? Because if we let it be in the response, the broker could make this list dynamically as provisioning happens.

Then we also need to talk about how does this list get updated as the instance is upgraded under the brokers control and new extensions are enabled. The only way to be sure is to fetch and process the open api doc.

spec.md
n3wscott: The answer to this from the googs perspective is yes. If the platform would like to invoke them then that should be ok, and if the platform provides all the information to a third-party that should work out as well.

duglin:
Platform invoked them

spec.md
rhodie27: Apology accepted. I agree, this could be done via vendor extentions, however, the overall theme I have been trying to push is to not make the service provider change their OpenAPI document to fit the service broker's needs. Perhaps the service broker author doesn't have that ability/power or the service provider doesn't want a million different vendor customizations in their pretty swagger. I guess we could always say that the Service Broker author needs to maintain their own OpenAPI document... Sounds messy.

spec.md
rhodie27: > I would also push back on this field. operation_display_names could be accomplished in other ways, like a lookup service that takes the adhears_to (plus more data?) url and returns the same list, maybe hosted by the broker.

The request from a few different people, (mostly RH), is they want a way to get a list of keywords associated with the service instance actions, prior to having to load and parse the openapi document. Adhears_to will not be available 100% of the time, and typically will not be used when you are interested in dynamically generating any returned openapi document.

> I think we need to draw a line on the data that is returned from a provision is data that is specific to the instance that was created, and when instances of the same type start to return the exact same data is a red flag for me.

I do agree we need to draw a line, I believe this is it for validating phase. We can always remove/change if it doesn't work IRL.

> Are we going to need the platform to remember the exact operation_display_names per instance? Because if we let it be in the response, the broker could make this list dynamically as provisioning happens.

The same could be said for the OpenAPI document, in general. Maybe it's different per service_instance...

> Then we also need to talk about how does this list get updated as the instance is upgraded under the brokers control and new extensions are enabled. The only way to be sure is to fetch and process the open api doc.

The same response will be required on a service_instance update. I just haven't added that to the PR yet.

spec.md
rhodie27: openapi_instance_key is confirmed

rhodie27: vote is to remove this for now. Mike to add language about the openapi doc must do this somehow.

spec.md
gberche-orange: Can the PR clarify the concurrency aspects that the platform/users should comply with when invoking extensions ? For instance, should a broker exposing a `restore` extension-api expect concurrent calls to the endpoint with potentially conflicting parameters ?

This relates to #467

spec.md
```diff
@@ -841,13 +841,93 @@ For success responses, the following fields are defined:
...
 | operation | string | For asynchronous responses, Service Brokers MAY return an identifier representing the operation. The value of this field MUST be provided by the Platform with requests to the [Last Operation](#polling-last-operation) endpoint in a percent-encoded query parameter. If present, MUST be a non-empty string. |
+| extension_apis | array of [Extension API](#extension-api-object) objects | For extensions to the Service Broker API, Service Brokers MAY return one or more `extension_api` objects that describe additional API endpoints via an OpenAPI document. See [Extension API Object](#extension-api-object) for more information. |
+
+##### Extension API Object
+
+The `extension_api` object MAY be used to describe any additional endpoint
+to the Open Service Broker API. An example of this could be lifecycle
+management of a Service Instance, (e.g. "Day Two Operations"), like Backup,
```
gberche-orange: I would suggest the specs to clarify when an extension_api should be defined and used by broker authors w.r.t. to asking 1st class support in the OSB API. 

Are platforms expected to provide UI/CLI tooling for any declared extensions (i.e. dynamically generating UI from the openAPI document) ? Would generated UIs and user workflows be restricted to single API endpoint calls, or would instead spawn across multiple API calls (e.g. list backups, restore backup) ?

Are platforms instead expected to additionally support for specific curated extensions and bring optimized user experience for them? If so, where/how would such curated extensions be defined and published ?

Can the specs recommend use-cases for extensions such as:
1- operator specific extension :  a platform operators is adding custom features to a broker and brings custom UI on along it.
2- vendor specific extensions (say K8S/CF vendor X way of offering backup which differ from vendor Y)
3- platform specific extensions (e.g. CF AR service metrics ingestion which differs from K8S metrics ingestion) that can't be supported by platform-specific [profiles](https://github.com/openservicebrokerapi/servicebroker/blob/master/profile.md#cloud-foundry-service-metadata)

For the mentioned day 2 operations `Backup, Restore, Stop, Start, Restart and Pause`, are they fitting any of the cases above ?

n3wscott: >I would suggest the specs to clarify when an extension_api should be defined and used by broker authors w.r.t. to asking 1st class support in the OSB API.

The current thinking is it is entirely up to the broker author, and the platform can choose to provide, at minimum, the OpenAPI UI (or even just the OpenAPI url). Then platforms can extend the UI based on 'adheres_to'.

> Are platforms expected to provide UI/CLI tooling for any declared extensions (i.e. dynamically generating UI from the openAPI document) ? Would generated UIs and user workflows be restricted to single API endpoint calls, or would instead spawn across multiple API calls (e.g. list backups, restore backup, delete backup) ?

We had talked about how to batch calls but it became clear that this could be designed forever and we decided to support just the single endpoint as a first step, but nothing stops a clever broker from providing batch/aggregation on their own. 

> Are platforms instead expected to additionally support specific curated extensions and bring optimized user experience for them? If so, where/how would such curated extensions be defined and published ?

`adheres_to` is the mechanism we think will do this. If it is proven that this is not enough we can always revisit the spec to add what is lacking.

> Can the specs recommend use-cases for extensions such as:
>1- operator specific extensions : a platform operators is adding custom features to a broker and brings custom UI on along it.
>2- service specific extensions (say stop/resume operation) that only applies to a category of services
>3- vendor specific extensions (say K8S/CF vendor X way of offering backup which differ from vendor Y)
>4- platform specific extensions (e.g. CF AR service metrics ingestion which differs from K8S metrics ingestion) that can't be supported by platform-specific profiles

I suspect this is out of scope for the current PR, at least until we see some real world emergent behavior for generic actions.

duglin:
I believe we were supposed to find out how much interest we each had on this one... from our side we are still interested in the idea in general but have no immediate need for it.  I guess that means its a "nice to have" but not a "show stopper" or "must have" for us right now.

Samze:
After a period of inactivity we are now revisiting generic instance extensions. There is a new [proposal doc here](https://docs.google.com/document/d/1OKlsQKUab3yvOyQjvh8HFFcIfkILyFMaq0xngTdTcek/edit).

mattmcneeney:
Closing this as we are planning on moving forward with https://github.com/openservicebrokerapi/servicebroker/pull/670

PR #437, opened 2018-02-01
Add support for accepts_updates=true on Update operation

duglin:
We have cases where upon doing an update we have a new dashboard_url
for the service instance. This PR allows for the Platform to indicate
that it supports the Broker sending back a new URL, and it allows for
the Broker to do it.


mattmcneeney:
Hey @duglin Can you explain the use case for changing the `dashboard_url` a bit more?

Does this happen when changing plans?

I wonder if either the GET endpoints or extensions proposals could help here...

avade:
I like the idea of the service broker being able to indicate to the platform that it should do a GET. Would work well in the async world.

Likely need to think more on this, also interested in the use case.

mattmcneeney:
This could also be an opportunity to help brokers who may not be able to return a `dashboard_url` at the start of an async provision (this has always felt weird to me).

duglin:
@mattmcneeney we have situations where updating the plan or parameters causes something similar to a migration and as a result a new dashboard_url is needed. In our cases the URL/creds of the service itself don't change, just the dashboard_url.

fmui:
@mattmcneeney We also have brokers that cannot return a `dashboard_url` at the start of an async provision. Something like this might help.

n3wscott:
LGTM

I think this is a good idea. It means that the broker could return a dashboard url which could be the pending or history page and then on update could repoint to the final or updated dashboard url.

duglin:
3 more needed

spec.md
```diff
@@ -1003,10 +1003,22 @@ For success responses, the following fields are defined:
 
 | Response Field | Type | Description |
 | --- | --- | --- |
+| dashboard_url | string | The updated URL of a web-based management user interface for the Service Instance; we refer to this as a service dashboard. The URL MUST contain enough information for the dashboard to identify the resource being accessed (`0129d920a083838` in the example below). Note: a Service Broker that wishes to return `dashboard_url` for a Service Instance MUST return it with the initial response to the update request, even if the service is updated asynchronously. If present, MUST be a non-empty string. |
```
pmorie: Is this really the updated URL?

duglin: I'm not following your question

mattmcneeney:
I’m happy to move this into validation through implementation, but whoever picks this up, it would be good to weigh up the pros and cons of using the new service instance GET endpoint to look for an updated dashboard URL after an update has completed vs adding this additional logic.

Using the GET endpoint has the additional benefit of working for async updates where the broker cannot return the dashboard URL at the start of the operation.

duglin:
Agreed to move to validate- thru-impl on today's call

jberkhahn:
This has been implemented in Kube Service Catalog, see the linked PR's up above and review this for merging into the spec.

duglin:
rebased and ready for review. Removing the "validate thru impl" flag per today's call.

duglin:
Need some reviews of this

duglin:
ping @pmorie @vaikas-google @fmui @zrob

spec.md
vaikas-google: I assume he means that it's not necessarily updated? It could be the original one?

duglin: That's an impl choice I guess - we don't mandate the value sent.

vaikas-google:
LGTM

zrob:
LGTM

spec.md
```diff
@@ -1082,10 +1082,22 @@ For success responses, the following fields are defined:
...
 | operation | string | For asynchronous responses, Service Brokers MAY return an identifier representing the operation. The value of this field MUST be provided by the Platform with requests to the [Last Operation](#polling-last-operation) endpoint in a percent-encoded query parameter. If present, MUST be a non-empty string. |
 
+The ability to include updated metadata in the response message was
+added in version 2.14. Any Platform sending an API Version Header with a
+value of 2.14 or greater MUST use the new values specified in the response
```
mattmcneeney: @n3wscott @duglin This seems to conflict with the compatibility matrix? We are saying platforms MUST support this in v2.14 and later.

n3wscott: Yeah it should be added, and that is why we need to commit the matrix PR :D

PR #446, opened 2018-02-12
Add Platform recommendations for dealing with "plan_updateable" and "bindable"

mattmcneeney:
The `"plan_updateable"` and `"bindable"` fields in the catalog define what behaviours a Service Broker support. This change adds text stating that Platforms SHOULD NOT make a bind or update service instance call to a Service Broker that sets these fields to false.

This seems especially important for `"bindable"` as the spec currently says:
```
Service Brokers that do not offer any bindable services do not need to implement the endpoint for bind requests.
```

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -334,7 +334,7 @@ $ curl http://username:password@service-broker-url/v2/catalog -H "X-Broker-API-V
 
 CLI clients will typically have restrictions on how names, such as service
 and plan names, will be provided by users. Therefore, this specification
-defines a "CLI-friendly" string as a short string that MUST only use 
+defines a "CLI-friendly" string as a short string that MUST only use
 alphanumeric characters and hyphens, with no spaces. This will make it easier
 for users when they have to type it as an argument on the command line.
 For comparison purposes, service and plan names MUST be treated as
@@ -870,8 +870,14 @@ Service Instance to other plans. By modifying parameters, users can change
 configuration options that are specific to a service or plan.
 
 To enable support for the update of the plan, a Service Broker MUST declare
-support per service by including `plan_updateable: true` in its [catalog
+support per service by including `"plan_updateable": true` in its [catalog
 endpoint](#catalog-management).
+
+If `"plan_updateable": true` is declared for a plan in the
+[Catalog](#catalog-management) endpoint, the Platform MAY request an update of
+a Service Instance using the given plan. Otherwise, Platforms MUST NOT make an
+update request to the Service Broker for any Service Instance using the given
+plan.
 
 Not all permutations of plan changes are expected to be supported. For
 example, a service might support upgrading from plan "shared small" to "shared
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -870,8 +870,10 @@ Service Instance to other plans. By modifying parameters, users can change
 configuration options that are specific to a service or plan.
 
 To enable support for the update of the plan, a Service Broker MUST declare
-support per service by including `plan_updateable: true` in its [catalog
-endpoint](#catalog-management).
+support per service by including `"plan_updateable": true` in its [catalog
+endpoint](#catalog-management). If `"plan_updateable": false` is declared,
+Platforms SHOULD NOT make an update request to the Service Broker for any
+Service Instance using the given plan.
```
duglin: Should the absence of this field mean "false" as well?  Seems like it.

duglin: Oh, just noticed you did that for bindable using slightly different wording. Any reason we can't do the same here?

duglin: SHOULD NOT or MUST NOT?

I can't think a reason why we should give a Platform a choice about this. Although, Broker should we prepared to deal with rogue Platforms.

mattmcneeney: The main reason I used SHOULD not MUST was to not introduce a breaking change. Will fix the wording!

duglin: Do you think the wording today allows for it? At best it seems like its ambiguous and this would be more like a clarifying change since IMO this PR doesn't really change the semantics of the spec - just makes it clear that "false" == absence == don't do it.

duglin: You don't wanna make it a MUST NOT, eh? :-)

mattmcneeney: I'll very happily make this a MUST NOT if everyone is on board!

mattmcneeney: Done :)

spec.md
```diff
@@ -1014,13 +1020,14 @@ For success responses, the following fields are defined:
 
 ## Binding
 
-If `bindable:true` is declared for a service or plan in the
+If `"bindable": true` is declared for a service or plan in the
 [Catalog](#catalog-management) endpoint, the Platform MAY request generation
-of a Service Binding.
+of a Service Binding. Otherwise, Platforms SHOULD NOT make a binding request to
```
duglin: MUST NOT? for consistency?

duglin:
LGTM

duglin:
easy review - 3 needed

fmui:
LGTM

spec.md
```diff
@@ -1014,13 +1020,14 @@ For success responses, the following fields are defined:
 
 ## Binding
 
-If `bindable:true` is declared for a service or plan in the
+If `"bindable": true` is declared for a service or plan in the
 [Catalog](#catalog-management) endpoint, the Platform MAY request generation
-of a Service Binding.
+of a Service Binding. Otherwise, Platforms MUST NOT make a binding request to
+the Service Broker any Service Instance using the given service or plan.
```
mattmcneeney: for any Service Instance

n3wscott:
LGTM

spec.md
duglin: @mattmcneeney did you want to update this?

mattmcneeney: Thanks, this one slipped past me. Updated!

n3wscott:
LGTM

duglin:
LGTM

fmui:
LGTM

duglin:
LGTM

duglin:
One more needed - @mattmcneeney ?

duglin:
LGTM

zrob:
lgtm

PR #456, opened 2018-02-15
Make 408's not result in orphan mitigation

### LINKED ISSUE #449: 408 (Request Timeout) is a client-side error and should not require orphan mitigation
In the [orphan mitigation section](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#orphans) there is a special case for 408 separate from other 4xx errors:

| Status Code Of Service Broker Response | Platform Interpretation Of Response | Platform Initiates Orphan Mitigation? |
| --- | --- | --- |
| 408 | Timeout failure | Yes |
| All other 4xx | Request rejected | No |

`408 Request Timeout` means:
> "The server did not receive a complete request message within the time that it was prepared to wait."

It's a client-side error just as other `4xx` codes, i.e. **server explicitly rejected the request** without processing it (hence no side effects that would require orphan mitigation). 
There is already a separate item for `Timeout` and `5xx` errors which require orphan mitigation, including the `504 Gateway Timeout` error code, in which case the platform indeed doesn't know whether broker has successfully provisioned a new instance or not.

duglin:
/cc @nilebox

Closes #449

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1430,7 +1430,7 @@ Platforms SHOULD initiate orphan mitigation in the following scenarios:
 | 201 | Success | No |
 | 201 with malformed response | Failure | Yes |
 | All other 2xx | Failure | Yes |
-| 408 | Timeout failure | Yes |
+| 408 | Client timeout failure (request not received at the server) | No |
 | All other 4xx | Request rejected | No |
 | 5xx | Service Broker error | Yes |
 | Timeout | Failure | Yes |
--- a/tools/verify-tables.sh
+++ b/tools/verify-tables.sh
@@ -101,9 +101,9 @@ function checkForPunc() {
       continue
     fi
 
-    # Ignore columns containing 4 words or less
-    count=$(echo "$col" | wc -w)
-    if (( "$count" <= 4 )); then
+    # Anything less than 4 words we ignore
+    count=$(echo "$col" | sed "s/(.*)//g" | sed "s/\[.*\]//g" | wc -w)
+    if (( "$count" < 4 )); then
       continue
     fi
 
@@ -111,7 +111,7 @@ function checkForPunc() {
       continue
     fi
 
-    echo "$file - $1: column $colNum doesn't end with a '. |' or '? |' - watch for extra/missing spaces."
+    echo "$file - $1: column $colNum has more than 3 words and doesn't end with a '. |' or '? |' - watch for extra/missing spaces."
 
   done
 }


duglin:
While we could technically merge the two 4xx lines, I think its important to call out the 408 by itself so we can be clear that this is a "client side" timeout and not a "server side" one - which are very very different w.r.t. orphan mitigation

nilebox:
TBH I would prefer to remove the special mention of `408` completely, as it's not actually special as we discussed in #449

duglin:
@nilebox my original version of the PR did merge them and within 5 minutes I got a ping about it because the distinction between client-side vs server-side timeout wasn't clear - so I added it back in, just for the sake of clarity.   We don't need to call-out `504` because it results in orphan mitigation which is what most people would expect from a "timeout".

nilebox:
@duglin I see your concern, but the distinction between client-side and server-side errors is an HTTP standard's problem, OSB doesn't bring anything additional there. Furthemore, HTTP standard makes the distinction very clear - `4xx` are client-side errors, and `5xx` are server-side errors. If we needed to alter this distinction (which OSB spec currently does), it would obviously require a special wording in the OSB spec. But since we don't (anymore), I don't see the value of keeping `408` separate from `4xx`.

spec.md
```diff
@@ -1430,7 +1430,7 @@ Platforms SHOULD initiate orphan mitigation in the following scenarios:
 | 201 | Success | No |
 | 201 with malformed response | Failure | Yes |
 | All other 2xx | Failure | Yes |
-| 408 | Timeout failure | Yes |
+| 408 | Client timeout failure (request not received at the server) | No |
```
n3wscott: I don't understand the usage of 'server' here. A client could get a 408 by having it's local timeout setting very small. Or you could be talking to a broker proxy and the second connection could timeout. Both cases could result in orphans. I am not sure how a broker author could prevent their implementation from returning a 408 AND/OR make sure that orphans are not possible.

duglin: perhaps "receiver" would be better, but in the other cases you mentioned I'm not sure the Platform would see a 408. As long as the message makes it to the next hop I don't think 408 would be appropriate. And I would think that hop before that one would return a 5xx since from its POV it did receive the entire request so it can't be a 408.  Right?

n3wscott: Looking up [408 in the rfc](https://tools.ietf.org/html/rfc7231#section-6.5.7)...

Ok I agree. In this case, server is fine with me.

n3wscott:
LGTM

fmui:
LGTM

mattmcneeney:
LGTM

zrob:
lgtm

n3wscott:
LGTM

duglin:
2 more reviews needed
@fmui @vaikas-google @mattmcneeney

fmui:
LGTM

vaikas-google:
LGTM

Issue #453, opened 2018-02-13
Spec should explicitly state that duplicate requests should return 202 AND the original operation.


khipkin:
Currently, the spec says that `202 Accepted` should be returned if the work is in progress for provisioning, updating a service instance, binding, unbinding, and deprovisioning when another request is received for the same work. Readers can assume that the operation returned with `202` should be the same as was returned by the original request.

I believe that this should be an explicit requirement stated in the spec for each of provisioning, updating a service instance, binding, unbinding, and deprovisioning.

mattmcneeney:
@khipkin Thanks for raising this. I think this makes sense and would be a useful piece of clarification in the spec. Do you want to raise a PR for this or are you happy for me to?

khipkin:
I can send a PR! Thanks.

khipkin:
First PR is for instance operations, requesting a merge into master.

Second PR will be for binding operations, requesting a merge into the async-bindings proposal branch.

mattmcneeney:
@khipkin It looks like this is all good to go, so I'll close this.

khipkin:
@mattmcneeney I was actually still waiting for the async-bindings proposal branch to get merged, since the async binding and unbinding descriptions still need to be updated.

mattmcneeney:
Ah good point! I've updated the PR now #334. Can you double check that looks OK to you?

PR #457, opened 2018-02-15
Spec says the operation returned with `202` for in-progress instance work should be the same as was returned by the original request.

khipkin:
Partially addresses #453.

Instance provisioning, updating, and deprovisioning response sections now specify that, when 202 is returned in response to duplicate requests while the work is still in progress, the operation ID returned should match that of the original response.


duglin:
LGTM

duglin:
Hopefully, easy review - 3 needed

duglin:
Instead of saying "Operation ID" the WG would prefer something like "The `operation` string..." just so people don't think we're implying it must be a GUID

khipkin:
Done!

n3wscott:
LGTM

duglin:
LGTM

duglin:
2 more easy reviews needed

fmui:
LGTM

jmrodri:
LGTM

zrob:
lgtm

mattmcneeney:
LGTM

PR #458, opened 2018-02-18
Be clear about how unique/persistent service/plan IDs are

duglin:
ping @vaikas-google - and what is Martin's github id?

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -362,7 +362,7 @@ It is therefore RECOMMENDED that implementations avoid such strings.
 | Response Field | Type | Description |
 | --- | --- | --- |
 | name* | string | A CLI-friendly name of the service. MUST only contain alphanumeric characters and hyphens (no spaces). MUST be unique across all service objects returned in this response. MUST be a non-empty string. |
-| id* | string | An identifier used to correlate this service in future requests to the Service Broker. This MUST be globally unique. MUST be a non-empty string. Using a GUID is RECOMMENDED. |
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -362,7 +362,7 @@ It is therefore RECOMMENDED that implementations avoid such strings.
 | Response Field | Type | Description |
 | --- | --- | --- |
 | name* | string | A CLI-friendly name of the service. MUST only contain alphanumeric characters and hyphens (no spaces). MUST be unique across all service objects returned in this response. MUST be a non-empty string. |
-| id* | string | An identifier used to correlate this service in future requests to the Service Broker. This MUST be globally unique. MUST be a non-empty string. Using a GUID is RECOMMENDED. |
+| id* | string | An identifier used to correlate this service in future requests to the Service Broker. This MUST be globally unique such that users MUST be able to assume that seeing the same value (no matter what Service Broker uses it) will always refer to this service. MUST be a non-empty string. Using a GUID is RECOMMENDED. |
```
n3wscott: users or Platforms? you can read this such that the platform appends something to the ID to make sure this holds true... So is this the job of the Service Broker or the Platform?

duglin: hmm should probably be "Platform" since its the user in this scenario. I'm not sure about the rest of your comment though. I agree that technically a Platform could modify an ID of a service, but that's only meaningful if it exposes that to its users - but that's out of scope for us. When talking to the Broker it must always use the ID the Broker provided. So, I guess my answer to your question is: Broker.

duglin: Made it "Platform" but I couldn't bring myself to leave off the idea of Platform's users because in the end they're the ones that will choose the service/plan so they're (ultimately) the people that care. See what you think....

n3wscott:
LGTM

duglin:
reviews needed

fmui:
LGTM

zrob:
lgtm

mattmcneeney:
LGTM

Issue #451, opened 2018-02-13
Change how we link to type definitions inside the spec


luksa:
I suggest we change the way we link to the definitions of OSBAPI types in the spec. 

For example, let's take the row that shows the "schemas" field in https://github.com/openservicebrokerapi/servicebroker/blob/v2.13/spec.md#plan-object. 

I suggest we replace the word "object" in the second column with the following text: "[Schema](https://github.com/openservicebrokerapi/servicebroker/blob/v2.13/spec.md#schema-object) object". As shown here, the name of the type links to the description of said type.

Similarly, the second column in the "plans" row under [Service Objects](https://github.com/openservicebrokerapi/servicebroker/blob/v2.13/spec.md#service-objects) would be "array of [Plan](https://github.com/openservicebrokerapi/servicebroker/blob/v2.13/spec.md#plan-object) objects" or simply "[][Plan](https://github.com/openservicebrokerapi/servicebroker/blob/v2.13/spec.md#plan-object)" instead of "array-of-objects" as we have now.

luksa:
Something along the lines of: https://github.com/luksa/servicebroker/blob/9620a17665768dcde2f4d6a426d7d7eea2f6f9a1/spec.md#plan-object

duglin:
@luksa want to open a PR so people can see the proposed edits in context?
ok if I assign this and #450 to you?

luksa:
Closing (fixed by #459)

Issue #450, opened 2018-02-13
Minor naming inconsistencies around schemas


luksa:
I'd like to point out a few inconsistencies related to how the various schema objects are named.

1. "[Schema object](https://github.com/openservicebrokerapi/servicebroker/blob/v2.13/spec.md#schema-object)" should be changed to "Schemas object", since it contains two schemas (also, the type in the OSBAPI OpenAPI spec is already called `SchemasObject`. See [here](https://github.com/openservicebrokerapi/servicebroker/blob/master/openapi.yaml#L481).

2. "[Service Instances object](https://github.com/openservicebrokerapi/servicebroker/blob/v2.13/spec.md#service-instances-object)" should be changed to "Service Instance Schema object" to match the OpenAPI spec [here](https://github.com/openservicebrokerapi/servicebroker/blob/master/openapi.yaml#L489)

3. "[Service Bindings Object](https://github.com/openservicebrokerapi/servicebroker/blob/v2.13/spec.md#service-bindings-object)" should be changed to "Service Binding Schema object" to match the OpenAPI spec [here](https://github.com/openservicebrokerapi/servicebroker/blob/master/openapi.yaml#L497)

Also, since `ServiceInstanceSchema` contains multiple schemas, it should be called `ServiceInstanceSchemas` to keep it consistent with the `Schemas` object. Same for `ServiceBindingSchemas`.

mattmcneeney:
@luksa These inconsistencies look legitimate and should be fixed. Do you want to open up a PR to do this?

luksa:
Closing (fixed in #459)

PR #459, opened 2018-02-21
Fix type links and names

luksa:
Addresses #450 and #451

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -353,11 +353,11 @@ It is therefore RECOMMENDED that implementations avoid such strings.
 
 | Response Field | Type | Description |
 | --- | --- | --- |
-| services* | [Service](#service-objects)\[\] | Schema of service objects defined below. MAY be empty. |
+| services* | [Service](#service-object)\[\] | Schema of service objects defined below. MAY be empty. |
 
 \* Fields with an asterisk are REQUIRED.
 
-##### Service Objects
+##### Service Object
 
 | Response Field | Type | Description |
 | --- | --- | --- |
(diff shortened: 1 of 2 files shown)


n3wscott:
This reads really nicely. My only comment is changing the logical section name to the variable name, example: _Input Parameters Object_ --> _InputParametersSchema Object_ reads more like code rather than spec. But I am not opposed to it, maybe others can comment on this point. 

LGTM

duglin:
hadn't really notice the squashing of those words - now that you point it out I agree - let's add spaces so its less codey.

luksa:
I've added back the spaces. 

Note that I didn't want to change the type names in the OpenAPI schema definition itself, so I kept `ServiceInstanceSchema` singular instead of changing it to `ServiceInstanceSchemas` as I suggested in #450. The singular form is perfectly fine, but I would like to point out that there is a slight inconsistency between `Schemas` and `ServiceInstanceSchema/ServiceBindingSchema`. Do we want to fix that?

Also, there's a mismatch between the spec and the OpenAPI schema in the `InputParametersSchema` object. The type is called `SchemaParameters` in the OpenAPI schema. I haven't renamed it, since #392 changes the name to "JSON Schema Objects" (this name needs to be changed as I pointed out in that PR).

n3wscott:
LGTM

pmorie:
LGTM

spec.md
```diff
@@ -353,24 +353,24 @@ It is therefore RECOMMENDED that implementations avoid such strings.
 
 | Response Field | Type | Description |
 | --- | --- | --- |
-| services* | array-of-service-objects | Schema of service objects defined below. MAY be empty. |
+| services* | [Service](#service-object)\[\] | Schema of service objects defined below. MAY be empty. |
```
duglin: Can you make these "An array of ..." to match how we normal talk about arrays?

n3wscott:
LGTM

duglin:
LGTM

pmorie:
LGTM

fmui:
LGTM

PR #464, opened 2018-03-02
Openapi schema fixes

luksa:
This fixes two things:
- removes "Object" suffix from some type names
- points the JSONSchema type definition to http://json-schema.org/draft-04/schema instead of it it being an arbitrary object (`type: object`)

I removed the "Object" suffix because it is unnecessary and makes the naming of these types inconsistent with other type definitions in the schema. For example, in the spec document, we call it "Plan Object", but in the OpenAPI schema, we simply call it `Plan`.

The following type definitions contained the "Object" suffix:
- SchemasObject
- ServiceInstanceSchemaObject
- ServiceBindingSchemaObject
- JSONSchemaObject

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -507,7 +507,7 @@ components:
           $ref: '#/components/schemas/JSONSchema'
 
     JSONSchema:
-      type: object
+      $ref: 'http://json-schema.org/draft-04/schema'
 
     ServiceInstanceProvisionRequest:
       type: object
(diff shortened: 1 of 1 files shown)


n3wscott:
LGTM

pmorie:
cc @spadgett

pmorie:
LGTM

duglin:
LGTM

mattmcneeney:
LGTM

PR #466, opened 2018-03-06
Add a `clusterID` to the profile for kube

duglin:
ping @pmorie - see if this aligns with what we talked about today.

### DIFF
--- a/profile.md
+++ b/profile.md
@@ -214,13 +214,25 @@ The following properties are defined for usage within a Kubernetes deployment:
   "namespace": "testing"
   ```
 
+- `clusterID`<br>
+  The unique identifier for the Kubernetes cluster from which the request
+  was sent. This property MUST be a non-empty string serialized as follows:
+
+  ```
+  "clusterID": "id-goes-here"
+  ```
+  For example:
+  ```
+  "clusterID": "644e1dd7-2a7f-18fb-b8ed-ed78c3f92c2b"
+  ```
+
 The following table specifies which properties MUST appear in each API:
 
 | Request API | Properties |
 | --- | --- |
-| `PUT /v2/service_instances/:instance_id` | `namespace` |
-| `PATCH /v2/service_instances/:instance_id` | `namespace` |
-| `PUT /v2/service_instances/:instance_id/service_bindings/:binding_id` | `namespace` |
+| `PUT /v2/service_instances/:instance_id` | `namespace`, `clusterID` |
+| `PATCH /v2/service_instances/:instance_id` | `namespace`, `clusterID` |
+| `PUT /v2/service_instances/:instance_id/service_bindings/:binding_id` | `namespace`, `clusterID` |
 
 Example:
 
@@ -229,7 +241,8 @@ part of a Kubernetes API call:
   ```
   "context": {
     "platform": "kubernetes",
-    "namespace": "development"
+    "namespace": "development",
+    "clusterID": "8263feba-9b8a-23ae-99ed-abcd1234feda"
   }
   ```
 
--- a/profile.md
+++ b/profile.md
@@ -131,6 +131,11 @@ This section will define a set of properties for each Platform and specify
 when each is meant to be used. Platforms MAY choose to provide additional
 properties beyond the ones defined in this document.
 
+Each property will include the version of the Open Service Broker
+API in which it was introduced. For any API call in which that property
+is mandatory, it is only mandatory for the specified version of the
+specification or higher.
+
 Aside from the Platform specific properties, defined in the following
 sections, there is one common property called `platform` that
 MUST also appear within `context` to indicate which Platform is being used.
@@ -154,7 +159,10 @@ for both.
 The following properties are defined for usage within a Cloud Foundry
 deployment:
 
-- `organization_guid`  
+- `organization_guid`
+
+  Version: 2.11
+
   The GUID of the organization that a Service Instance is associated with.
   This property MUST be a non-empty string serialized as follows:
   ```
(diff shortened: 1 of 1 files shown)


profile.md
```diff
@@ -214,13 +214,25 @@ The following properties are defined for usage within a Kubernetes deployment:
   "namespace": "testing"
   ```
 
+- `clusterID`<br>
```
n3wscott: I don't think you want that `<br>`

profile.md
```diff
@@ -229,7 +241,8 @@ part of a Kubernetes API call:
   ```
   "context": {
     "platform": "kubernetes",
-    "namespace": "development"
+    "namespace": "development",
+	"clusterID": "8263feba-9b8a-23ae-99ed-abcd1234feda"
```
n3wscott: there is a tabs and spaces mismatch here.

profile.md
```diff
@@ -214,13 +214,25 @@ The following properties are defined for usage within a Kubernetes deployment:
...
+  ```
+
 The following table specifies which properties MUST appear in each API:
 
 | Request API | Properties |
 | --- | --- |
-| `PUT /v2/service_instances/:instance_id` | `namespace` |
-| `PATCH /v2/service_instances/:instance_id` | `namespace` |
-| `PUT /v2/service_instances/:instance_id/service_bindings/:binding_id` | `namespace` |
+| `PUT /v2/service_instances/:instance_id` | `namespace`, `clusterID` |
```
n3wscott: does MUST break the spec for backwards compatibility?

if a 2.new broker client sends this to a 2.old broker, it would ignore it and perhaps this is ok.
if a 2.old broker client tries to talk to a 2.new broker, it would error. 

That seems like a breaking change and should move to v3.

profile.md
duglin: Actually I do - I wanted it to insert a LF, just like the other ones have.

profile.md
duglin: I think we're in sort of a catch-22 situation here. We want this is be required for kube once morgan's svc-cat PR is merged.
Perhaps some wording for each field to indicate which release it was added and to then say its required for all releases after that version?

n3wscott: But that is an API breaking change, and if you are using semver, the proper thing to do is to bump the major version.

profile.md
n3wscott: oh weird. why not just have a new line and tab in and have markdown do the right thing? Perhaps it is all the rest that are also wrong :D

profile.md
duglin: Worded correctly it wouldn't be a breaking change. If we said "starting with v2.14 all Kube platforms must include it" then all existing kubes are ok. We just need to make sure we get the clusterID PR merged into svc-cat to align with the v2.14 version bump. Old brokers are ok since they'll work with either.

profile.md
duglin: See: https://gist.github.com/duglin/d2d0ee2bb6761f1187e0bf2f58a5aa39  the last clusterID/busterID examples. Perhaps I'm doing it wrong.

profile.md
n3wscott: I think we have a disagreement about (semver)[https://semver.org/]. There is no wording that would make this a non-api breaking change without either a) bumping the spec to v3, or b) make it optional for all v2 implementations.

duglin: we're in kind of a weird spot. The profile itself is optional, but we need a way for a broker to know what fields to expect AND need to be able to add new fields w/o a version bump. Note, again, adding new fields won't break brokers so it all a matter of whether the platforms are ok with a version gated new field.  Let me work on some tweaks...

duglin: @n3wscott see what you think. While this is still technically a breaking change I'm now considering the old stuff to be a bug. I don't believe we ever intended for Kube to only get one shot at getting the list of context fields right w/o forcing a major version bump. So my new version tries to address that by at least not breaking existing Platforms but allows for us to place requirements on newer ones (all within the scope of the entire `context` thing being optional overall).

n3wscott: Thanks for the update. I think this makes it more clear. And I am less un-easy about this change given it is for the profile for k8s and we have a pretty tight loop over k8s integrations. There is now a pretty clear expectation around version of the spec and how it relates to the profile.

n3wscott:
LGTM

profile.md
```diff
@@ -214,13 +228,30 @@ The following properties are defined for usage within a Kubernetes deployment:
   "namespace": "testing"
   ```
 
-The following table specifies which properties MUST appear in each API:
+
+- `clusterID`
+
+  Version: 2.13
```
MHBauer: 2.13 is out already, 2.14? Or can this profile change at any time?

duglin: doi! fixed

duglin:
Need re-reviews due to new commit

duglin:
@n3wscott can you review again - it doesn't seem to have picked yours up.
Also, we need 3 more reviews.

n3wscott:
LGTM

...robots.

n3wscott:
> it doesn't seem to have picked yours up.

I wonder if the issue is because that was not a generic top level PR comment but a end of review comment... I bet that is the issue.

avade:
LGTM - we should look at adding the "WHY" this field is important or use cases.

duglin:
LGTM testing something

duglin:
not sure what's going on, but @avade and @n3wscott can you guys re LGT M ?

vaikas-google:
LGTM

duglin:
@pmorie @avade @fmui can we get some reviews?

duglin:
actually, @avade we just need you to redo yours

mattmcneeney:
LGTM

fmui:
LGTM

avade:
LGTM

PR #480, opened 2018-03-27
Add some text explaining clusterid

### LINKED ISSUE #478: Clarify 'cluster ID' part of kube profile
I don't think the change merged in #466 is specific enough, and I worry that we're backing Kubernetes integration with this API into a corner.

I'm primarily worried that the addition to the spec isn't clear enough about what 'unique' means. Does it mean that each kube cluster has to have a single, immutable, unique ID forever? If so, I think we should change the language. I worry that someone could change the ID of a kube cluster and break compliance with the spec. It's going to be basically impossible to guarantee that a unique value for an instance of a cluster never changes.

duglin:
Closes #478

### DIFF
--- a/profile.md
+++ b/profile.md
@@ -234,7 +234,14 @@ The following properties are defined for usage within a Kubernetes deployment:
   Version: 2.14
 
   The unique identifier for the Kubernetes cluster from which the request
-  was sent. This property MUST be a non-empty string serialized as follows:
+  was sent. This property can be useful when a Service Broker tries to
+  call back to a Platform that is managing multiple Kubernetes clusters.
+  Typically this value will be static for the lifetime of the cluster,
+  however, that is a Platform implementation choice. Brokers ought to take
+  care that, while atypical, it is possible for the value to change over
+  time.
+  
+  This property MUST be a non-empty string serialized as follows:
 
   ```
   "clusterID": "id-goes-here"


profile.md
```diff
@@ -234,7 +234,14 @@ The following properties are defined for usage within a Kubernetes deployment:
   Version: 2.14
 
   The unique identifier for the Kubernetes cluster from which the request
-  was sent. This property MUST be a non-empty string serialized as follows:
+  was sent. This property can be useful when a Service Broker tries to
+  call back to a Platform that is managing multiple Kubernetes clusters.
+  Typically, this value will be static for the lifetime of the cluster,
```
pmorie: There's no way to guarantee this from an implementation standpoint - I think merging this to the spec was a mistake now.

pmorie:
I retract my previous comments on this PR - LGTM

profile.md
```diff
@@ -234,7 +234,14 @@ The following properties are defined for usage within a Kubernetes deployment:
...
+  however, that is a Platform implementation choice. Brokers ought to take
```
n3wscott: better punctuation?: Typically this value will be static for the lifetime of the cluster, however, that is a Platform implementation choice.

In general this has a lot of `can` and `will` and other feedback on other PRs has been that these words need to either be spec words or needs to be reworked to not be noisy

profile.md
```diff
@@ -234,7 +234,14 @@ The following properties are defined for usage within a Kubernetes deployment:
...
+  care that while atypical, it is possible for the value to change over
+  time.
```
n3wscott: > it is possible for the value to change over  time.

What should happen in this case? What does it mean for the cluster id to change and what happens to the peers that knew what that old id was? (and maybe tied to resources in that cluster?)

profile.md
duglin: I don't think they can be RFC2119 keywords since we're not mandating anything - just providing info. Not sure what you mean by "noisy"

profile.md
duglin: I'm inclined to leave that as an impl detail because I'm afraid that anything I say might be interpreted as telling people what they MUST do, even if I don't use MUST in the text. For example, if I was to try to say something it might be something like:
- - -
When communicating back to the Platform, Service Brokers will use the clusterID of the most recent message sent by the Platform related to the resource in the Service Broker being accessed.
- - -
But that implies they shouldn't cache it, and while I agree that would be good not to, I'm not sure its our place to say so.  But if you can think of other text I'm not against considering it.

profile.md
duglin: ok I fixed the punctuation issue

n3wscott: shoot... one more: 

> Brokers ought to take care that while atypical, it is possible for the value to change over time.

should be 

> Brokers ought to take care that, while atypical, it is possible for the value to change over time.

duglin: fixed

pmorie:
LGTM

[Approved with PullApprove](https://pullapprove.com/openservicebrokerapi/servicebroker/pull-request/480/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=openservicebrokerapi/servicebroker)

duglin:
2 needed

fmui:
LGTM

vaikas-google:
lgtm

PR #481, opened 2018-03-27
Remove language about cluster ID

pmorie:
Removes cluster ID language temporarily until we can discuss implementation in k8s further.

I have significant concerns about our ability to implement the feature as described here, and given the other discussions ongoing about the relationship of the spec to implementation concerns in the platform, I think it would be most prudent to remove this language until we can settle on a spec change and implementation in k8s that everyone can agree to.

I would like us to merge this, and explore an optional addition of the same field with weaker guarantees. I would also like to ensure that we have an optional implementation of this feature merged in the kubernetes catalog that users can try out and give us feedback, per our normal process in this spec, before we commit a permanent change.

### DIFF
--- a/profile.md
+++ b/profile.md
@@ -228,43 +228,6 @@ The following properties are defined for usage within a Kubernetes deployment:
   "namespace": "testing"
   ```
 
-
-- `clusterid`
-
-  Version: 2.14
-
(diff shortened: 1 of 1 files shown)


pmorie:
I’m concerned this doesn’t read how I intended it to at all, so I’m closing for now until I can propose specific changes.

Issue #462, opened 2018-03-01
Dashboard Client ID should be globally unique


sttomm:
The [Dashboard Client](https://github.com/openservicebrokerapi/servicebroker/blob/v2.13/spec.md#dashboard-client-object) section of the spec only says the id must be a non-empty string. I think it must be a globally unique id. The platform will create an OIDC client with that name for the service, right? For not getting conflicts between the clients of the different services, this OIDC client name must be globally unique.

n3wscott:
There is no reason I can think of that the Oauth client id needs to be globally unique. 

Here is what the spec says for that id: `The id of the Oauth client that the dashboard will use.` 

It is just the id of the Oauth client. That id could be reused for multiple services from the same broker, or even for all services from brokers provided by a vendor depending on implementation.

If you still have a question or if this did not answer your concern, please reopen this issue.

sttomm:
Thanks for you answer! I am not quite sure about the implementation of this in the platform. So the most important question from my side is, does the platform automatically create the OAuth client? Or must the client be created independently of the catalog update done by the platform? Is this a manual process, or how it is handled in Cloud Foundry or intended in general?

If the platform creates the client, how should conflicts be handled in case of different services using different passwords for the client? The first service wins and the client will be created with this password?

n3wscott:
The OSB spec says these details should be sorted out between the platform and the broker out-of-band. 

The OSB spec also does not give guidance on OAuth token lifecycles. The way current implementations are working is unique to each platform. It could be a manual process (providing your platform with a valid access token) or it could be done within the platform via refresh tokens. 

@jboyd01 was working on adding some OAuth language to the OSB spec, maybe he'd have some opinions.

I am also not clear on your usage of "password" and mixing that with the question about OAuth client.

n3wscott:
Seems like there is still some questions so I will reopen until this is sorted out.

sttomm:
Thanks for reopening it, i didn't have the rights to do so.

[...]

PR #483, opened 2018-03-28
Move dashboard_client to CF catalog extension described in profile.md

mattmcneeney:
The `dashboard_client` object is only used by service brokers that are being used with Pivotal Cloud Foundry. We should therefore investigate if it is feasible to deprecate this field (similar to what we did with `organization_guid` and `space_guid`) and move it into the service-level metadata specific for Cloud Foundry environments.

_Adding the do not merge flag as I need some feedback from the CF side before moving this into review._

### DIFF
--- a/profile.md
+++ b/profile.md
@@ -26,6 +26,8 @@ or REQUIRED level requirements defined herein.
   - [Cloud Foundry Bind Resource Object](#cloud-foundry-bind-resource-object)
 - [Service Metadata](#service-metadata)
   - [Cloud Foundry Service Metadata](#cloud-foundry-service-metadata)
+- [Catalog Extensions](#catalog-extensions)
+  - [Cloud Foundry Catalog Extensions](#cloud-foundry-catalog-extensions)
 
 ## Notations and Terminology
 
@@ -240,7 +242,7 @@ The following properties are defined for usage within a Kubernetes deployment:
   however, that is a Platform implementation choice. Brokers ought to take
   care that, while atypical, it is possible for the value to change over
   time.
-  
+
   This property MUST be a non-empty string serialized as follows:
 
   ```
@@ -428,3 +430,39 @@ specific behaviour.
 | Broker API Field | Type | Description |
 | --- | --- | --- |
 | metadata.shareable | string | Allows Service Instances to be shared across orgs and spaces. |
+
+## Catalog Extensions
+
+In addition to the fields described in the [Catalog](spec.md#catalog-management),
+Service Brokers MAY also expose the following fields to enable Platform specific behaviour.
+
+#### Cloud Foundry Catalog Extensions
+
+| Field | Type | Description |
+| --- | --- | --- |
+| services[n].dashboard_client | [DashboardClient](#dashboard-client-object) | Contains the data necessary to activate the Dashboard SSO feature for this service. |
+
+##### Dashboard Client Object
+
+| Response Field | Type | Description |
+| --- | --- | --- |
+| id | string | The id of the OAuth client that the dashboard will use. If present, MUST be a non-empty string. |
+| secret | string | A secret for the dashboard client. If present, MUST be a non-empty string. |
+| redirect_uri | string | A URI for the service dashboard. Validated by the OAuth token server when the dashboard requests a token. |
+
+#### Example Catalog
+
+```
+{
+  "services": [{
+    "name": "fake-service",
+    "id": "acb56d7c-XXXX-XXXX-XXXX-feb140a59a66",
+    "description": "A fake service.",
+    "dashboard_client": {
+      "id": "398e2f8e-XXXX-XXXX-XXXX-19a71ecbcf64",
+      "secret": "277cabb0-XXXX-XXXX-XXXX-7822c0a90e5d",
+      "redirect_uri": "http://localhost:1234"
+    }
+  }]
+}
+```
--- a/spec.md
+++ b/spec.md
@@ -396,7 +396,7 @@ It is therefore RECOMMENDED that implementations avoid such strings.
 | instances_retrievable | boolean | Specifies whether the [Fetching a Service Instance](#fetching-a-service-instance) endpoint is supported for all plans. |
 | bindings_retrievable | boolean | Specifies whether the [Fetching a Service Binding](#fetching-a-service-binding) endpoint is supported for all plans. |
 | metadata | object | An opaque object of metadata for a Service Offering. It is expected that Platforms will treat this as a blob. Note that there are [conventions](profile.md#service-metadata) in existing Service Brokers and Platforms for fields that aid in the display of catalog data. |
-| dashboard_client | [DashboardClient](#dashboard-client-object) | Contains the data necessary to activate the Dashboard SSO feature for this service. |
+| dashboard_client | [DashboardClient](profile.md#dashboard-client-object) | A Cloud Foundry extension described in [Catalog Extensions](profile.md#catalog-extensions). Contains the data necessary to activate the Dashboard SSO feature for this service. |
 | plan_updateable | boolean | Whether the service supports upgrade/downgrade for some plans. Please note that the misspelling of the attribute `plan_updatable` as `plan_updateable` was done by mistake. We have opted to keep that misspelling instead of fixing it and thus breaking backward compatibility. Defaults to false. |
 | plans* | array of [Plan](#plan-object) objects | A list of plans for this service, schema is defined below. MUST contain at least one plan. |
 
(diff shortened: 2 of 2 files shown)


mattmcneeney:
@zrob I'd appreciate your feedback on this one. We can handle this deprecation similar to what we've done for other fields in CAPI, but we will have to do a substantial outreach to broker authors to let them know that at some point in the future, the top-level field will be removed.

cc @samze

spec.md
```diff
@@ -378,7 +378,7 @@ It is therefore RECOMMENDED that implementations avoid such strings.
 | requires | array of strings | A list of permissions that the user would have to give the service, if they provision it. The only permissions currently supported are `syslog_drain`, `route_forwarding` and `volume_mount`. |
 | bindable* | boolean | Specifies whether Service Instances of the service can be bound to applications. This specifies the default for all plans of this service. Plans can override this field (see [Plan Object](#plan-object)). |
 | metadata | object | An opaque object of metadata for a Service Offering. It is expected that Platforms will treat this as a blob. Note that there are [conventions](profile.md#service-metadata) in existing Service Brokers and Platforms for fields that aid in the display of catalog data. |
-| dashboard_client | [DashboardClient](#dashboard-client-object) | Contains the data necessary to activate the Dashboard SSO feature for this service. |
+| dashboard_client | [DashboardClient](profile.md#dashboard-client-object) | Deprecated; has been moved to the [Cloud Foundry Service Metadata](profile.md#cloud-foundry-service-metadata) object. Contains the data necessary to activate the Dashboard SSO feature for this service. |
```
n3wscott: perhaps remove `Contains the data necessary to activate the Dashboard SSO feature for this service.` as well?

n3wscott:
LGTM, my only nit is to remove the description for the deprecated field so that someone goes to look at the new location inside of profile.md for more information. But I am ok as is as well...

zrob:
I agree this is CF specific and should be moved from the top-level of the spec.  I think there are a few things to address.

1. It should still be configurable per service
1. CF platform should be able to support both old and new location
1. It inherently holds sensitive information.  This means that even in the metadata section this could appear on the Service object in k8s.  So a broker that is multi-platform may be surprised at the visibility of this secret information.

I think this proposal meets the first two issues. 

The third issue does not have a clear answer since we don't have a concept of "sensitive" in the spec that a platform could care about, so I don't think we should block on this, but maybe we need to consider this as a separate concern?

Before merging we should give CF a chance to update to use the new location.

n3wscott:
> Before merging we should give CF a chance to update to use the new location.

The field is just being marked deprecated, and not removed. So CF can still go on using the field in the current location even if the PR is merged.

In the future, perhaps CF should look into a Generic Action Auth call on the broker for a service and that would fix point 3?

duglin:
@mattmcneeney this will close #462 right?

duglin:
Look to see if we can resolve #462 as part of this work

mattmcneeney:
This is now covering #462

profile.md
```diff
@@ -421,3 +421,12 @@ specific behaviour.
 | Broker API Field | Type | Description |
 | --- | --- | --- |
 | metadata.shareable | string | Allows Service Instances to be shared across orgs and spaces. |
+| metadata.dashboard_client | [DashboardClient](#dashboard-client-object) | Contains the data necessary to activate the Dashboard SSO feature for this service. |
+
+##### Dashboard Client Object
+
+| Response Field | Type | Description |
+| --- | --- | --- |
+| id | string | The id of the oAuth client that the dashboard will use. If present, MUST be a globally unique non-empty string. |
```
zrob: oAuth -> OAuth

mattmcneeney: fixed

spec.md
```diff
@@ -393,15 +393,6 @@ company). Additionally, some Platforms might modify the service names before
 presenting them to their users. This specification places no requirements on
 how Platforms might expose these values to their users.
 
-##### Dashboard Client Object
```
duglin: If its just deprecated, I'm not sure we can fully remove it from the spec yet - don't we usually keep the old fields around?

spec.md
```diff
@@ -378,7 +378,7 @@ It is therefore RECOMMENDED that implementations avoid such strings.
 | requires | array of strings | A list of permissions that the user would have to give the service, if they provision it. The only permissions currently supported are `syslog_drain`, `route_forwarding` and `volume_mount`. |
 | bindable* | boolean | Specifies whether Service Instances of the service can be bound to applications. This specifies the default for all plans of this service. Plans can override this field (see [Plan Object](#plan-object)). |
 | metadata | object | An opaque object of metadata for a Service Offering. It is expected that Platforms will treat this as a blob. Note that there are [conventions](profile.md#service-metadata) in existing Service Brokers and Platforms for fields that aid in the display of catalog data. |
-| dashboard_client | [DashboardClient](#dashboard-client-object) | Contains the data necessary to activate the Dashboard SSO feature for this service. |
+| dashboard_client | [DashboardClient](profile.md#dashboard-client-object) | Deprecated (see the [Cloud Foundry Service Metadata](profile.md#cloud-foundry-service-metadata) object). |
```
duglin: I think this needs to point to the old stuff, until its fully removed, but keep the deprecation statement and pointer to the new stuff.

duglin:
SGTM just a minor comment on the spec changes.

n3wscott:
how about adding a login call to OSBAPI?

profile.md
```diff
@@ -421,3 +421,12 @@ specific behaviour.
 | Broker API Field | Type | Description |
 | --- | --- | --- |
 | metadata.shareable | string | Allows Service Instances to be shared across orgs and spaces. |
+| metadata.dashboard_client | [DashboardClient](#dashboard-client-object) | Contains the data necessary to activate the Dashboard SSO feature for this service. |
```
pmorie: I think we should take a step back here - this doesn't seem like a great fit for metadata or the catalog endpoint in general, as-is.

In kubernetes, we take whatever's in metadata and put it into a user-visible field. AFAIK we don't have an expectation that the platform name is sent to the catalog endpoint. That means that when a broker returns this field, it will be visible to every user that can see the service resource.  That means that if you return this information, you've opened it up to a wider audience than you probably expected.

I don't wish to stall this at all, but I think we need to begin reasoning about the catalog endpoint as returning information that may be visible to everyone, even users with a low degree of privilege. Could we consider moving this to a separate resource or subresource in the future?

spec.md
mattmcneeney: Since we're still linking to it elsewhere, I'd be keen to remove this to keep things tidier. You can still find out what it is in the versioned profile.md - it's just moved. What do you think?

spec.md
mattmcneeney: See above. Since both `spec.md` and `profile.md` are versioned, can we not move things between these without 'breaking' anything?

profile.md
mattmcneeney: So in this regard, CF and k8s are behaving differently. In CF, the platform uses information in `metadata` to understand behaviours but doesn't make this information visible to the end user. We also only show some information from the catalog in general, hence this kind of information has been stored here.

Do you think this is probably a good input for the v3 design?

duglin: In general it does seem like we may need to have a clean separation between data that's ok for end users to see and data that is not ok for end users to see - and each needs to be clearly identified in the spec so that Platforms can act appropriately.  In this case, if we were to move this data into some "secretMetadata" section then Kube can know that it should do something special with it.

spec.md
duglin: My point is that existing brokers might look at the spec and see this field and, even though its deprecated, they're still allowed to send it. Which means we need to still clearly define it.  We can't leave a hole in the supported functionality.  See how we dealt with org_id - we tagged it as deprecated in favor of context but still fully defined it.

spec.md
duglin: By that logic we should remove all CF-isms from the core spec **today**.  :-)   I think each version of a spec should be complete on its own. It feels wrong to mention something that is only defined in an old version of the spec.

profile.md
Samze: Actually CF and K8s are the same here. We expose all information returned in metadata on the CF API. Historically metadata has been populated with data for client tooling. This would be the first time we have to "hide" something from the response.

If we're considering a different abstraction for broker-to-platform metadata. Do we want to move it to the current metadata object just to have to move it again later?

mattmcneeney:
Moving `dashboard_client` to metadata isn't a good solution as this field is visible to end users in both CF and k8s. However we do all agree that we shouldn't have platform-specific fields in the spec.

Next plan:
* Wait for v3, _or_
* Move this to a 'vendor extension'?

mattmcneeney:
Thoughts on this rework @duglin @zrob ?

duglin:
I can live with that.
LGTM

zrob:
This makes sense to me.  We also talked about using this as a chance to address #462.  I'm happy to go with this or see one more update to mention the unique constraint.

zrob:
LGTM

duglin:
One more little review needed

duglin:
rebase needed

mattmcneeney:
@duglin rebased

duglin:
LGTM

Issue #427, opened 2018-01-23
Proposal: Add declaration of service instance dependencies


fmui:
OSB defines how applications are connected to service instances. We have scenarios where we have to bind service instances to other service instances. A service instance A needs a service instance B.

Today, we solve this problem with service keys. When service instance A is provisioned, the service key of service instance B has to be provided as a parameter. Because there is no standard, we see several variations of this pattern. 

Is this a common problem? Do you think declaration of service instance dependencies should be part of OSB?

Please find the proposal document here: https://docs.google.com/document/d/1jBwoXRKtQqGzanhhvb-84YEGLYzqbiNbXkQ-XSeN-KA

duglin:
@fmui while there's a syntactical difference between using "parameters" and "dependencies", is there really a semantic difference?  I'm asking because it still feels like we're just passing along extra metadata just with a different tag name. And I'm wondering if it makes sense to start down the path of defining a new top-level field each time we may need to do add more data.

Also, have you considered just having the service call the platform (or broker) directly to ask for the dependent service to be created? Does it make sense to have the service manage its own dependencies via OSBAPI calls? Perhaps instead of passing in the creds to a service instance we should be passing in (and standardizing) the creds needed to invoke OSBAPI calls.  Just pondering...

krancour:
@fmui at MS, we've come to similar conclusions as you and we've begun experimenting with this in a feature branch of the [Open Service Broker for Azure](https://github.com/Azure/open-service-broker-azure).

Our approach has been very similar to what you have outlined. In our catalog, we have modeled the parent/child nature of the relationship between certain services. (Right now, we're prototyping this with SQL Server and databases _on_ the server.)

We've modified provisioning and deprovisioning in the following ways:

- Provisioning a "parent" ("container") service instance requires an `alias` parameter that can be used to uniquely identify that service instance.
- Provisioning a "child" service instance requires a `parentAlias` parameter as a reference to the service instance upon which it depends.
- On provisioning, if a child's parent is still in a provisioning state or doesn't exist yet (this is possible if parent and child are provisioned simultaneously in a Helm chart, for instance), provisioning of the child is deferred until the parent is fully provisioned.
- On deprovisioning, if a parent still has children, deprovisioning is deferred until it has no children remaining. (Deprovisioning does not cascade. It just waits until it can proceed.)

We have a lot of kinks to still work out.

[...]

PR #488, opened 2018-04-03
Service dependencies specification

fmui:
See issue #427.

First draft of the service dependencies specification.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -13,6 +13,7 @@
   - [Service Broker Errors](#service-broker-errors)
   - [Catalog Management](#catalog-management)
     - [Adding a Service Broker to the Platform](#adding-a-service-broker-to-the-platform)
+  - [Service Dependencies](#service-dependencies)
   - [Synchronous and Asynchronous Operations](#synchronous-and-asynchronous-operations)
     - [Synchronous Operations](#synchronous-operations)
     - [Asynchronous Operations](#asynchronous-operations)
@@ -422,6 +423,7 @@ how Platforms might expose these values to their users.
 | free | boolean | When false, Service Instances of this plan have a cost. The default is true. |
 | bindable | boolean | Specifies whether Service Instances of the Service Plan can be bound to applications. This field is OPTIONAL. If specified, this takes precedence over the `bindable` attribute of the service. If not specified, the default is derived from the service. |
 | schemas | [Schemas](#schemas-object) | Schema definitions for Service Instances and bindings for the plan. |
+| dependencies | [Dependencies](#dependencies) | Dependencies |
 
 \* Fields with an asterisk are REQUIRED.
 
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -592,6 +594,204 @@ After implementing the first endpoint `GET /v2/catalog` documented
...
+
+```
+{
+  "parameters": {
+    "database": {
+      "service_id": "71ef3933-8520-4f26-b808-f483d23f94a6",
+      "service_name": "mysql",
+      "plan_id": "1b367d58-6f17-4cb3-9ef8-3a69e752b826",
+      "plan_name": "small",
+      "credentials": {
```
mhrivnak: Since this contains a binding, why not call the key `binding`? The word `credentials` doesn't communicate that other things besides credentials can exist in the object (as we see in this example), which I suspect is why the word `binding` was chosen for the API.

spec.md
```diff
@@ -592,6 +594,204 @@ After implementing the first endpoint `GET /v2/catalog` documented
...
+\* Fields with an asterisk are REQUIRED.
+
+### Service Dependencies
+
+| Field | Type | Description |
+| --- | --- | --- |
+| parameter* | string | Name of the parameter. MUST be a non-empty string and MUST be unique within all dependency parameters. |
+| description* | string | A short description of the dependency. MUST be a non-empty string. |
+| optional  | boolean | Whether the dependency is optional or not. Defaults to false. |
+| service_id | string | Service ID of the instance. MUST be a non-empty string. MUST NOT be used in conjunction with the fields `plan_ids` and `service_name`. |
```
n3wscott: Why not have service_id be service_ids and allow a list?

spec.md
```diff
@@ -592,6 +594,204 @@ After implementing the first endpoint `GET /v2/catalog` documented
...
+| plan_ids | array-of-strings | An array of plan IDs. The instance must match one of those plans. MUST be a non-empty array of non-empty strings. MUST NOT be used in conjunction with the fields `service_id` and `service_name`. |
+| service_name | string | Service name of the instance. MUST be a non-empty string. MUST NOT be used in conjunction with the fields `service_id` and `plan_ids`. |
+| plan_names | array-of-strings | An array of plan names. The instance must match one of those plans. MUST be a non-empty array of non-empty strings. REQUIRES the field `service_name`. |
+
+\* Fields with an asterisk are REQUIRED.
+
+### Instance Dependencies
```
n3wscott: how would a broker present a service class/plan that has an instance dependency if that dependency did not exist? It seems to be impossible for the platform to show unless you already have the parent object.

spec.md
```diff
@@ -592,6 +594,204 @@ After implementing the first endpoint `GET /v2/catalog` documented
...
+
+| Field | Type | Description |
+| --- | --- | --- |
+| parameter* | string | Name of the parameter. MUST be a non-empty string and MUST be unique within all dependency parameters. |
```
n3wscott: field i think

spec.md
```diff
@@ -592,6 +594,204 @@ After implementing the first endpoint `GET /v2/catalog` documented
...
+| description* | string | A short description of the dependency. MUST be a non-empty string. |
```
n3wscott: both parameter and description seem to be a duplication of what would should be in the parameters for json schema, unless you are assuming this would be next to parameters, in that case it needs it's own json schema object.

spec.md
```diff
@@ -809,6 +1009,7 @@ The following HTTP Headers are defined for this operation:
 | organization_guid* | string | Deprecated in favor of `context`. The Platform GUID for the organization under which the Service Instance is to be provisioned. Although most Service Brokers will not use this field, it might be helpful for executing operations on a user's behalf. MUST be a non-empty string. |
 | space_guid* | string | Deprecated in favor of `context`. The identifier for the project space within the Platform organization. Although most Service Brokers will not use this field, it might be helpful for executing operations on a user's behalf. MUST be a non-empty string. |
 | parameters | object | Configuration options for the Service Instance. Service Brokers SHOULD ensure that the client has provided valid configuration parameters and values for the operation. |
+| alias | string | Alias for this instance. |
```
n3wscott: I think alias is suppose to be used like instance name. and I am assuming the platform allows this detail and this is a way to get the platforms name of the instance into the broker's context.

Why not call it name? or instance_name... Or leave it out and always operate on service_id, and have the platform deal with the translation.

spec.md
```diff
@@ -592,6 +594,204 @@ After implementing the first endpoint `GET /v2/catalog` documented
...
+
+### Instance Dependencies
+
+| Field | Type | Description |
+| --- | --- | --- |
+| parameter* | string | Name of the parameter. MUST be a non-empty string and MUST be unique within all dependency parameters. |
+| description* | string | A short description of the dependency. MUST be a non-empty string. |
+| optional  | boolean | Whether the dependency is optional or not. Defaults to false. |
+| instance\_id\_parameter | string | The name of the parameter that contains the service instance ID of an existing service. MUST be a non-empty string. MUST NOT be used in conjunction with the field `alias_parameter`. |
+| alias_parameter | string | The name of the parameter that contains the alias of an existing service. MUST be a non-empty string. MUST NOT be used in conjunction with the field `instance_id_parameter`. |
```
n3wscott: also field i think. Also I think it should be a path.

I don't understand why there is a spec defined location for alias in provision but it does not end up in a first class location for the instance.

spec.md
```diff
@@ -592,6 +594,204 @@ After implementing the first endpoint `GET /v2/catalog` documented
...
+\* Fields with an asterisk are REQUIRED.
+
+### Instance Dependencies
+
+| Field | Type | Description |
+| --- | --- | --- |
+| parameter* | string | Name of the parameter. MUST be a non-empty string and MUST be unique within all dependency parameters. |
+| description* | string | A short description of the dependency. MUST be a non-empty string. |
+| optional  | boolean | Whether the dependency is optional or not. Defaults to false. |
+| instance\_id\_parameter | string | The name of the parameter that contains the service instance ID of an existing service. MUST be a non-empty string. MUST NOT be used in conjunction with the field `alias_parameter`. |
```
n3wscott: also field i think. Also I think it should be a path.

spec.md
n3wscott: I am not sure I see the value in alias. I suppose it's goal is to allow you to depend on an instance called `my_sql_database` but this has no mechanism for allowing the platform to know which of the currently available service instances might be able to fit into that slot, because I can name my services anything I want...

Or is this intended to be a dynamic field that is back-channeled into the catalog and the platform must assume it needs to do a catalog fetch after every provision?

spec.md
```diff
@@ -592,6 +594,204 @@ After implementing the first endpoint `GET /v2/catalog` documented
...
+### Service Dependencies
+
+| Field | Type | Description |
+| --- | --- | --- |
+| parameter* | string | Name of the parameter. MUST be a non-empty string and MUST be unique within all dependency parameters. |
+| description* | string | A short description of the dependency. MUST be a non-empty string. |
+| optional  | boolean | Whether the dependency is optional or not. Defaults to false. |
+| service_id | string | Service ID of the instance. MUST be a non-empty string. MUST NOT be used in conjunction with the fields `plan_ids` and `service_name`. |
+| plan_ids | array-of-strings | An array of plan IDs. The instance must match one of those plans. MUST be a non-empty array of non-empty strings. MUST NOT be used in conjunction with the fields `service_id` and `service_name`. |
+| service_name | string | Service name of the instance. MUST be a non-empty string. MUST NOT be used in conjunction with the fields `service_id` and `plan_ids`. |
```
n3wscott: Should this be a list?

The spec says this:

>Additionally, some Platforms might modify the service names before presenting them to their users.

So the platform will have to also translate this name to have it make sense to the user.

spec.md
```diff
@@ -592,6 +594,204 @@ After implementing the first endpoint `GET /v2/catalog` documented
...
+Dependencies are injected as parameters when a service instance is provisioned.
+The names of the parameters are declared in the catalog.
+
+The parameter values are objects and are different for bindable and
+non-bindable services. They all have the following fields in common.
+
+| Field | Type | Description |
+| --- | --- | --- |
+| service_id* | string | Service ID of the instance. |
+| service_name* | string | Service name of the instance. |
```
n3wscott: do you really need the service_name and plan_name here? those are easy to look up from service_id, and also not tied to a service_id as updates to the service class happen, you could have a new service_id with the old service_name.

spec.md
```diff
@@ -592,6 +594,204 @@ After implementing the first endpoint `GET /v2/catalog` documented
...
+  The service instance accepts any bindings and service instance IDs and
+  evaluates them at runtime.
+* Service dependencies are specified by a service or a set of plans.
+  When the service instance is provisioned, the platform has to provide a
+  binding or a service instance ID of a matching service instance.
+* Instance dependencies declare dependencies to a specific instance either
+  by a service instance ID or by a service instance alias.
+
+Dependencies are injected as parameters when a service instance is provisioned.
+The names of the parameters are declared in the catalog.
```
n3wscott: does this mean the field that the broker will inject the data that was in the catalog is configurable? is that needed?

spec.md
```diff
@@ -592,6 +594,204 @@ After implementing the first endpoint `GET /v2/catalog` documented
...
+| plan_id* | string | Plan ID of the instance. |
+| plan_name* | string | Plan name of the instance. |
+
+\* Fields with an asterisk are REQUIRED.
+
+For bindable services, the value MUST contain be the response of a binding request. 
+
+```
+{
+  "parameters": {
```
n3wscott: I would recommend pulling out the dependency fields into their own field next to parameters, like: 
```
{
  "dependencies" {
    "service_id": "71ef3933-8520-4f26-b808-f483d23f94a6",
    "service_name": "mysql",
    "plan_id": "1b367d58-6f17-4cb3-9ef8-3a69e752b826",
    "plan_name": "small",
  },
  "parameters": {
    "database": {
      "credentials": {
        "uri": "mysql://mysqluser:pass@mysqlhost:3306/dbname",
        "username": "mysqluser",
        "password": "pass",
        "host": "mysqlhost",
        "port": 3306,
        "database": "dbname"
      }
    }
  }
}
```

spec.md
```diff
@@ -592,6 +594,204 @@ After implementing the first endpoint `GET /v2/catalog` documented
...
+| Field | Type | Description |
+| --- | --- | --- |
+| service_id* | string | Service ID of the instance. |
+| service_name* | string | Service name of the instance. |
+| plan_id* | string | Plan ID of the instance. |
+| plan_name* | string | Plan name of the instance. |
+
+\* Fields with an asterisk are REQUIRED.
+
+For bindable services, the value MUST contain be the response of a binding request.
```
n3wscott: > For bindable services, the value MUST contain be the response of a binding request. 

I am having a really hard time trying to figure out what this means. The values [in the above table>] must be contained inside the request for binding? What is required about response? I am confused because only requests have "parameters" but this is mixing the request/response demands.

spec.md
```diff
@@ -592,6 +594,204 @@ After implementing the first endpoint `GET /v2/catalog` documented
...
+      }
+    }
+  }
+}
+```
+
+Instances of non-bindable services can only be provided to the same broker.
+The platform MUST set the instance ID into the parameter value object.
+
+| Field | Type | Description |
```
n3wscott: Is this a table of allowed additive fields for non-bindable services?

spec.md
```diff
@@ -592,6 +594,204 @@ After implementing the first endpoint `GET /v2/catalog` documented
...
+      "instance_id": "0c7ec11f-977b-4027-a4d2-3dd81a9fbeb6"
+    }
+  }
+```
+
+The value of an open dependencies parameter is an array of objects, following
+the structures above. Additionally, the platform MAY add a dependency name,
+set by the user to each object.
+
+| Field | Type | Description |
```
n3wscott: Is this a table of allowed additive fields for "open deps"?

It took me a very long time to understand this was what this was attempting to say. Perhaps one table with restrictions on the fields?

spec.md
```diff
@@ -592,6 +594,204 @@ After implementing the first endpoint `GET /v2/catalog` documented
...
+| --- | --- | --- |
+| name* | string | Dependency name. |
+
+\* Fields with an asterisk are REQUIRED.
+
+```
+{
+  "parameters": {
+    "some_dependencies ": [
```
n3wscott: If this is suppose to be additive to the OSBAPI, I would recommend not mix instance parameters with broker dependencies. It is really confusing and I suspect this would be difficult to implement cleanly in broker.

n3wscott:
Overall I think this change does not do enough to allow the ecosystem to play nice together. I understand the use-case but I think this implementation is too opinionated about the solution it recommends brokers to implement. I don't see an easy way for broker to broker dependencies, such as the use case where I want to bring my broker that provisions my application, and that application depends on a database of some kind, it would be real nice to be able to say "in my application broker, please tie in a database that speaks SQL".

I think the customizable parameter targets are confusing. And in the very least they need to be paths inside the json object relative to the root of the object.

It would be interesting to use the concept of `adheres_to` when specify a generic dep. 

I think more detail needs to be added around what the platform is expected to do when it provisions something and the catalog can change with more items, or deprovision and items disappear.

I wonder if there is also a chance to add a concept of a service depending on a binding. At the moment we make bindings but we do not get to say who they are for. But a binding could be shared as well so each binding could have n parties using it. I am not sure on that point...

mhrivnak:
I agree with @n3wscott 's summary. It would be helpful to focus more on illustrating the big picture and how the key use cases are fulfilled. Reading this proposal, I found it somewhat difficult to keep track of where things fit in to a high-level workflow, and who-or-what is the actor performing each step.

It may help to focus more on addressing these general questions:

1. What exactly is the mechanism and algorithm for match-making? There are a lot of details to explore around how a Plan (or other resource maybe) would express what it requires and what it provides. It will help to first describe the data and algorithm _without talking about API operations_.
2. What component is expected to execute the match-making algorithm to determine which resource fulfills a requirement? Is it always the Platform?
3. Once some actor determines which resource fulfills a requirement, what then? What specific actions must that actor take? For example, would the Platform be responsible for creating a binding and passing it (along with additional info about the providing service?) to the service being provisioned?

krancour:
Adding an opinion here that I expressed at the f2f. I believe that solving "adheres to" is a necessary pre-requisite for some of this.

duglin:
Refresh my memory - what's the status of this? Awaiting a PR?

PR #500, opened 2018-04-17
Adding Compatibility Matrix

n3wscott:
first idea, please let me know if the group likes this direction

### DIFF
--- a/compatibility.md
+++ b/compatibility.md
@@ -2,7 +2,7 @@
 
 #### Previous releases
 | Release or Feature | Introduced | Deprecated | Cloud Foundry | Kubernetes |
-| --- | -- | --- | --- | --- |
+| --- | --- | --- | --- | --- |
 | `credentials` binding | v2.10 | | ✔️ | ✔️ |
 | `syslog_drain` binding | v2.10 | | ✔️ | - |
 | `route_forwarding` binding | v2.10 | | ✔️ | - |
--- a/compatibility.md
+++ b/compatibility.md
@@ -12,7 +12,7 @@
 | `volume_mounts` bindings | v2.13 | ✔️ | - |
 | `originating identity` header | v2.13 | ✔️ | ✔️ |
 | [*v2.13*](release-notes.md#v213) | Sep 27, 2017 | ✔️ | ✔️ |
-| `clusterid` in context | v2.14 | - | ✔️ |
+| k8s `context.clusterid` | v2.14 | n/a | ✔️ |
 | allow periods and uppercase letters in name fields | v2.14 | - | ✔️ |
 | Opaque Bearer Token Authentication | v2.14 | - | ✔️ |
 | Async Bindings | v2.14 | - | ✔️ |
--- a/compatibility.md
+++ b/compatibility.md
@@ -1,0 +1,26 @@
+# Platform Compatibility for OSBAPI
+
+
+| Release or Feature | Introduced | Cloud Foundry | Kubernetes |
+| --- | -- | --- | --- |
+| Plan Bindable | v2.11 | ✔️ | ✔️ |
+| [*v2.11*](release-notes.md#v211) | Aug 11, 2016 | ✔️ | ✔️ |
+| `context` for PUT,PATCH instance | v2.12 | ✔️ | ✔️ |
+| [*v2.12*](release-notes.md#v212) | June 13, 2017 | ✔️ | ✔️ |
+| `schemas` in catalog | v2.13 | ✔️ | ✔️ |
+| `context` for PUT binding | v2.13 | ✔️ | ✔️ |
+| `volume_mounts` bindings | v2.13 | ✔️ | - |
+| `originating identity` header | v2.13 | ✔️ | ✔️ |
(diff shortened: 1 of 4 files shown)


n3wscott:
This also can serve as a nice landing place to see future work that is happening through links in the table.

compatibility.md
```diff
@@ -0,0 +1,26 @@
...
+| Plan Bindable | v2.11 | ✔️ | ✔️ |
+| [*v2.11*](release-notes.md#v211) | Aug 11, 2016 | ✔️ | ✔️ |
+| `context` for PUT,PATCH instance | v2.12 | ✔️ | ✔️ |
+| [*v2.12*](release-notes.md#v212) | June 13, 2017 | ✔️ | ✔️ |
+| `schemas` in catalog | v2.13 | ✔️ | ✔️ |
+| `context` for PUT binding | v2.13 | ✔️ | ✔️ |
+| `volume_mounts` bindings | v2.13 | ✔️ | - |
+| `originating identity` header | v2.13 | ✔️ | ✔️ |
+| [*v2.13*](release-notes.md#v213) | Sep 27, 2017 | ✔️ | ✔️ |
+| `clusterid` in context | v2.14 | - | ✔️ |
```
duglin: This is a Kube-specific field so I wonder if its really fair to show it in the table since people might ask when (or "if") CF will support it.

duglin:
Regardless of why question, LGTM

duglin:
3 reviews needed

mattmcneeney:
@n3wscott Let's get async bindings added :)

compatibility.md
```diff
@@ -0,0 +1,26 @@
...
+| allow periods and uppercase letters in name fields | v2.14 | - | ✔️ |
+| Opaque Bearer Token Authentication | v2.14 | - | ✔️ |
+| GET endpoint for Service Instance | v2.14 | - | - |
```
Samze: CF supports both GET for Service Instances and Bindings

compatibility.md
```diff
@@ -0,0 +1,26 @@
...
+| GET endpoint for Service Binding | v2.14 | - | ✔️ |
```
Samze: CF supports GET for both Service Instances and Bindings

Samze: Thanks! Not as bad as my K8s knowledge I'm sure.

jmrodri:
LGTM

compatibility.md
duglin: the "n/a" idea you mentioned is a good one

zrob: perhaps say something like:  `clusterid` in context for Kubernetes

and then put NA for CF

duglin: would `context.clusterid` be easier/faster for people to parse?

duglin:
@n3wscott there are some open questions for you on this

duglin:
ping @n3wscott to address open questions

n3wscott:
> Not sure if we want to use that or delete it!

Delete it? I personally don't find the wiki discoverable.

compatibility.md
n3wscott: Fair, I will update.

duglin:
If we merge them, which I'm ok with, can we get all of the features from the wiki moved into this? e.g. not all bindings are mentioned

duglin:
thanks for adding in the wiki stuff
LGTM

zrob:
lgtm

fmui:
LGTM

mattmcneeney:
@n3wscott I started making a PR to your branch but accidentally just updated it... can you let me know if this looks OK to you?

duglin:
@mattmcneeney your new version almost makes it look more like a roadmap instead of a just a feature matrix. Which isn't necessarily a bad thing, but I think we might need a new title then.

One thing that's not clear to me is the order. Do the releases come before or after the features added to the release? The stuff between v2.11 an v2.12 seem to do both.

mattmcneeney:
@duglin I got that impression too (even before my accidental change), especially since we have upcoming stuff in the list. WDYT @n3wscott?

If the aim of this is to only highlight platform support of optional features, should we try to make that more explicit?

duglin:
@n3wscott an update is requested

n3wscott:
Close this?

compatibility.md
```diff
@@ -0,0 +1,32 @@
...
+| `context` for creating and updating a service instance | v2.12 | | ✔️ | ✔️ |
+| [*v2.12*](release-notes.md#v212) | June 13, 2017 | | ✔️ | ✔️ |
+| `schemas` in catalog | v2.13 | | ✔️ | ✔️ |
+| `context` for creating a service binding | v2.13 | | ✔️ | ✔️ |
+| `originating identity` header | v2.13 | | ✔️ | ✔️ |
+| Opaque Bearer Token Authentication | v2.13 | | - | ✔️ |
+| [*v2.13*](release-notes.md#v213) | Sep 27, 2017 | | ✔️ | ✔️ |
+
+
+#### Coming soon
```
mattmcneeney: @n3wscott Can you just delete this section?

PLUS add a line to the Contributing guide to update the above table on a release?

duglin:
LGTM

zrob:
lgtm

PR #311, opened 2017-09-05
Add deprecated field to services and plans with note about name changes

pmorie:
Action item from the August 2017 F2F meeting in Seattle.

Adds a `deprecated` field to the service and plan objects and adds guidance about avoiding name changes.

PR #504, opened 2018-04-18
Adding Deprecation support for Services and Plans

n3wscott:
A follow-up implementation PR to the [short proposal](https://docs.google.com/document/d/1pcA-6WFtFNspghB4FYN3Udw4t8q4M1WQJzEWamvrlC8/view#) to add deprecation support to OSBAPI.

Please review and provide feedback. I pivoted off of the[ original PR](https://github.com/openservicebrokerapi/servicebroker/pull/311) by @pmorie but I dropped the language around asking Brokers to not change service names. That can be added as guidance as a follow-up if it remains an issue.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -616,7 +616,7 @@ schema being used.
 | description | string | A short description of the deprecation. |
 | since | date string | ISO 8601 formatted date when the Service Broker deprecated the service or plan. |
 | eol | date string | ISO 8601 formatted date when the Service Broker expects to remove the service or plan from the catalog. |
-| alternatives |  array of [CatalogObjectId](#catalog-object-id-object) objects  | An array of services or plans that the Service Broker considers to be good alternatives. |
+| alternatives | array of [CatalogObjectId](#catalog-object-id-object) objects | An array of services or plans that the Service Broker considers to be good alternatives. |
 
 \* Fields with an asterisk are REQUIRED.
 
--- a/spec.md
+++ b/spec.md
@@ -614,8 +614,8 @@ schema being used.
 | Response Field | Type | Description |
 | --- | --- | --- |
 | description | string | A short description of the deprecation. |
-| since | date string | Date and time when the Service Broker deprecated the service or plan. If specified, it is expected to be an ISO 8601 formatted date and time, with time zone. |
-| eol | date string | Date and time when the Service Broker expects to end-of-life (remove) the service or plan from the catalog. If specified, it is expected to be an ISO 8601 formatted date and time, with time zone. |
+| since | date string | Date and time when the Service Broker deprecated the service or plan. If specified, this field MUST be an ISO 8601 formatted date and time, with time zone. |
+| eol | date string | Date and time when the Service Broker expects to end-of-life (remove) the service or plan from the catalog. If specified, this field MUST be an ISO 8601 formatted date and time, with time zone. |
 | alternatives | array of [CatalogObjectId](#catalog-object-id-object) objects | An array of services or plans that the Service Broker considers to be good alternatives. |
 
 \* Fields with an asterisk are REQUIRED.
--- a/spec.md
+++ b/spec.md
@@ -594,7 +594,16 @@ schema being used.
       "id": "009aba-XXXX-XXXX-XXXX-dace631cd648",
       "description": "This is deprecated.",
       "free": false,
-      "deprecation": {}
+      "deprecation": {
+        "description": "Don't use this anymore.",
+        "since": "2018-04-20",
+        "eol": "2020-02-28",
+        "alternatives": [{
+            "description": "Use this plan instead.",
+            "plan_id": "0f4008b5-XXXX-XXXX-XXXX-dace631cd648"
+          }
+        ]
+      }
     }]
   }]
 }
@@ -607,9 +616,17 @@ schema being used.
 | description | string | A short description of the deprecation. |
 | since | date string | ISO 8601 formatted date when the Service Broker deprecated the service or plan. |
 | eol | date string | ISO 8601 formatted date when the Service Broker expects to remove the service or plan from the catalog. |
-| alternatives | array of strings | An array of alternatives services or plans that should be considered from the broker. This is a list of names. If the deprecated object is a service, these are service names. If the deprecated object is a plan, these are plan names within the same service. |
+| alternatives |  array of [CatalogObjectId](#catalog-object-id-object) objects  | An array of services or plans that the Service Broker considers to be good alternatives. |
 
 \* Fields with an asterisk are REQUIRED.
+
+##### Catalog Object Id Object 
+
+| Response Field | Type | Description |
+| --- | --- | --- |
+| description | string | A short description to provide contextual information about this catalog object reference. |
+| service_id | string | If present, it MUST be the ID of a service. |
+| plan_id | string | If present, it MUST be the ID of a plan. |
 
 ### Adding a Service Broker to the Platform
 
--- a/spec.md
+++ b/spec.md
@@ -648,8 +648,8 @@ Service Brokers MUST allow existing Service Instances to be updated and
 bindable as they were before depreciation. Deprecation MUST NOT effect the
 usability of existing Service Instances and Bindings.
 
-The Service Broker SHOULD allow the Platform to continue to
-provision new Service Instances of the deprecated service or plan.
+The Service Broker MUST allow the Platform to continue to provision new Service
+Instances of the deprecated service or plan.
 
 ## Synchronous and Asynchronous Operations
 
--- a/spec.md
+++ b/spec.md
@@ -596,8 +596,6 @@ schema being used.
       "free": false,
       "deprecation": {
         "description": "Don't use this anymore.",
-        "since": "2018-04-20T17:00:00+00:00",
-        "eol": "2020-02-28T08:00:00+00:00",
         "alternatives": [{
             "description": "Use this plan instead.",
             "plan_id": "0f4008b5-XXXX-XXXX-XXXX-dace631cd648"
@@ -613,10 +611,8 @@ schema being used.
 
 | Response Field | Type | Description |
 | --- | --- | --- |
-| description | string | A short description of the deprecation. |
-| since | date string | Date and time when the Service Broker deprecated the service or plan. If specified, this field MUST be an ISO 8601 formatted date and time, with time zone. |
-| eol | date string | Date and time when the Service Broker expects to end-of-life (remove) the service or plan from the catalog. If specified, this field MUST be an ISO 8601 formatted date and time, with time zone. |
-| alternatives | array of [CatalogObjectId](#catalog-object-id-object) objects | An array of services or plans that the Service Broker considers to be good alternatives. |
+| description* | string | A short description of the deprecation. MUST be a non-empty string. |
+| alternatives | array of [CatalogObjectId](#catalog-object-id-object) objects | An array of services or plans that the Service Broker considers to be good alternatives. If present, MUST contain at least one alternative. |
 
 \* Fields with an asterisk are REQUIRED.
 
@@ -625,8 +621,8 @@ schema being used.
 | Response Field | Type | Description |
 | --- | --- | --- |
 | description | string | A short description to provide contextual information about this catalog object reference. |
-| service_id | string | If present, it MUST be the ID of a service. |
-| plan_id | string | If present, it MUST be the ID of a plan. |
+| service_id | string | If present, it MUST be the ID of a service. If omitted, `plan_id` MUST be present. |
+| plan_id | string | If present, it MUST be the ID of a plan. If omitted, `service_id` MUST be present. |
 
 \* Fields with an asterisk are REQUIRED.
 
--- a/spec.md
+++ b/spec.md
@@ -596,8 +596,8 @@ schema being used.
       "free": false,
       "deprecation": {
         "description": "Don't use this anymore.",
-        "since": "2018-04-20",
-        "eol": "2020-02-28",
+        "since": "2018-04-20T17:00:00+00:00",
+        "eol": "2020-02-28T08:00:00+00:00",
         "alternatives": [{
             "description": "Use this plan instead.",
             "plan_id": "0f4008b5-XXXX-XXXX-XXXX-dace631cd648"
@@ -614,8 +614,8 @@ schema being used.
 | Response Field | Type | Description |
 | --- | --- | --- |
 | description | string | A short description of the deprecation. |
-| since | date string | ISO 8601 formatted date when the Service Broker deprecated the service or plan. |
-| eol | date string | ISO 8601 formatted date when the Service Broker expects to remove the service or plan from the catalog. |
+| since | date string | Date and time when the Service Broker deprecated the service or plan. If specified, it is expected to be an ISO 8601 formatted date and time, with time zone. |
+| eol | date string | Date and time when the Service Broker expects to end-of-life (remove) the service or plan from the catalog. If specified, it is expected to be an ISO 8601 formatted date and time, with time zone. |
 | alternatives | array of [CatalogObjectId](#catalog-object-id-object) objects | An array of services or plans that the Service Broker considers to be good alternatives. |
 
 \* Fields with an asterisk are REQUIRED.
--- a/spec.md
+++ b/spec.md
@@ -429,7 +429,7 @@ how Platforms might expose these values to their users.
 | free | boolean | When false, Service Instances of this plan have a cost. The default is true. |
 | bindable | boolean | Specifies whether Service Instances of the Service Plan can be bound to applications. This field is OPTIONAL. If specified, this takes precedence over the `bindable` attribute of the service. If not specified, the default is derived from the service. |
 | schemas | [Schemas](#schemas-object) | Schema definitions for Service Instances and bindings for the plan. |
-| deprecation | [Deprecation](#deprecation-object) | Contains details of the deprecation status of the plan. This field is OPTIONAL. If ommited, the plan is not deprecated. See [deprecating services and plans](#deprecating-services-and-plans) for more information. |
+| deprecation | [Deprecation](#deprecation-object) | Contains details of the deprecation status of the plan. This field is OPTIONAL. If omitted, the plan is not deprecated. See [deprecating services and plans](#deprecating-services-and-plans) for more information. |
 
 \* Fields with an asterisk are REQUIRED.
 
@@ -599,11 +599,10 @@ schema being used.
 
 | Response Field | Type | Description |
 | --- | --- | --- |
-| deprecated* | boolean | Indicates that this service or plan is deprecated and will be removed in the future. See [deprecating services and plans](#deprecating-services-and-plans) for more information. |
 | description | string | A short description of the deprecation. |
 | since | date string | ISO 8601 formatted date when the Service Broker deprecated the service or plan. |
 | eol | date string | ISO 8601 formatted date when the Service Broker expects to remove the service or plan from the catalog. |
-| alterntives | array of strings | An array of alterntives services or plans that should be concidered from the broker. This is a list of names. If the debrecated object is a service, these are service names. If the debrecated object is a plan, these are plan names within the same service. |
+| alternatives | array of strings | An array of alternatives services or plans that should be considered from the broker. This is a list of names. If the deprecated object is a service, these are service names. If the deprecated object is a plan, these are plan names within the same service. |
 
 \* Fields with an asterisk are REQUIRED.
 
--- a/spec.md
+++ b/spec.md
@@ -637,7 +637,7 @@ with your Platform to make your services and plans available to end users.
 ### Deprecating services and plans
 
 A Service Broker can signal to a Platform (through the Catalog) that a service
-or plan is deprecated. The Service Broker MAY provide optional messaging and
+or plan is deprecated. The Service Broker MAY provide messaging and
 alternatives for the deprecated service or plan. This gives the Platform the
 opportunity to display meaningful advice to their users to migrate away from
 the deprecated service or plan.
--- a/spec.md
+++ b/spec.md
@@ -292,7 +292,6 @@ use these error codes for the specified failure scenarios.
 | AsyncRequired | This request requires client support for asynchronous service operations. | The query parameter `accepts_incomplete=true` MUST be included the request. |
 | ConcurrencyError | The Service Broker does not support concurrent requests that mutate the same resource. | Clients MUST wait until pending requests have completed for the specified resources. |
 | RequiresApp | The request body is missing the `app_guid` field. | The `app_guid` MUST be included in the request body. |
-| DeprecatedError | The Service Broker does not support provisioning (or updating to) a deprecated service or plan. | Another service or plan should be specified. |
 
 ## Catalog Management
 
@@ -396,7 +395,7 @@ It is therefore RECOMMENDED that implementations avoid such strings.
 | dashboard_client | [DashboardClient](#dashboard-client-object) | Contains the data necessary to activate the Dashboard SSO feature for this service. |
 | plan_updateable | boolean | Whether the service supports upgrade/downgrade for some plans. Please note that the misspelling of the attribute `plan_updatable` as `plan_updateable` was done by mistake. We have opted to keep that misspelling instead of fixing it and thus breaking backward compatibility. Defaults to false. |
 | plans* | array of [Plan](#plan-object) objects | A list of plans for this service, schema is defined below. MUST contain at least one plan. |
-| deprecation | [Deprecation](#deprecation-object) | Contains details of the deprecation status of the service. This field is OPTIONAL. If omitted, the service is not deprecated. See [deprecating services and plans](#deprecating-services-and-plans) for more information. |
+| deprecation | [Deprecation](#deprecation-object) | Contains details of the deprecation status of the service. This field is OPTIONAL. If specified, the service is deprecated. See [deprecating services and plans](#deprecating-services-and-plans) for more information. |
 
 \* Fields with an asterisk are REQUIRED.
 
@@ -429,7 +428,7 @@ how Platforms might expose these values to their users.
 | free | boolean | When false, Service Instances of this plan have a cost. The default is true. |
 | bindable | boolean | Specifies whether Service Instances of the Service Plan can be bound to applications. This field is OPTIONAL. If specified, this takes precedence over the `bindable` attribute of the service. If not specified, the default is derived from the service. |
 | schemas | [Schemas](#schemas-object) | Schema definitions for Service Instances and bindings for the plan. |
-| deprecation | [Deprecation](#deprecation-object) | Contains details of the deprecation status of the plan. This field is OPTIONAL. If omitted, the plan is not deprecated. See [deprecating services and plans](#deprecating-services-and-plans) for more information. |
+| deprecation | [Deprecation](#deprecation-object) | Contains details of the deprecation status of the plan. This field is OPTIONAL. If specified, the plan is deprecated. See [deprecating services and plans](#deprecating-services-and-plans) for more information. |
 
 \* Fields with an asterisk are REQUIRED.
 
@@ -590,6 +589,12 @@ schema being used.
           "40 concurrent connections"
         ]
       }
+    }, {
+      "name": "fake-deprecated-plan",
+      "id": "009aba-XXXX-XXXX-XXXX-dace631cd648",
+      "description": "This is deprecated.",
+      "free": false,
+      "deprecation": {}
     }]
   }]
 }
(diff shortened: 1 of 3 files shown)


spec.md
```diff
@@ -591,13 +595,41 @@ schema being used.
...
 
+##### Deprecation Object 
+
+| Response Field | Type | Description |
+| --- | --- | --- |
+| deprecated* | boolean | Indicates that this service or plan is deprecated and will be removed in the future. See [deprecating services and plans](#deprecating-services-and-plans) for more information. |
+| description | string | A short description of the deprecation. |
+| since | date string | ISO 8601 formatted date when the Service Broker deprecated the service or plan. |
+| eol | date string | ISO 8601 formatted date when the Service Broker expects to remove the service or plan from the catalog. |
+| alterntives | array of strings | An array of alterntives services or plans that should be concidered from the broker. This is a list of names. If the debrecated object is a service, these are service names. If the debrecated object is a plan, these are plan names within the same service. |
```
martinmaly: consider supporting a tuple of (alternative service, alternative plan) for deprecated plan.

spec.md
```diff
@@ -591,13 +595,41 @@ schema being used.
 }
 ```
 
+##### Deprecation Object 
+
+| --- | --- | --- |
+| deprecated* | boolean | Indicates that this service or plan is deprecated and will be removed in the future. See [deprecating services and plans](#deprecating-services-and-plans) for more information. |
```
martinmaly: there is a redundancy between presence of `deprecation` field in the parent object, and this boolean value.
Another redundancy is the since and eol timestamps.

kibbles-n-bytes: I think we should remove this required boolean, and have the presence of the object be the flag that it's deprecated. Makes things a bit cleaner, since this boolean will always be true (otherwise they wouldn't include this object in the first place).

The broker can pass back an empty deprecation object if they truly don't want to include any of the other information... But they should include it. :stuck_out_tongue:

spec.md
```diff
@@ -591,13 +595,41 @@ schema being used.
...
+opportunity to display meaningful advice to their users to migrate away from
+the deprecated service or plan.
+
+Service Brokers MUST allow existing Service Instances to be updated as they
+were before deprecration. Service Brokers MAY block provisioning of a Service
+Instance using a deprecated service or plan, or block updating an existing
+Service Instance to a deprecated plan by return error code `DeprecatedError`
+(see [Service Broker Errors](#service-broker-errors)).
+
+The Service Broker SHOULD allow the Platform to continue to
```
duglin: The "SHOULD allow" here and the "MAY block", while not technically in conflict because both allow a choice, as a reader they feel contradictory.  I think what you were trying to say is:
- brokers SHOULD allow platforms to continue to provision new instances using the old stuff
- platforms SHOULD encourage their users to migrate away from deprecated services/plans
- but if brokers do block it then they MUST use the DeprecatedError

Right?

Also, we should talk about Bindings too. E.g. People can still create new Bindings for deprecated services/plans.

spec.md
```diff
@@ -591,13 +595,41 @@ schema being used.
...
 
+### Deprecating services and plans
+
+A Service Broker can signal to a Platform (through the Catalog) that a service
+or plan is deprecated. The Service Broker MAY provide optional messaging and
+alterntives for the deprecated service or plan. This gives the Platform the
+opportunity to display meaningful advice to their users to migrate away from
+the deprecated service or plan.
+
+Service Brokers MUST allow existing Service Instances to be updated as they
```
duglin: not just "updated" but also existing Instances and Bindings of those services/plans MUST still be useable.

duglin:
Overall I like the direction. I agree with other's comments.

spec.md
n3wscott: We decided to remove the error. If the platform wants to enforce this, then they can, it was decided that it is not the role of the broker to error on deprecated provisionings, if that is the case, just delete the service or plan.

spec.md
n3wscott: I wanted to allow a broker to say "yo, this was deprecated a year ago, we gave you time."

spec.md
```diff
@@ -394,6 +395,7 @@ It is therefore RECOMMENDED that implementations avoid such strings.
 | dashboard_client | [DashboardClient](#dashboard-client-object) | Contains the data necessary to activate the Dashboard SSO feature for this service. |
 | plan_updateable | boolean | Whether the service supports upgrade/downgrade for some plans. Please note that the misspelling of the attribute `plan_updatable` as `plan_updateable` was done by mistake. We have opted to keep that misspelling instead of fixing it and thus breaking backward compatibility. Defaults to false. |
 | plans* | array of [Plan](#plan-object) objects | A list of plans for this service, schema is defined below. MUST contain at least one plan. |
+| deprecation | [Deprecation](#deprecation-object) | Contains details of the deprecation status of the service. This field is OPTIONAL. If specified, the service is deprecated. See [deprecating services and plans](#deprecating-services-and-plans) for more information. |
```
duglin: remove "This field is OPTIONAL." .  The lack of a `*` tells people that. None of the other optional ones say that.

spec.md
```diff
@@ -426,6 +428,7 @@ how Platforms might expose these values to their users.
 | free | boolean | When false, Service Instances of this plan have a cost. The default is true. |
 | bindable | boolean | Specifies whether Service Instances of the Service Plan can be bound to applications. This field is OPTIONAL. If specified, this takes precedence over the `bindable` attribute of the service. If not specified, the default is derived from the service. |
 | schemas | [Schemas](#schemas-object) | Schema definitions for Service Instances and bindings for the plan. |
+| deprecation | [Deprecation](#deprecation-object) | Contains details of the deprecation status of the plan. This field is OPTIONAL. If specified, the plan is deprecated. See [deprecating services and plans](#deprecating-services-and-plans) for more information. |
```
duglin: ditto: remove "This field is OPTIONAL."

duglin: oh man we do say it on 'bindable".  Geez, why? we should remove that.

spec.md
```diff
@@ -586,18 +589,66 @@ schema being used.
...
   }]
 }
 ```
 
+##### Deprecation Object 
+
+| --- | --- | --- |
+| description | string | A short description of the deprecation. |
+| since | date string | ISO 8601 formatted date when the Service Broker deprecated the service or plan. |
```
duglin: Do we need to mandate it include the full date, meaning timezone? Otherwise we don't know the exact time .

spec.md
```diff
@@ -586,18 +589,66 @@ schema being used.
...
+| eol | date string | ISO 8601 formatted date when the Service Broker expects to remove the service or plan from the catalog. |
```
duglin: either say what "eol" means (end of life), or say it in words "removeDate". I kind of like being explicit, so we should change "since"  to "deprecratedDate".

spec.md
duglin: some of these are required, I assume, so we should add `*` and the legend at the end.

spec.md
```diff
@@ -586,18 +589,66 @@ schema being used.
...
+
+\* Fields with an asterisk are REQUIRED.
+
+##### Catalog Object Id Object 
+
+| Response Field | Type | Description |
+| --- | --- | --- |
+| description | string | A short description to provide contextual information about this catalog object reference. |
+| service_id | string | If present, it MUST be the ID of a service. |
+| plan_id | string | If present, it MUST be the ID of a plan. |
```
duglin: ditto on missing `*` stuff

spec.md
n3wscott: is there a better way to say this? like ISO 8601 with time and timezone?

spec.md
n3wscott: gone.

spec.md
n3wscott: missing? missing `\* Fields with an asterisk are REQUIRED.` text you mean?

duglin: yea

spec.md
duglin: Maybe: An ISO 8601 formatted date and time, with time zone, when the Service Broker...
?

spec.md
n3wscott: :( I kinda like `eol` and `since`, they have context because of the object they are apart of. I thought these were pretty standard usage from my past. defining eol is fine.

spec.md
n3wscott: fixed.

spec.md
n3wscott: works for me, thanks!

spec.md
```diff
@@ -586,18 +589,68 @@ schema being used.
...
+alternatives for the deprecated service or plan. This gives the Platform the
+opportunity to display meaningful advice to their users to migrate away from
+the deprecated service or plan.
+
+Service Brokers MUST allow existing Service Instances to be updated and
+bindable as they were before depreciation. Deprecation MUST NOT effect the
+usability of existing Service Instances and Bindings.
+
+The Service Broker SHOULD allow the Platform to continue to
+provision new Service Instances of the deprecated service or plan.
```
krancour: Could I propose relaxing this from SHOULD to MAY? I believe, in _many_ cases, a service or plan will be deprecated precisely because the thing it provides _cannot_ be provisioned any longer.

spec.md
```diff
@@ -586,18 +589,68 @@ schema being used.
...
 ```
 
+##### Deprecation Object 
+
+| --- | --- | --- |
+| description | string | A short description of the deprecation. |
+| since | date string | Date and time when the Service Broker deprecated the service or plan. If specified, it is expected to be an ISO 8601 formatted date and time, with time zone. |
+| eol | date string | Date and time when the Service Broker expects to end-of-life (remove) the service or plan from the catalog. If specified, it is expected to be an ISO 8601 formatted date and time, with time zone. |
+| alternatives | array of [CatalogObjectId](#catalog-object-id-object) objects | An array of services or plans that the Service Broker considers to be good alternatives. |
```
krancour: Can I suggest that this deprecation object should have an additional required field that enumerates which operations are (or aren't) still supported for the moment? It would be useful for platforms and end users alike to understand, for instance, that the service or plan in question can still be deprovisioned or unbound, but not provisioned or bound.

(I know this PR proposes binding MUST still be supported for a deprecated service or plan, but I have a comment to make about that as well.)

spec.md
```diff
@@ -586,18 +589,68 @@ schema being used.
...
+
+A Service Broker can signal to a Platform (through the Catalog) that a service
+or plan is deprecated. The Service Broker MAY provide messaging and
+alternatives for the deprecated service or plan. This gives the Platform the
+opportunity to display meaningful advice to their users to migrate away from
+the deprecated service or plan.
+
+Service Brokers MUST allow existing Service Instances to be updated and
+bindable as they were before depreciation. Deprecation MUST NOT effect the
+usability of existing Service Instances and Bindings.
```
krancour: > Service Brokers MUST allow existing Service Instances to be updated and bindable as they were before depreciation.

I'm concerned that situations will arise very quickly where this is not possible. Could we relax this to SHOULD?

> Deprecation MUST NOT effect the usability of existing Service Instances and Bindings.

I do, however, agree with that point.

spec.md
n3wscott: I had original be thinking down this line but we (kibbles and I) had a debate and the winning factor was how deprecated is used in software, you can set your compile to throw errors if you use a deprecated method, but the API does not get to say. The platform can add restrictions, etc. But the brokers role is just to say "yo, this will go away soon"

What you want should happen after the broker has removed the plan or class from the catalog and eol has elapsed. 

So to be clear, deprecated means nothing has changed about this plan/class but change might happen soon, and the platform can help the user out by warning them.

spec.md
n3wscott: If the broker no longer supports the class/plan they should remove it from the catalog, and send a nasty note to their users about their dumb faces.

spec.md
n3wscott: In that case, it is not a deprecation but a removal. Deprecation is to be used as an off-ramp, not an off-cliff. eol vs sol

n3wscott:
Double checking, please give this comment a 👍 if the api looks ok and I will swagger/openapi it

spec.md
```diff
@@ -586,18 +589,68 @@ schema being used.
...
   }]
 }
 ```
 
+##### Deprecation Object 
+
+| --- | --- | --- |
+| description | string | A short description of the deprecation. |
+| since | date string | Date and time when the Service Broker deprecated the service or plan. If specified, it is expected to be an ISO 8601 formatted date and time, with time zone. |
```
duglin: s/it is expected to be/it MUST be/

spec.md
```diff
@@ -586,18 +589,68 @@ schema being used.
...
+| eol | date string | Date and time when the Service Broker expects to end-of-life (remove) the service or plan from the catalog. If specified, it is expected to be an ISO 8601 formatted date and time, with time zone. |
```
duglin: MUST be an ISO...

spec.md
duglin: Its odd that none of these fields are required. Was that intentional? Seems like at least one of them should be - if not both 'since' and 'eol'. Even if the 'eol' is 2100, at least that gives people a sense of how quickly they need to move.

spec.md
duglin: If this field is present then it MUST contain at least one entry.

spec.md
```diff
@@ -586,18 +589,68 @@ schema being used.
...
+| alternatives | array of [CatalogObjectId](#catalog-object-id-object) objects | An array of services or plans that the Service Broker considers to be good alternatives. |
+
+\* Fields with an asterisk are REQUIRED.
+
+##### Catalog Object Id Object 
+
+| Response Field | Type | Description |
```
duglin: Something should be required in this. No idea what it would mean to have an empty one.

n3wscott: I thought of this, it could be that the broker just wants to pass a description that just says "no alternative. go away"

duglin: no alternative is ok, but we should have something in here, even a start date. If the entire struct is empty then why have it at all. And if we don't have it then the referencing property is empty - and that just seems silly.

spec.md
```diff
@@ -586,18 +589,68 @@ schema being used.
...
+or plan is deprecated. The Service Broker MAY provide messaging and
+alternatives for the deprecated service or plan. This gives the Platform the
+opportunity to display meaningful advice to their users to migrate away from
+the deprecated service or plan.
+
+Service Brokers MUST allow existing Service Instances to be updated and
+bindable as they were before depreciation. Deprecation MUST NOT effect the
+usability of existing Service Instances and Bindings.
+
+The Service Broker SHOULD allow the Platform to continue to
```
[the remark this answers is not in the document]
n3wscott: This was leftover from the time long ago when I thought it was a good idea to let the broker prevent provisioning. That time has passed and will update this to MUST. Thanks

spec.md
n3wscott: If the service/plan is just deprecated, then `deprecation` should be an empty object. The existence of `deprecation` on the plan or class means it is deprecated. It is up to the broker to give more hints.

spec.md
duglin: yea, because then we enter the rathole of adding text to makes sure people don't do what they might naturally want to do like add `omitempty` on things.  In many cases empty == nil == not there. Trying to draw a distinction between those, while not impossible, is more trouble than its worth. I don't think its a burden to ask for at least one of these fields to be there.

n3wscott: I disagree, I think it is fine to have something like this:

```
{
      "name": "fake-deprecated-plan",
      "id": "009aba-XXXX-XXXX-XXXX-dace631cd648",
      "description": "This is deprecated.",
      "free": false,
      "deprecation": {}
        ]
      }
    }
```

krancour:
I've made some comments in-line with the diffs, and @n3wscott, you've replied to some of those, and I appreciate the thoughtful responses. I'm moving the conversation out of the diffs and into the general comments on the PR because changes unrelated to my comments are making my comments appear outdated and hidden.

Respectfully, I object quite strongly to the direction here. I do so on the following grounds:

This _requires_ deprecated plans to continue functioning normally _until_ they are EOL'ed. Specifically, it is required that one can still provision instances of those plans and bind to instances of those plans. I've raised that it might no longer be _possible_ to provision instances of a given plan or bind to said instances and it's been dismissed with the suggestion that if that is the case, the plan in question is not merely deprecated-- it is EOL already. That's fine, but it means we're treating the sum of all actions that can be taken in relation to a plan as an all-or-nothing proposition. You either can do _all_ of provisioning, binding, unbinding, and deprovisioning, or you can do _none_ of them (because the plan no longer exists). imho, this doesn't comport with reality. I propose that there will be frequently occurring scenarios where the underlying service provider that a broker utilizes (e.g. Azure, GCP, AWS, etc.) can no longer provision a certain thing, but if instances of that certain thing already exist, they _can_ deprovisioned.

As written, this PR makes no concession for that very realistic possibility. The result, is broker authors are put between a rock and a hard place. They can deprecate a plan and let provisioning (and maybe binding) fail, but rest easy knowing deprovisioning (and maybe unbinding) still work _OR_ they can pre-maturely EOL the plan to eliminate the possibility of failed provisioning, but in doing so, existing instances of said plan become orphaned with no recourse for deprovisioning except for manual intervention by a privileged operator.

n3wscott:
First, let's agree on a definition of deprecation. I am using the following, taken from [wikipedia](https://en.wikipedia.org/wiki/Deprecation): 

>deprecation is the discouragement of use of some terminology, feature, design, or practice, typically because it has been superseded or is no longer considered efficient or safe, without completely removing it or prohibiting its use. It can also imply that a feature, design, or practice will be removed or discontinued entirely in the future.

And specifically I am highlighting this part: `without completely removing it or prohibiting its use`


> This requires deprecated plans to continue functioning normally until they are EOL'ed. Specifically, it is required that one can still provision instances of those plans and bind to instances of those plans.

I think this agrees with the definition of deprecated.

> I've raised that it might no longer be possible to provision instances of a given plan or bind to said instances and it's been dismissed with the suggestion that if that is the case, the plan in question is not merely deprecated-- it is EOL already. 

Agreed.

> it means we're treating the sum of all actions that can be taken in relation to a plan as an all-or-nothing proposition. You either can do all of provisioning, binding, unbinding, and deprovisioning, or you can do none of them (because the plan no longer exists).

I also think this agrees with the definition of deprecation.

> this doesn't comport with reality. I propose that there will be frequently occurring scenarios where the underlying service provider that a broker utilizes (e.g. Azure, GCP, AWS, etc.) can no longer provision a certain thing, but if instances of that certain thing already exist, they can deprovision them.

This case is already supported today. The purpose for adding the deprecation field is to signal to new catalog listings that there is a reason to not used this service/plan and then allow the platform to block it or call it out explicitly. 

In the case where you already have an instance, you can always deprovision it regardless of the service/plan in the current catalog. It becomes brokers choice if you want to also allow new bindings to it, but that is always the case, the broker can always prevent or block actions.

> As written, this PR makes no concession for that very realistic possibility. The result, is broker authors are put between a rock and a hard place. They can deprecate a plan and let provisioning (and maybe binding) fail, but rest easy knowing deprovisioning (and maybe unbinding) still work

This case the Broker should remove the service and plan from the catalog, they no longer support it. The broker made the choice to not provide an off-ramp for the users and as a result it will be a bad experience for their customers.

> OR they can pre-maturely EOL the plan to eliminate the possibility of failed provisioning, but in doing so, existing instances of said plan become orphaned with no recourse for deprovisioning except for manual intervention by a privileged operator.

Deprovisioning is not dependent on the service or plan being in the catalog so this logic leap is not true. For example, k8s will mark a plan as being deleted by the broker but the catalog object stays around until nothing references it.

---------------------------

Reading your concerns I think you would like a second concept to signal service degradation on the service or plan level, which is not the same concept as deprecation.  I have not given this topic any thought and it is not immediately clear how changes to the catalog would be expected to filter back to a instance of a degraded service/plan.

krancour:
>> this doesn't comport with reality. I propose that there will be frequently occurring scenarios where the underlying service provider that a broker utilizes (e.g. Azure, GCP, AWS, etc.) can no longer provision a certain thing, but if instances of that certain thing already exist, they can deprovision them.

> This case is already supported today.

> In the case where you already have an instance, you can always deprovision it regardless of the service/plan in the current catalog.

> Deprovisioning is not dependent on the service or plan being in the catalog so this logic leap is not true. For example, k8s will mark a plan as being deleted by the broker but the catalog object stays around until nothing references it.

That wasn't clear to me. Thank you for pointing that out. I'm pleased to know that is how k8s service catalog deals with this scenario. I am a little bit concerned that brokers are at the mercy of individual platforms to each do the right thing, but that's a very minor concern in comparison to the one I'd been trying to express.

@n3wscott thank you for taking the time to address my concerns. I obviously withdraw my objections.

> Reading your concerns I think you would like a second concept to signal service degradation on the service or plan level, which is not the same concept as deprecation.

Essentially yes, that is what I'd been arguing in favor of. As a separate concept, it might not be a bad idea, but speaking for myself, as long as I have a way to keep _deprovisioning_ instances of a given plan while no longer _provisioning_  instances of the same plan-- and it seems I do-- then I'm all set.

Thanks again.

spec.md
n3wscott: I have made description be required and non empty string. So lazy broker will do this:

```
{
      "name": "fake-deprecated-plan",
      "id": "009aba-XXXX-XXXX-XXXX-dace631cd648",
      "description": "This is deprecated.",
      "free": false,
      "deprecation": {
         "description": "deprecated"
      }
}
```

n3wscott: Also added MUST around service_id or plan_id is needed for the alternatives object to be valid.

n3wscott:
I have removed `eol` and `since`, and made description on `depreciation` required. Also make sure that for alternatives there has to be at least `plan_id` OR `service_id` provided. This is MVP+the idea of alternatives. eol and since were nice ideas and maybe useful but lacked a solid usecase for them so I cut them.

n3wscott:
Added Swagger and OpenApi changes.

krancour:
> eol and since were nice ideas and maybe useful but lacked a solid usecase for them so I cut them.

Would you expect that if one wanted to furnish this information, that they would simply include it in the description?

n3wscott:
>>eol and since were nice ideas and maybe useful but lacked a solid usecase for them so I cut them.

>Would you expect that if one wanted to furnish this information, that they would simply include it in the description?

I would expect them to either leverage the description, or use a vendor extension. If brokers start using eol and since to programmatically provide that data, and platforms find it useful we can bring it back into the spec. I was advised to keep the spec as light weight as possible and I did not have a direct need for these yet.

spec.md
```diff
@@ -586,18 +589,64 @@ schema being used.
...
 
 ### Adding a Service Broker to the Platform
 
 After implementing the first endpoint `GET /v2/catalog` documented
 [above](#catalog-management), the Service Broker will need to be registered
 with your Platform to make your services and plans available to end users.
 
+### Deprecating services and plans
+
+A Service Broker can signal to a Platform (through the Catalog) that a service
```
mattmcneeney: ~service~ Service Offering
~plan~ Service Plan

spec.md
```diff
@@ -586,18 +589,64 @@ schema being used.
...
+or plan is deprecated. The Service Broker MAY provide messaging and
+alternatives for the deprecated service or plan. This gives the Platform the
+opportunity to display meaningful advice to their users to migrate away from
+the deprecated service or plan.
+
+Service Brokers MUST allow existing Service Instances to be updated and
+bindable as they were before depreciation. Deprecation MUST NOT effect the
+usability of existing Service Instances and Bindings.
```
mattmcneeney: ~Bindings~ Service Bindings

spec.md
mattmcneeney: (same again below)

spec.md
mattmcneeney: So we are saying service brokers MUST provide alternatives? What if they are actually deprecating a service entirely?

spec.md
```diff
@@ -410,7 +412,7 @@ how Platforms might expose these values to their users.
 
 | Response Field | Type | Description |
 | --- | --- | --- |
-| id | string | The id of the Oauth client that the dashboard will use. If present, MUST be a non-empty string. |
+| id | string | The id of the OAuth client that the dashboard will use. If present, MUST be a non-empty string. |
```
pmorie: I would take changes that are unrelated to this PR and make them into different PRs, or they'll be waiting in limbo for this to merge.

spec.md
```diff
@@ -424,8 +426,9 @@ how Platforms might expose these values to their users.
 | description* | string | A short description of the plan. MUST be a non-empty string. |
 | metadata | object | An opaque object of metadata for a Service Plan. It is expected that Platforms will treat this as a blob. Note that there are [conventions](profile.md#service-metadata) in existing Service Brokers and Platforms for fields that aid in the display of catalog data. |
 | free | boolean | When false, Service Instances of this plan have a cost. The default is true. |
-| bindable | boolean | Specifies whether Service Instances of the Service Plan can be bound to applications. This field is OPTIONAL. If specified, this takes precedence over the `bindable` attribute of the service. If not specified, the default is derived from the service. |
+| bindable | boolean | Specifies whether Service Instances of the Service Plan can be bound to applications. If specified, this takes precedence over the `bindable` attribute of the service. If not specified, the default is derived from the service. |
```
pmorie: Limbo

spec.md
```diff
@@ -635,7 +684,7 @@ update, and deprovision.
 
 For a Service Broker to return an asynchronous response, the query parameter
 `accepts_incomplete=true` MUST be included the request. If the parameter is not
-included or is set to `false`, and the Service Broker cannot fulfil the request
+included or is set to `false`, and the Service Broker cannot fulfill the request
```
pmorie: limbo

spec.md
```diff
@@ -586,18 +589,64 @@ schema being used.
...
   }]
 }
 ```
 
+##### Deprecation Object 
+
+| --- | --- | --- |
+| description* | string | A short description of the deprecation. MUST be a non-empty string. |
+| alternatives | array of [CatalogObjectId](#catalog-object-id-object) objects | An array of services or plans that the Service Broker considers to be good alternatives. If present, MUST contain at least one alternative. |
```
pmorie: There might be something to think about here re: deprecation or EOL.

spec.md
```diff
@@ -586,18 +589,64 @@ schema being used.
...
+
+\* Fields with an asterisk are REQUIRED.
+
+##### Catalog Object Id Object 
+
+| Response Field | Type | Description |
+| --- | --- | --- |
+| description | string | A short description to provide contextual information about this catalog object reference. |
+| service_id | string | If present, it MUST be the ID of a service. If omitted, `plan_id` MUST be present. |
+| plan_id | string | If present, it MUST be the ID of a plan. If omitted, `service_id` MUST be present. |
```
pmorie: Do we need to qualify that a deprecation for a service must not contain a planID, and vice-versa?

spec.md
```diff
@@ -586,18 +589,64 @@ schema being used.
...
 ### Adding a Service Broker to the Platform
 
 After implementing the first endpoint `GET /v2/catalog` documented
 [above](#catalog-management), the Service Broker will need to be registered
 with your Platform to make your services and plans available to end users.
 
+### Deprecating services and plans
+
+A Service Broker can signal to a Platform (through the Catalog) that a service
+or plan is deprecated. The Service Broker MAY provide messaging and
```
pmorie: "messaging" seems a little ambiguous here - is that desired?

spec.md
```diff
@@ -586,18 +589,64 @@ schema being used.
...
+alternatives for the deprecated service or plan. This gives the Platform the
+opportunity to display meaningful advice to their users to migrate away from
+the deprecated service or plan.
+
+Service Brokers MUST allow existing Service Instances to be updated and
+bindable as they were before depreciation. Deprecation MUST NOT effect the
+usability of existing Service Instances and Bindings.
+
+The Service Broker MUST allow the Platform to continue to provision new Service
```
pmorie: I'm not sure I'm on board with MUST here. I don't think a platform should be obligated to allow users to provision deprecated services, and I'm not sure broker authors will be crazy about this either.

Do we need to add an indicator the broker author can use to say whether new instances are allowed to be provisioned?

mattmcneeney:
We spoke about this on the weekly call today. The consensus was that this is potentially useful for service providers, but it is not at the top of anybody's priority list and nobody volunteered to own it. I'll leave the `needs-discussion` label on this so that we can pick it up as and when this becomes important for folks.

spec.md
```diff
@@ -199,6 +206,7 @@ A web-friendly display name is camel-cased with spaces and punctuation supported
 | [dashboard_client](#dashboard-client-object) | object | Contains the data necessary to activate the Dashboard SSO feature for this service. |
 | plan_updateable | boolean | Whether the service supports upgrade/downgrade for some plans. Please note that the misspelling of the attribute `plan_updatable` to `plan_updateable` was done by mistake. We have opted to keep that misspelling instead of fixing it and thus breaking backward compatibility. Defaults to false. |
 | [plans*](#plan-object) | array-of-objects | A list of plans for this service, schema is defined below. MUST contain at least one plan. |
+| deprecated | boolean | Indicates that this service and all of its plans are deprecated and will be removed in the future. See [deprecating services and plans](#deprecating-services-and-plans) for more information. |
```
GEricson: s/will/might/
The future may find a reason to revert from deprecated to not-deprecated.

n3wscott:
I am closing this, looks like the group has moved on. Can fork and remake if folks want to bring it back.

Issue #498, opened 2018-04-17
Return dashboard_url in polling last operation of service instance


leonwanghui:
## Background
Currently for provision request, the service broker will return `dashboard_url` field in the response body.
## Problem
For aync operation, some brokers can't return the `dashboard_url` value in time, but in PollLastOperation request there is no field that indicates the required `dashboard_url` value, so this may cause some issues that the platform could not get the dashboard_url info and pass it to users.
## Goal
Add `dashboard_url` field definition in response body of poll last operation, so that the platform can get the info of dashboard_url when brokers finish provisioning operation.

jmrodri:
This is EXACTLY the issue I wanted to talk about. I can take ownership of this one.

duglin:
Assigned @leon but Jesus will help

leonwanghui:
Thanks @jmrodri and @duglin

mattmcneeney:
The `dashboard_url` can now be fetched via the [Fetching a Service Instance](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#fetching-a-service-instance) endpoint, so closing this!

gberche-orange:
@mattmcneeney The [GET /v2/service_instances/:instance_id dashboard_url response field](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#body-5) mentions
> Note: a Service Broker that wishes to return dashboard_url for a Service Instance MUST return it with the initial response to the provision request, even if the service is provisioned asynchronously.

Should we read it is valid for a broker to return a different value of the `dashboard_url` between the  initial response to the provision request, and the `GET /v2/service_instances/:instance_id` response ?

PR #507, opened 2018-04-19
Add resource message in the response of polling the last operation

leonwanghui:
Solve #498 

Signed-off-by: leonwanghui <wanghui71leon@gmail.com>

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -654,7 +654,10 @@ An asynchronous response triggers the Platform to poll the endpoint
 indicates that the requested operation has succeeded or failed. Service Brokers
 MAY include a status message with each response for the `last_operation`
 endpoint that provides visibility to end users as to the progress of the
-operation.
+operation. For Service Brokers which can't return the resource message in
+asynchronous response, they SHOULD guarantee resource message included in the
+response of the `last_operation` endpoint when the asynchronous operation
+completed.
 
 ## Blocking Operations
 
@@ -749,6 +752,7 @@ For success responses, the following fields are defined:
 | --- | --- | --- |
 | state* | string | Valid values are `in progress`, `succeeded`, and `failed`. While `"state": "in progress"`, the Platform SHOULD continue polling. A response with `"state": "succeeded"` or `"state": "failed"` MUST cause the Platform to cease polling. |
 | description | string | A user-facing message displayed to the Platform API client. Can be used to tell the user details about the status of the operation. If present, MUST be a non-empty string. |
+| resource | object | An opaque object of resource message for fetching the result of the operation. If present, MUST be included when `state` is `succeeded`. |
 
 \* Fields with an asterisk are REQUIRED.
 
@@ -756,6 +760,16 @@ For success responses, the following fields are defined:
 {
   "state": "in progress",
   "description": "Creating service (10% complete)."
+}
+```
+or
+```
+{
+  "state": "succeeded",
+  "description": "Service created.",
+  "resource": {
+    "dashboard_url": "http://example-dashboard.example.com/9189kdfsk0vfnku"
+  }
 }
 ```
 


n3wscott:
As we talked about in the weekly call, this is a pretty strange API choice. RESTful design should have a single way to get resource information. There is now a GET allowed for bindings and instances, GET should be used. 

I am not apposed to a flag for signaling that the result of the async operation has modified the resource in question and a GET is now required. But passing the resource in get last operation is not a good design in my opinion.

leonwanghui:
@duglin @jmrodri

spec.md
```diff
@@ -654,7 +654,10 @@ An asynchronous response triggers the Platform to poll the endpoint
 indicates that the requested operation has succeeded or failed. Service Brokers
 MAY include a status message with each response for the `last_operation`
 endpoint that provides visibility to end users as to the progress of the
-operation.
+operation. For Service Brokers which can't return the resource message in
+asynchronous response, they SHOULD guarantee resource message included in the
+response of the `last_operation` endpoint when the asynchronous operation
+completed.
```
duglin: I think we'd need to make it clear that even if the dashboard_url isn't ready yet, they still need to send something because the spec requires it - even if the URL is temporary.

duglin:
I'm pretty sure I'm in the minority on this but I'll say it anyway :-)
I actually don't think returning the resource (or whatever you would have seen in the sync response case) in the last_operation() response is that bad. I say this for a couple of reasons:
1 - To me, there should be no difference between sync and async message flows w.r.t. the net results/outputs. At the end of the async ops I should have the same data I would get from doing the sync op.  Forcing me to do an extra GET at the end goes against that thought.
2 - GET is optional and requiring it for async update() feels like a large burden if you wouldn't (couldn't) have supported GET in the first place.
3 - forcing someone to do an extra message exchange just to get something that we know they're going to immediately turn around and do a GET for is kind of silly and sub-optimal.

spec.md
leonwanghui: So basically you mean the platform would check the dashboard_url even if the state is` in progress`? Or service broker can return an empty one if they are not ready?

leonwanghui: @duglin Got it, what I meant is service brokers can pass anything they want before the operation is completed, but they have to make the right one included if the `dashborad_url` exists. Does that ok to you? I will update the spec, thanks!

mattmcneeney:
@duglin I partly agree, but it's a nuisance for Platforms to have to remember three places to check for `dashboard_url` (sync provision, last operation and GET). I think we should talk about this one on the weekly call.

jmrodri:
typo: in PR title. reponse should be response

spec.md
```diff
@@ -749,6 +752,7 @@ For success responses, the following fields are defined:
 | --- | --- | --- |
 | state* | string | Valid values are `in progress`, `succeeded`, and `failed`. While `"state": "in progress"`, the Platform SHOULD continue polling. A response with `"state": "succeeded"` or `"state": "failed"` MUST cause the Platform to cease polling. |
 | description | string | A user-facing message displayed to the Platform API client. Can be used to tell the user details about the status of the operation. If present, MUST be a non-empty string. |
+| resource | object | An opaque object of resource message for fetching the result of the operation. If present, MUST be included when `state` is `succeeded`. |
```
jmrodri: What do we mean by "If present, MUST be included..." If present where?

jmrodri: Also, does the resource object look like anything? or is it just any object? How does the platform know what to do with it?

spec.md
```diff
@@ -758,6 +762,16 @@ For success responses, the following fields are defined:
...
 }
 ```
+or
+```
+{
+  "state": "succeeded",
+  "description": "Service created.",
+  "resource": {
+    "dashboard_url": "http://example-dashboard.example.com/9189kdfsk0vfnku"
+  }
```
jmrodri: So will resource always be just dashboard_url? or will it be any opaque object? or will it be a set of define attributes that we will add to as we need new fields?

leonwanghui:
@jmrodri  Thanks, it's changed now

mattmcneeney:
This was discussed in detail during this week's call. There seemed to be differing opinions on whether or not this was a sensible change.

I'm interested to dig into whether or not this actually simplifies things for the broker:
* With this change, the broker needs to remember the `operation` ID that refers to the provision/update, and remember the `dashboard_url` that corresponds to the service instance
* Without this change, the broker needs to remember the `dashboard_url` that corresponds to the service instance and implement the `GET /v2/service_instances/:guid` endpoint

The latter seems simpler to me for service brokers to implement. The GET service instance endpoint doesn't require any other information to be returned.

duglin:
Brokers have to remember the async operation anyway - if nothing else so they can return its status (at least once) back to the platform. Keeping track of one extra value in that case should be trivial, and probably a lot less code than adding an entire GET endpoint and keeping those values for the lifetime of the instance.

mattmcneeney:
That's the bit I'm questioning @duglin - is it actually less code?

> and keeping those values for the lifetime of the instance.

This is the scary bit too. There is no guidance on how long brokers should hold onto the state of operations, so there are no guarantees that a platform would be able to get the data related to the instance when it wants to.

mattmcneeney:
Got it. Well even if it isn't more code, I feel fairly opposed to having the `last_operation` endpoint return arbitrary data, especially given its ephemeralness. Fetching this information using the GET endpoint (as long as I haven't greatly underestimated the complexity for @leonwanghui's broker to support this) is my preference.

leonwanghui:
> Without this change, the broker needs to remember the dashboard_url that corresponds to the service instance and implement the GET /v2/service_instances/:guid endpoint

@mattmcneeney Hi, IMO without this change, service broker should also need to be perceptive of when `dashboard_url` is created and inform platform to fetch a real dashboard_url value. If so, he would remeber the `operation` ID that refers to the provision/update anyway.

> I feel fairly opposed to having the last_operation endpoint return arbitrary data, especially given its ephemeralness.

I am sorry for not making it clear before, I thought that we would add more parameters (credentials, route_services, volume_mounts and so on) after we support async binding, so I designed the `resource` field as extensible for avoiding break change later on. And I'm ok to put some restriction on this by designing a set of define attributes.

As @duglin explained, we haven't supported GET endpoint yet. From my perspective, there is quite a few work to implement the feature in GET endpoint way.

leonwanghui:
As discussed on the meeting last week, this PR is closed for not handling the issue #498 well by moving all response to pool_last_operation. There is a more elegant way that enables service broker trigger the GET method when required. Since I am not familiar the latter way, I would appreciate it if someone could present it by pulling another PR.

PR #517, opened 2018-05-15
 GET MAY return creds unable to be used

### LINKED ISSUE #516: GET should be allowed to return redacted credentials.
Discussion internal to IBM views the GET endpoint for bindings containing credentials as encouraging improper storage of passwords/secrets.

In general, we're not supposed to be storing passwords in plaintext or any retrievable form (hashed & salted).

Should we add text indicating that it is up to the broker to determine what the content of the `credentials` field is? The field may not contain enough information to connect to the service being provided if credentials are stored in a manner that makes them unavailable.

MHBauer:
resolves #516

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1195,6 +1195,12 @@ utilize the Service Instance. Credentials SHOULD be unique whenever possible, so
 access can be revoked for each binding without affecting consumers of other
 bindings for the Service Instance.
 
+This field MAY be omitted if the Service Broker either cannot return the credentials
+or does not want to for security reasons. During an asynchronous bind 
+operation the Service Broker MUST include this field in at least the first response
+after the Service Binding has been successfully created, otherwise the Platform will
+not have the credentials to pass along. 
+
 #### Log Drain
 
 There are a class of Service Offerings that provide aggregation, indexing, and
@@ -1360,7 +1366,7 @@ For `200 OK` and `201 Created` response codes, the following fields are defined:
 
 | Response Field | Type | Description |
 | --- | --- | --- |
-| credentials | object | A free-form hash of credentials that can be used by applications or users to access the Service Instance. This field MAY be omitted if the Service Broker either cannot return the credentials or does not want to for security reasons. Note that during an asynchronous bind operation the Service Broker MUST include this field in at least the first response after the Service Binding has been successfully created, otherwise the Platform will not have the credentials to pass along to the application/user. |
+| credentials | object | A free-form hash of credentials that can be used by applications or users to access the service. MUST be returned if the Service Broker supports generation of credentials. |
 | syslog_drain_url | string | A URL to which logs MUST be streamed. `"requires":["syslog_drain"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
 | route_service_url | string | A URL to which the Platform MUST proxy requests for the address sent with `bind_resource.route` in the request body. `"requires":["route_forwarding"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
 | volume_mounts | array of [VolumeMount](#volume-mount-object) objects | An array of configuration for remote storage devices to be mounted into an application container filesystem. `"requires":["volume_mount"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -1374,7 +1374,7 @@ For success responses, the following fields are defined:
 
 | Response Field | Type | Description |
 | --- | --- | --- |
-| credentials | object | A free-form hash of credentials that can be used by applications or users to access the service. |
+| credentials | object | A free-form hash of credentials that can be used by applications or users to access the service. If credentials have been stored in a format that does not allow retrieval of a usable result (e.g. a salted and hashed password), the broker MAY omit the field from `credentials`. This can result in a return value that is not usable to access the service. |
```
duglin: s/from `credentials`//
I assume they should just skip the entire "credentials" field, and not just certain items in it. Returning a partial list of creds could be very misleading so I think it should be all or nothing.

duglin:
@MHBauer did you want to address @n3wscott's comment?

spec.md
```diff
@@ -1374,7 +1374,7 @@ For success responses, the following fields are defined:
 
 | Response Field | Type | Description |
 | --- | --- | --- |
-| credentials | object | A free-form hash of credentials that can be used by applications or users to access the service. |
+| credentials | object | A free-form hash of credentials that can be used by applications or users to access the service. If credentials have been stored in a format that does not allow retrieval of a usable result (e.g. a salted and hashed password), the broker MAY omit this field. This can result in a return value that is not usable to access the service. |
```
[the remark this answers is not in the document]
duglin: I would suggest we just remove that sentence.

re: async-binding... perhaps we need to add something else here to make it clear that each GET can decide when to return the credentials.  Meaning, the first GET might do so, but then subsequent GETs may not because they're stored in a hash. This allows for a one-time-only retrieval so async bindings work, but after that things remain secure.  But, in the case of async it MUST support at least one GET of the creds.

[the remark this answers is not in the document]
MHBauer: Is removing that sentence acceptable?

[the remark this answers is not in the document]
MHBauer: `If credentials have been stored in a format that does not allow retrieval of a usable result the broker MAY omit this field.`
@n3wscott @mattmcneeney Is only that sentence being added acceptable? I feel like it makes things murkier rather than clearer without the example.

MHBauer:
@mattmcneeney During the call today, you took specific issue with the given example. I do not recall the exact context, but I feel like it adds one specific intent without restricting it to solely that case. I'll work with @duglin on alternative phrases and positions.

mattmcneeney:
Sounds good, thanks @MHBauer. My only concern with the current text is that it suggests the fact a credential is hashed/salted is a reason not to return it, but from a broker author's perspective, their concern really is security.

Maybe something like this:
```
A free-form hash of credentials that can be used by applications or users to access the Service Instance. This field MAY be omitted if the Service Broker either cannot return the credentials or does not want to for security reasons.
```

duglin:
How about a minor addition:
```
A free-form hash of credentials that can be used by applications or users to access
the Service Instance. This field MAY be omitted if the Service Broker either cannot
return the credentials or does not want to for security reasons. Note that during
an asynchronous bind operation the Service Broker MUST include this field in at least
the first response after the Service Binding has been successfully created, otherwise
the Platform will not have the credentials to pass along to the application/user.
```

MHBauer:
Does it all just go into one giant line? Seems to render fine, but looks terrible in plaintext.

spec.md
```diff
@@ -1374,7 +1374,7 @@ For success responses, the following fields are defined:
 
 | Response Field | Type | Description |
 | --- | --- | --- |
-| credentials | object | A free-form hash of credentials that can be used by applications or users to access the service. |
+| credentials | object | A free-form hash of credentials that can be used by applications or users to access the Service Instance. This field MAY be omitted if the Service Broker either cannot return the credentials or does not want to for security reasons. Note that during an asynchronous bind operation the Service Broker MUST include this field in at least the first response after the Service Binding has been successfully created, otherwise the Platform will not have the credentials to pass along to the application/user. |
```
MHBauer: @n3wscott does this address the async bind case?

mattmcneeney: It's a little weird to have async bind information in this PR since it may be submitted before #334

MHBauer: @mattmcneeney does that mean you want it back out and a reminder in #334 ?

duglin:
Please review but adding DO NOT MERGE since we need the async PR to go in first.

spec.md
fmui: This addition introduces an inconstancy. If we allow the omission of the credentials here, we also have to allow the omission for the same reasons when a binding is (re-)created with the identical parameters (status code 200 OK).

duglin: @MHBauer please make similar edits to the create per Florian's suggestion - then we'll discuss at the f2f

MHBauer: I disagree. The door shuts eventually, but that doesn't mean it's shut from the start.

For synchronous binding the credentials can be returned the once before becoming irretrievable.

For async binding we have to preserve the response for only so long.

MHBauer: I feel like this is very much in conflict with stateless-ness. It seems like it would now be saying "must be preserved forever". Being able to do this now implies that the credential generation is deterministic. I'm not sure this is clarifying anything anymore.

MHBauer: What's the equivalent of ` MUST include this field in at least the first response`? We now appear to be saying, "MUST include forever".

spec.md
```diff
@@ -1280,7 +1280,7 @@ For success responses, the following fields are defined:
...
```
duglin: I think the "This field MAY be omitted..." sentence needs to be prefixed with "If a 200 is returned..." or something like that.

spec.md
```diff
@@ -1195,6 +1195,12 @@ utilize the Service Instance. Credentials SHOULD be unique whenever possible, so
 access can be revoked for each binding without affecting consumers of other
 bindings for the Service Instance.
 
+This field MAY be omitted if the Service Broker either cannot return the credentials
+or does not want to for security reasons. During an asynchronous bind 
+operation the Service Broker MUST include this field in at least the first response
```
mattmcneeney: first GET response (with link)

spec.md
```diff
@@ -1195,6 +1195,12 @@ utilize the Service Instance. Credentials SHOULD be unique whenever possible, so
...
+after the Service Binding has been successfully created, otherwise the Platform will
+not have the credentials to pass along. 
+
```
MHBauer: Dug suggested 
```
Note: some brokers (due to security concerns... blah blah) MAY choose to only return the credentials for a binding once even though there could be replayed binding requests or GETs of the binding resource.
```

spec.md
MHBauer: link to what?

MHBauer:
What's next to move this forward? Going back to some comment in each table? Or keep the guidance on the description of a credentials binding?

fmui:
LGTM

duglin:
LGTM

zrob:
lgtm

tinygrasshopper:
In my mind this conflicts with the MUST on the binding response for credentials  

```
| credentials | object | A free-form hash of credentials that can be used by applications or users to access the service. MUST be returned if the Service Broker supports generation of credentials and the Service Binding was provisioned asynchronously. |
```

Should we change the response description for fetch binding to reflect this optionality?

Issue #509, opened 2018-04-23
Short/Long Description Fields should also specify valid Character lengths


codevulture:
Description fields with names short and long makes sense but there should also be specified guidelines for how much character lengths are permitted in short/long for better understanding/Usage.

At various places descriptions are present like: https://github.com/openservicebrokerapi/servicebroker/blob/v2.13/spec.md#request

jmrodri:
The only thing I don't like about specifying character lengths is that each platform / broker can define what is short and long for them. Was there a specific problem that you encountered that required knowing the specific lengths?

mattmcneeney:
Platforms and tooling will likely have very different lower and upper limits for strings (depending on the database they use for example). We could try to specify a minimum length that must be supported, but in reality, it's very unlikely that an engineer working on a UI for a platform would ever come to this repository looking for what that limit is. So I'm not sure we would gain much from this change. I'd be up for adding some guidance to the spec though, such as `Note that different Platforms and other tooling may have different limitations on how the information contained in the catalog can be displayed, so care should be taken to ensure all strings are of a sensible length.`

codevulture:
@mattmcneeney : thanks. But again `sensible length` is something which could mean anything :) , I think we can close the bug if this is upto the platform but i was hoping we specify something like short desc should not be more than say 1000 chars . Please let me know your opinion.

mattmcneeney:
It fundamentally is up to the Platform, but I agree that at least adding a recommended max to the spec will hopefully encourage Platform authors to at least support that size. Thoughts @duglin ?

bingosummer:
We find, if the plan description is more than 255 characters, we can't create the service broker. CF CLI will throw an error.
```
FAILED
Server error, status code: 500, error code: 10011, message: Database error
```
From cloud_controller_ng.log in the `api` VM, we can get the error:
```
[...]

PR #518, opened 2018-05-15
Add guidance for supported string lengths

mattmcneeney:
See #509

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -321,6 +321,10 @@ Instances of these plans. For example, Platforms might restrict the actions
 that users can perform on existing Service Instances if the associated service
 or plan is deleted. Consider your deprecation strategy.
 
+Platforms MAY have limits on the length of strings that they can handle or
+display to end users, such as the description of a Service or Service Plan. It
+is RECOMMENDED that strings do not exceed 255 characters to increase the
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -321,6 +321,10 @@ Instances of these plans. For example, Platforms might restrict the actions
 that users can perform on existing Service Instances if the associated service
 or plan is deleted. Consider your deprecation strategy.
 
+Platforms MAY have limits on the length of strings that they can handle or
+display to end users, such as the description of a Service or Service Plan. It
+is RECOMMENDED that strings do not exceed 255 characters to increase the
+likelihood of having compatibility with any Platform.
```
n3wscott: Sorry to sandbag one more comment, but perhaps we want to give an out to the platform for truncation? like:

`Platforms MAY truncate strings longer than 255 characters.` or something like that. This says the platforms have limits as a warning to the broker but does not say what the platform will do to a string from a broker that is too long.

mattmcneeney:
@n3wscott Ha I didn't even realise I was doing that... Serves me right for editing in the browser!

zrob:
lgtm

fmui:
LGTM

PR #521, opened 2018-05-23
Remove version-specific text from the specification

mattmcneeney:
The specification is versioned, so I don't believe we need this bit of text referring to a specific version of the spec supporting a new feature.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -951,10 +951,10 @@ support per service by including `"plan_updateable": true` in its [catalog
 endpoint](#catalog-management).
 
 If `"plan_updateable": true` is declared for a plan in the
-[Catalog](#catalog-management) endpoint, the Platform MAY request a plan change 
-on a Service Instance using the given plan. Otherwise, Platforms MUST NOT make 
-any plan change requests to the Service Broker for any Service Instance using the 
-given plan, but MAY request an update to the Service Instance parameters.
+[Catalog](#catalog-management) endpoint, the Platform MAY request a plan change
+on a Service Instance using the given plan. Otherwise, Platforms MUST NOT make
+any plan change requests to the Service Broker for any Service Instance using
+the given plan, but MAY request an update to the Service Instance parameters.
 
 Not all permutations of plan changes are expected to be supported. For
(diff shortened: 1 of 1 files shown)


mattmcneeney:
cc @duglin @n3wscott following yesterday's call

duglin:
@mattmcneeney I think finding a way to drop the version string is ok, but this proposal drops the requirement that the Platform MUST detect the new value and use it going forward. Or are you thinking that its not required because its obvious?

mattmcneeney:
@duglin exactly. For provision, we don't say that Platforms MUST use the `dashboard_url` that is returned, so I wasn't sure why we would say so here!

spec.md
```diff
@@ -1086,19 +1086,9 @@ For success responses, the following fields are defined:
 
 | Response Field | Type | Description |
 | --- | --- | --- |
-| dashboard_url | string | The updated URL of a web-based management user interface for the Service Instance; we refer to this as a service dashboard. The URL MUST contain enough information for the dashboard to identify the resource being accessed (`0129d920a083838` in the example below). Note: a Service Broker that wishes to return `dashboard_url` for a Service Instance MUST return it with the initial response to the update request, even if the service is updated asynchronously. If present, MUST be a non-empty string. |
+| dashboard_url | string | The updated URL of a web-based management user interface for the Service Instance; we refer to this as a service dashboard. The URL MUST contain enough information for the dashboard to identify the resource being accessed (`0129d920a083838` in the example below). Note: a Service Broker that wishes to return `dashboard_url` for a Service Instance it MUST return it with the initial response to the update request, even if the service is being updated asynchronously. If present, MUST be a non-empty string. |
```
duglin: I think the added "it" before "MUST return" isn't correct. Unless you also want to add something like "When" to the start of the sentence.

duglin:
@mattmcneeney ok I can buy into that

spec.md
mattmcneeney: Oops - not sure what happened in my brain there! fixing

fmui:
LGTM

duglin:
Just one more review needed

zrob:
lgtm

PR #522, opened 2018-05-23
Clarify plan id for polling last operation after an update

### LINKED ISSUE #396: Clarify which plan ID to include in a last operation poll for an async plan update
For an asynchronous plan update on an instance, which plan ID should be used for last operation requests? Should it be the ID of the plan to which the instance originally belonged? Or should it be the ID of the plan to which the broker is in the process of moving the instance?

mattmcneeney:
Resolves #396


duglin:
LGTM

I don't think we have a choice since the op may fail

mattmcneeney:
@n3wscott to change svc-cat to use the old `plan_id` before we merge this (hopefully this week)

duglin:
LGTM

zrob:
lgtm

n3wscott:
Service Catalog still does the old way. This will have to be updated.

LGTM

PR #523, opened 2018-05-24
Fix inconsistent asynchronous operations text

### LINKED ISSUE #520: Inconsistency in asynchronous text surrounding 422
The spec contains both a MUST and a SHOULD on whether brokers return 422 & `AsyncRequired` if they do not support a synchronous request. 

`
If this parameter is not included in the request, and the Service Broker can only provision a Service Instance of the requested plan asynchronously, the Service Broker MUST reject the request with a 422 Unprocessable Entity.
`
[link](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#parameters-1)

`
If the parameter is not included or is set to false, and the Service Broker cannot fulfil the request synchronously (guaranteeing that the operation is complete on response), then the Service Broker SHOULD reject the request with the status code 422 Unprocessable Entity and a response body containing error code "AsyncRequired"(see Service Broker Errors).  
`
[link](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#asynchronous-operations)

`
422 Unprocessable Entity | MUST be returned if the Service Broker only supports asynchronous provisioning for the requested plan and the request did not include ?accepts_incomplete=true. The response body MUST contain a response body containing error code "AsyncRequired" (see Service Broker Errors).
`
[link](https://github.com/duglin/servicebroker/blob/d7e3267c429c9aa97c836ac436067782aaea5f0e/spec.md#response-2)

There a bunch of places where this is inconsistent. I'm guessing this was just a mistake and that a broker MUST return 422 and MUST return `AsyncRequired` if it does not support sync.

Happy to create a PR.

Samze:
closes #520

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -641,7 +641,7 @@ For a Service Broker to return an asynchronous response, the query parameter
 `accepts_incomplete=true` MUST be included the request. If the parameter is not
 included or is set to `false`, and the Service Broker cannot fulfil the request
 synchronously (guaranteeing that the operation is complete on response), then
-the Service Broker SHOULD reject the request with the status code `422
+the Service Broker MUST reject the request with the status code `422
 Unprocessable Entity` and a response body containing error code
 `"AsyncRequired"` (see [Service Broker Errors](#service-broker-errors)). The
 error response MAY include a helpful error message in the `description` field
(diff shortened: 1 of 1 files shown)


Samze:
@duglin in #520 you suggested removing the duplication, but this largely exists because we repeat the text for each async operations (Provision/Update/Delete). I think keeping the duplication makes the spec easier to read but harder to maintain and we should make it as easy as possible to read.

duglin:
LGTM

avade:
LGTM

PR #526, opened 2018-06-06
Add support for version header in responses

duglin:
This allows for brokers to tell platforms what version of the spec
MUST be used when processing the response message.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -137,21 +137,36 @@ For changes in older versions, see the [release notes](https://github.com/opense
 
 ## API Version Header
 
-Requests from the Platform to the Service Broker MUST contain a header that
-declares the version number of the Open Service Broker API that the Platform
-is using:
+This specification defines an HTTP header that MUST be included in all
+messages (requests and responses) by endpoints adhering to this specification:
 
 `X-Broker-API-Version: 2.13`
 
+This header declares the version number of the Open Service Broker API
+specification that the sender is using.
+
 The version numbers are in the format `MAJOR.MINOR` using semantic versioning.
 
-This header allows Service Brokers to reject requests from Platforms for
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -137,22 +137,37 @@ For changes in older versions, see the [release notes](https://github.com/opense
 
 ## API Version Header
 
-Requests from the Platform to the Service Broker MUST contain a header that
-declares the version number of the Open Service Broker API that the Platform
-is using:
+This specification defines an HTTP header that MUST be included in all
```
pmorie: As written this seems like a backward-incompatible change, since brokers will have to include the header on all responses.

Is that intentional?

spec.md
```diff
@@ -137,22 +137,37 @@ For changes in older versions, see the [release notes](https://github.com/opense
...
 used instead.
 
+When included in a response message, it allows the receiver to understand
+which version of the Open Service Broker API specification it MUST adhere to
+when processing the message. Due to the message being a response, and there
+not being a way to reject it, receivers MUST support all previous minor
+versions of the specification corresponding to the version number used in the
+request message. Likewise, response message's version number MUST:
+- use the same major number as the corresponding request
+- use a minor version number equal to, or less than, the correspdoning request
```
fmui: What should a platform do if the broker provides a higher version number? Should it ignore it or try to roll back the last operation?

duglin: At that point the broker is non-compliant (per this bullet) but if it does happen then I think the platform would ignore it since the operation is already done. I could add text to say that if people want, but I'm not sure if saying what should be done when one side violates the spec is a form of encouraging it  :-)

duglin:
Closed per yesterday's f2f talks. We don't think we need this right now but we can reopen it if we find a need for it.

PR #527, opened 2018-06-07
Clarify platform behavior when dashboard url is not returned 

tinygrasshopper:
while polling on update.

Have not expressed an opinion on how the broker can clear out an existing dashboard url, but clarified current behaviour in a way that its backwards compatible.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1169,7 +1169,7 @@ For success responses, the following fields are defined:
 
 | Response Field | Type | Description |
 | --- | --- | --- |
-| dashboard_url | string | The updated URL of a web-based management user interface for the Service Instance; we refer to this as a service dashboard. The URL MUST contain enough information for the dashboard to identify the resource being accessed (`0129d920a083838` in the example below). Note: a Service Broker that wishes to return `dashboard_url` for a Service Instance MUST return it with the initial response to the update request, even if the Service Instance is being updated asynchronously. If present, MUST be a non-empty string. |
+| dashboard_url | string | The updated URL of a web-based management user interface for the Service Instance; we refer to this as a service dashboard. The URL MUST contain enough information for the dashboard to identify the resource being accessed (`0129d920a083838` in the example below). Note: a Service Broker that wishes to return `dashboard_url` for a Service Instance MUST return it with the initial response to the update request, even if the Service Instance is being updated asynchronously. If present, MUST be a non-empty string. If not present, the Platform MUST retain the previous value of the `dashboard_url`. |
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -1086,7 +1086,7 @@ For success responses, the following fields are defined:
...
 | --- | --- | --- |
-| dashboard_url | string | The updated URL of a web-based management user interface for the Service Instance; we refer to this as a service dashboard. The URL MUST contain enough information for the dashboard to identify the resource being accessed (`0129d920a083838` in the example below). Note: a Service Broker that wishes to return `dashboard_url` for a Service Instance MUST return it with the initial response to the update request, even if the service is updated asynchronously. If present, MUST be a non-empty string. |
+| dashboard_url | string | The updated URL of a web-based management user interface for the Service Instance; we refer to this as a service dashboard. The URL MUST contain enough information for the dashboard to identify the resource being accessed (`0129d920a083838` in the example below). Note: a Service Broker that wishes to return `dashboard_url` for a Service Instance MUST return it with the initial response to the update request, even if the service is updated asynchronously. If present, MUST be a non-empty string. If not present, the Platform MUST retain the previous value of the `dashboard_url`. |
```
mattmcneeney: `s/not/NOT` else travis will complain

pmorie: We might as well call out that updating to an unset value is not currently allowed by the spec.

pmorie:
One minor nit - there's a misspelling in the commit message ('clarify')

spec.md
duglin: I think @tinygrasshopper is working on a way to deal with that.

duglin: @mattmcneeney which `not` are you referring to?

mattmcneeney: The not in `If not present`. Would travis not shout about that?

tinygrasshopper: We thought about the unset dashboard url case, all the solutions we came up for the usecase had a high cost of change in the spec. I am tending toward not solving it right now, as we dont have any known usecases of the removal flow right. With this change I would like to clarify the existing behaviour. 

@pmorie I am in two minds about calling out unsupported flows in the spec, as there is no precedence for that in the rest of the spec.

duglin:
rebase needed

duglin:
SGTM just waiting for the rebase

mattmcneeney:
@tinygrasshopper We discussed this on today's call and are happy to start reviewing, but it needs a rebase first!

tinygrasshopper:
@mattmcneeney rebased

duglin:
LGTM

PR #535, opened 2018-07-03
Make identity header naming consistent

williammartin:
Another small fixup around the `Originating-Identity` header. Might seem trivial but as I got to this section I wasn't initially sure which was the "correct" version, until I got to later sections where the hyphen is repeated. Might just help to clarify this by being consistent.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -224,12 +224,12 @@ will need to provide this identification information to the Service Broker on
 each request. Platforms MAY support this feature, and if they do, they MUST
 adhere to the following:
 - For any OSBAPI request that is the result of an action taken by a Platform's
-  user, there MUST be an associated `OriginatingIdentity` header on that HTTP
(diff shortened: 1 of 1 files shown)


duglin:
should we be really explicit and make it `X-Broker-API-Originating-Identity` ?

williammartin:
Yeh, I was wondering whether we should do that _or_ have a section on headers in the specification that they _all_ start with `X-Broker-API`, then we could miss out the prefix in other places. Happy to fix the PR with either.

duglin:
For now, let's just take the easy path and use the actual header name

williammartin:
Cool, fixed up.

duglin:
Speedy!  Thanks!
LGTM

pmorie:
LGTM

duglin:
one more review needed

PR #537, opened 2018-07-03
Note that change policy only applies to minor bump

### LINKED ISSUE #536: What is the change policy across major versions?
While reading the [change policy](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#change-policy), it seems to be written as if either each version of the OSBAPI would be a different spec, or that there can be no breaking changes across major version changes? 

For example, `Existing endpoints and fields MUST NOT be removed or renamed.` makes sense for compatibility across minor versions, but I suspect this is not a guarantee the OSBAPI wants to make across major versions?

Do we need to add a small note to this effect?

williammartin:
[fixes #536]

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -4,7 +4,7 @@
   - [API Overview](#api-overview)
   - [Notations and Terminology](#notations-and-terminology)
   - [Changes](#changes)
-    - [Change Policy](#change-policy)
+    - [Change Policy](#change-policy-for-minor-versions)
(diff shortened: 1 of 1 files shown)


duglin:
LGTM

fmui:
LGTM

mattmcneeney:
@williammartin the anchor in the ToC needs updating as Travis is failing. We will then need to re-review this.

duglin:
LGTM

Issue #528, opened 2018-06-13
Incompatible change in v2.12: the broker MUST return 200 OK if the binding already exists 


fmui:
In the v2.11 spec, the broker had a choice to return 200 OK when the platform sends identical parameters for an exiting binding.
The expected broker behavior changed in v2.12 to a MUST.

v2.11:

> 200 OK  May be returned if the binding already exists and the requested parameters are identical to the existing binding.

v2.12:

> 200 OK  MUST be returned if the binding already exists and the requested parameters are identical to the existing binding.

Although v2.11 was ambiguous, we are now in a situation where v2.11 brokers may not work with platform supporting v2.12 or v2.13.

duglin:
Closing since #540 is closed/merged.

PR #540, opened 2018-07-04
Fix incompatible change in v2.12

fmui:
This pull requests addresses issue #528.
It should not be merged before the PRs #334 and #517 are discussed and merged.

Please note that is a breaking change for platforms. If a platform sends identical parameters to an existing binding, the broker may respond with 200 OK with credentials, 200 OK without credentials, 201 created with credentials, or 409 Conflict.


duglin:
LGTM

PR #541, opened 2018-07-09
Be clear about the state of resources after a failure

duglin:
I'd like to keep the first MUST but if people are afraid that we can't
due to it technically being a breaking spec change, then we can drop it
down to a SHOULD.  But since 4xx's are means to be error for "bad input"
I figured brokers should catch these before any real processing on the
resource happened.


mattmcneeney:
@duglin Does this text make sense given that we use `410 Gone` in the spec for delete operations with the description `The Platform MUST consider this response a success and forget about the resource.`?

duglin:
I think a 410 is still ok with this text because its still "Gone" after the broker responded with a "410 Gone". Its still semantically in the same state.  Right?

duglin:
Updates the text based on yesterday's meeting

Issue #514, opened 2018-05-02
Reverse lookup of service instance id during provisioning


jakkahub:
We are trying to get service name of an instance id during provisioning. We know, instance id is generated by the platform when creating a service instance. This is passed to the service broker but the name isn’t. We can call the CC API `GET /v2/service_instances/:guid` to get the service name of an instance id.

But this didn't worked for us, since we are trying to get service name before the service broker returns with 202 to CF as it need to know that the broker has accepted the request.

Here is our flow and we are trying to make step 3 work in between steps 2 and 4.
1) Submit service provisioning command cf create-service
2) CF calls service brokers `PUT /v2/service_instances/:instance_id` endpoint
3) Service broker’s service instance creation implementation gets service name for instance guid from CF using `GET /v2/service_instances/:guid`
4) Service broker returns below json with status code `202 Accepted` to CF
_{
  "dashboard_url": "http://example-dashboard.example.com/9189kdfsk0vfnku",
  "operation": "task_10"
}_
5) CF polls the Service Instance Last Operation Endpoint for operation status

We need service name that was provided in `cf create-service` command to create service instance on our PAAS.

Is there any solution or workaround for our use case? Or else any option to pass all command line parameters in `cf create-service` command to the service broker’s provisioning endpoint ?

duglin:
I can see value in knowing the name as well, however we'd need to deal with issues like renaming. IMO whenever you allow a user to pick the value of something that is meaningful to them (like 'name') then you have to allow for fat-fingers and that means renames. Delete-n-recreate, to be blunt, sucks as a model.  So, to me update() would need to support changing the name - for instances. Do we need this on bindings too?

So, I'm leaning towards option 1.  Kube doesn't use the user-provided name as the Kube "name" so its not static.

duglin:
@mattmcneeney do you have time to work on a PR for this? If not, I might be able to get someone from our side to take a try

Samze:
In CF we treat the service instance name as a user chosen value. We support the `cf rename-service <old_name> <new_name>` command which currently only updates our internal representation and has no service broker interaction. For either solutions 1 or 2, brokers need to be made aware of renames. Update sounds like the place to do this. 

Note that CF also recently introduced optional named service bindings, as a way to more easily identify a binding in the app container's `VCAP_SERVICES` env, but there is no way today to rename them. 

@jakkahub do you have this requirement for service binding names too?

jakkahub:
[...]

PR #544, opened 2018-07-11
Add name field to provision and update (corresponding to the name being used to represent the Service Instance in the Platform)

mattmcneeney:
See #514 

Needs further discussion so adding the DO NOT MERGE label for now, but please review.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -890,6 +890,20 @@ dedicated database server running on its own VM or an empty schema on a shared
 database server. For non-data services, provisioning could just mean an
 account on an multi-tenant SaaS application.
 
+Platforms MAY choose to include a `name` field when
+[Provisioning](#provisioning) (or [Updating](#updating-a-service-instance)) a
+Service Instance. If present, this field is meant to be for informational
+purposes only (Service Brokers SHOULD NOT use it as any kind of _primary key_
+or _identifier_ for the Service Instance; the `instance_id` field is meant to
+be used for that purpose). The `name` field is made available in order to allow
+Service Brokers to provide an improved user experience, i.e. by showing the
+name of the Service Instance from the Platform (alongside a less user-friendly
+instance ID). Because this field is OPTIONAL, Service Brokers will need to be
+prepared for cases where this field is not provided. Additionally, Platforms
+SHOULD notify Service Brokers if the name of the Service Instance in the
+Platform is modified by sending an [Update](#updating-a-service-instance)
+request with the new `name` field.
+
 ### Request
 
 #### Route
@@ -924,6 +938,7 @@ The following HTTP Headers are defined for this operation:
 | organization_guid* | string | Deprecated in favor of `context`. The Platform GUID for the organization under which the Service Instance is to be provisioned. Although most Service Brokers will not use this field, it might be helpful for executing operations on a user's behalf. MUST be a non-empty string. |
 | space_guid* | string | Deprecated in favor of `context`. The identifier for the project space within the Platform organization. Although most Service Brokers will not use this field, it might be helpful for executing operations on a user's behalf. MUST be a non-empty string. |
 | parameters | object | Configuration parameters for the Service Instance. Service Brokers SHOULD ensure that the client has provided valid configuration parameters and values for the operation. |
+| name | string | The name being used to represent the Service Instance in the Platform. This field is NOT REQUIRED to be unique across all of the Service Instances that have been provisioned through the Service Broker. |
 
 \* Fields with an asterisk are REQUIRED.
 
@@ -940,7 +955,8 @@ The following HTTP Headers are defined for this operation:
   "parameters": {
     "parameter1": 1,
     "parameter2": "foo"
-  }
+  },
+  "name": "instance-1"
 }
 ```
 
@@ -958,7 +974,8 @@ $ curl http://username:password@service-broker-url/v2/service_instances/:instanc
   "parameters": {
     "parameter1": 1,
     "parameter2": "foo"
-  }
+  },
+  "name": "instance-1"
 }' -X PUT -H "X-Broker-API-Version: 2.14" -H "Content-Type: application/json"
 ```
 
@@ -1109,6 +1126,7 @@ The following HTTP Headers are defined for this operation:
 | plan_id | string | If present, MUST be the ID of a plan from the service that has been requested. If this field is not present in the request message, then the Service Broker MUST NOT change the plan of the Service Instance as a result of this request. |
 | parameters | object | Configuration parameters for the Service Instance. Service Brokers SHOULD ensure that the client has provided valid configuration parameters and values for the operation. See "Note" below. |
 | previous_values | [PreviousValues](#previous-values-object) | Information about the Service Instance prior to the update. |
+| name | string | The name being used to represent the Service Instance in the Platform. This field is NOT REQUIRED to be unique across all of the Service Instances that have been provisioned through the Service Broker. |
 
 \* Fields with an asterisk are REQUIRED.
 
@@ -1150,6 +1168,7 @@ the user did not explicitly specify in their request for the update.
     "parameter1": 1,
     "parameter2": "foo"
   },
+  "name": "instance-1",
   "previous_values": {
     "plan_id": "old-plan-id-here",
     "service_id": "service-id-here",
@@ -1172,6 +1191,7 @@ $ curl http://username:password@service-broker-url/v2/service_instances/:instanc
     "parameter1": 1,
     "parameter2": "foo"
   },
+  "name": "instance-1",
   "previous_values": {
     "plan_id": "old-plan-id-here",
     "service_id": "service-id-here",


mattmcneeney:
cc @kieron-pivotal I wanted to flag this as supporting this would require ODB changes

duglin:
I went back to talk to our guys on this and their preferred approach would be to define an API for the Broker to querythe paltform for this info, however since that's a huge change (since we have no broker->platform APIs at all right now), they consider this the next best option.

spec.md
```diff
@@ -818,6 +818,7 @@ The following HTTP Headers are defined for this operation:
 | context | object | Platform specific contextual information under which the Service Instance is to be provisioned. Although most Service Brokers will not use this field, it could be helpful in determining data placement or applying custom business rules. `context` will replace `organization_guid` and `space_guid` in future versions of the specification - in the interim both SHOULD be used to ensure interoperability with old and new implementations. |
 | organization_guid* | string | Deprecated in favor of `context`. The Platform GUID for the organization under which the Service Instance is to be provisioned. Although most Service Brokers will not use this field, it might be helpful for executing operations on a user's behalf. MUST be a non-empty string. |
 | space_guid* | string | Deprecated in favor of `context`. The identifier for the project space within the Platform organization. Although most Service Brokers will not use this field, it might be helpful for executing operations on a user's behalf. MUST be a non-empty string. |
+| name | string | The name being used to represent the Service Instance in the Platform. |
```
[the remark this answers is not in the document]
mattmcneeney: I don't think so. It's an optional field, so brokers should support both cases where this is and isn't provided (like other optional fields).

mattmcneeney:
Jatin asked why we are sending mutable platform metadata on a provision and update call, which opens up an avenue where the platform has to try to keep the broker in sync. This also seems unaligned with the text describing what update should be used for:
>By implementing this endpoint, Service Broker authors can enable users to modify two attributes of an existing Service Instance: the Service Plan and parameters. By changing the Service Plan, users can upgrade or downgrade their Service Instance to other plans. By modifying parameters, users can change configuration options that are specific to a service or plan.

We started revisiting moving this information to the `context` object, understanding that this means we may send update requests with no actual changes.

Discuss...

duglin:
rebase needed

mattmcneeney:
I'll take the action to update this to put the `name` field back in `context`, and we can go from there..

mattmcneeney:
Updated! `name` is now a Cloud Foundry context object field. If the svc-cat folks want to use this too then shout and I can have a go at adding it to the Kubernetes context object too.

Thoughts @duglin @tinygrasshopper?

jagadish-kb:
Context having name certainly helps (we need name in some cases additionally apart from Dashboard). Since context now carries name, can we request for API spec to mandate name changes as well to invoke [broker update api](
https://github.com/openservicebrokerapi/servicebroker/blob/v2.14/spec.md#updating-a-service-instance) ?  

Currently [renaming a service instance](
https://cli.cloudfoundry.org/en-US/cf/rename-service.html) does not result in a call back to service broker. 

One could argue that this is out scope for the API spec and is part of the Platform to either invoke or not invoke it based on their CLI/API set. However I feel making this explicit in the API spec would at-least provide the right guidance to the platform.

tinygrasshopper:
My scepticism around putting mutable identifiers in the api and having to keep them in sync using update service endpoint persists. Currently all other entities in the context map are immutable. 

I think there are also considerations about what to do when the broker update call fails when a user tries to rename a service on the platform, should the platform reject the rename request? In the case of an async update, should the platform update the name on the platform during the time it is polling the broker to check if the update request is successful. Does the spec need to provide any guidance for such scenarios?

zhongyi-zhang:
@krancour @jeremyrickard what do you think of the adding? Looks it overlaps "alias" in [OSBA](https://github.com/Azure/open-service-broker-azure). In other words, OSBA ever wanted such a field. That's why OSBA brought in "alias" as a substitution. Compared to "alias", the instance name can be gave by Cloud Controller / Service Catalog -- users don't need to specify an extra "alias".
It wouldn't impact OSBA if OSBA just ignores the adding. But it would be a breaking change if OSBA tries to refactor to take advantage of this.

mattmcneeney:
I have removed the _do not merge_ label from this PR, as I think it is a fairly simple one for Platform's to implement (ignoring rename...) and provides some value.

cc @williammartin for adding support to CF

Let's try to get this merged!

profile.md
```diff
@@ -189,13 +189,27 @@ deployment:
   "space_guid": "aaaa1234-da91-4f12-8ffa-b51d0336aaaa"
   ```
 
+- `name`
```
duglin: Since 'name' isn't platform specific (which is what I sort of view 'context' for), I'm wondering why we're not putting this into the main body of the request as an optional field?

mattmcneeney: I kind of think `name` _is_ platform specific. There is no requirement for platforms to _name_ service instances (and I think some CLI tools exist that behave as platforms and don't have names).

mattmcneeney:
> @mattmcneeney does this imply that platforms will start making update calls when the name field is changed on the platform? How would that invocation look like, would it be an update without any parameters and plan changes? How would existing brokers 'know' how to react to the additional update invocations in the lifecycle.

Yes - this does imply that Platforms would make an update request to the broker if a service instance is renamed (probably with no parameters and no plan change). Existing brokers MAY reject that request, so we'd have to implement Platform-side logic to handle the fact that this is new behaviour which may not be supported. Service brokers may just return a `200 OK` as per the spec if the instance/binding already exists and no changes are being made.

> I think we should also introduce a new field in the catalog something like name_updatetable, which signifies the service instances name needs to be updated on the broker everytime there is a name change on the platform.

How important do you think it is for brokers to opt-in to this behaviour? I'm not against it, but it does raise the complexity of this feature, and it seems weird to have a flag in the catalog controlling some contextual information that gets sent through.

profile.md
duglin: I probably didn't use the right phrase, I view things like org/space/namespace as environment-specific. While things like instanceID and name (while platform specific) aren't really about the environment as much as they're about the thing being operated on. I view context as "extra info to help provide, well, context about where the resource lives", not so much about the resource itself. The spec say:
- - -
Platform specific contextual information under which the Service Instance is to be provisioned.
- - -
Which I think this is consistent. If we put info about the resource itself I wonder where we draw the line then.... why not put service, plan and parameters under there too?

mattmcneeney:
@duglin I didn't want to change this PR (again) unless we can get agreement on where this info should go.

I know @tinygrasshopper had thoughts too. If I move this to be a top-level field in the create and update requests, do you think we would have agreement?

krancour:
@zhongyi-zhang sorry... just saw your comment from way back on Aug 28. This serves a different purpose than "alias" in OSBA. Email for details if you need some clarification.

zhongyi-zhang:
@krancour yeah, I think so. I just feel like it can be generally used for covering the purpose of `alias`. Anyway, OSBA can keep using `alias`.

mattmcneeney:
@duglin I can see that argument! It also makes the update call a little easier, as again, `name` can be sent at the top-level (though I do still worry some brokers may not like an UPDATE call with just a `name` field, when a user renames a service instance in the platform, but the platform can just be careful to handle errors in these cases).

Update inbound...

mattmcneeney:
@duglin what do you think now?

mattmcneeney:
Updated @duglin

duglin:
From a technical perspective I like it. Thanks.

I'm wondering if we should add some text, someplace, to help explain this field bit more though. I'm worried that people won't realize that it's really just an "FYI" field and not really one that's meant to be used for anything serious - like a key into a DB - unlike our IDs. For example, a person coming from Kube (where "name" is critical) might not realize how temporal (or optional) this field is despite the spec not saying its required.

Perhaps around line 901 we could add something along the lines of:
- - -
Platforms MAY choose to include a `name` field as part of the metadata about a Service Instance. When present, this field is meant to be for informational purposes only. Meaning, Service Brokers are not expected to use it as any kind of "primary key" or identifier for the Service Instance. The `instance_id` is meant to be used for that purpose. The `name` field is made available simply to allow Service Brokers to use it in situations such as a user interface where they would like to show the name of the Service Instance rather than (a less UX friendly) instance ID. Because this field is OPTIONAL, Service Brokers will need to be prepared for cases where this field is not provided.  Additionally, Platforms SHOULD notify Service Brokers of updates to this field (via the `PATCH` operation).
- - -

mattmcneeney:
I took your words and changed them a little @duglin! Feedback please!

spec.md
```diff
@@ -890,6 +890,20 @@ dedicated database server running on its own VM or an empty schema on a shared
...
 account on an multi-tenant SaaS application.
 
+Platforms MAY choose to include a `name` field when
+[Provisioning](#provisioning) (or [Updating](#updating-a-service-instance)) a
+Service Instance. If present, this field is meant to be for informational
+purposes only (Service Brokers SHOULD NOT use it as any kind of _primary key_
+or _identifier_ for the Service Instance; the `instance_id` field is meant to
+be used for that purpose). The `name` field is made available in order to allow
+Service Brokers to provide an improved user experience, i.e. by showing the
+name of the Service Instance from the Platform (alongside a less user-friendly
```
duglin: missing trailing ")"

duglin:
I like it

spec.md
mattmcneeney: fixed!

mattmcneeney:
Discussion point from today's call:
* Should this be an opt-in feature for service brokers? Is this too heavyweight?
* For brokers who don't need the `name` field, they are going to receive an additional call that they don't need to receive
* Should we make this extensible for other platform metadata? Or is the `name` the only important field?

mattmcneeney:
I've thought a little more about the above discussion points, and personally I think this approach now outlined in this PR seems reasonable. Broker authors can ignore the name field if they don't want to use it, and the worst case scenario seems to be that they'd need to handle the case where an update request arrives with no plan or parameters. And any failure that occurs here could be hidden by the platform (we'll probably do this for `cf rename-service`).

What do you think @openservicebrokerapi/osbapi-pmc ?

mattmcneeney:
Ping again @openservicebrokerapi/osbapi-pmc I'd like to get this reviewed and merged if possible!

mattmcneeney:
I'm still not sure what would belong in #620 that isn't covered by either `name` or `context.*`. @tinygrasshopper may disagree with me, but he's on vacation now. We can leave this until January, but personally I think `name` is such a commonly used thing that it could belong at the top level.

mattmcneeney:
Closing this PR as I'm going to create a new one which covers both this and #620

PR #545, opened 2018-07-11
Enforce Content-Type in the specification

### LINKED ISSUE #543: Should the spec be more specific about Content-Type header?
I notice a bunch of the example `curl` requests on the spec set the `Content-Type` to `application/json`. Is this actually part of the specification and if so, should we make that explicit somewhere?

williammartin:
[fixes #543]

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -11,6 +11,7 @@
   - [URL Properties](#url-properties)
   - [Originating Identity](#originating-identity)
   - [Service Broker Errors](#service-broker-errors)
+  - [Content Type](#content-type)
   - [Catalog Management](#catalog-management)
     - [Adding a Service Broker to the Platform](#adding-a-service-broker-to-the-platform)
   - [Synchronous and Asynchronous Operations](#synchronous-and-asynchronous-operations)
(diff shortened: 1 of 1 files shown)


williammartin:
Not totally sure about the RFC verbs here and backwards compatibility but I thought I'd be strict and we can walk our way back from the edge if this is not intended.

williammartin:
I didn't add it to the required headers section of each request...it feels a bit weird to do so and also not to so I decided to do less work.

spec.md
```diff
@@ -292,6 +293,11 @@ use these error codes for the specified failure scenarios.
 | ConcurrencyError | The Service Broker does not support concurrent requests that mutate the same resource. | Clients MUST wait until pending requests have completed for the specified resources. |
 | RequiresApp | The request body is missing the `app_guid` field. | The `app_guid` MUST be included in the request body. |
 
+### Content Type
+
+All requests and responses defined in this specification with accompanying bodies MUST contain a `Content-Type` header set to `application/json`.
+If the `Content-Type` is not set, Service Brokers MUST respond with `400 Bad Request`.
```
leonwanghui: From my understanding, many `http server` implementations would add `Content-Type` header and set its default value to `application/json`, so I'm not quite sure if this constraint is required.

mattmcneeney: For backwards compatibility, we might want to change both MUSTs here to SHOULD.

williammartin: @leonwanghui I don't think we should rely upon the potential behaviour of clients (the go http client for example does not set this header) or servers.

Seems like it's either part of the spec or not, and if a bunch of clients/servers fulfil that with no effort from the user that's great.

williammartin: @mattmcneeney I'm fine with either, I just think it should be specified if it's desirable behaviour (I think it is) because I futzed for an hour around writing a client for the `overview-broker` before I guessed this was what it was looking for. It's codified in all the curl snippets, but not mentioned anywhere else.

duglin:
LGTM

fmui:
LGTM

PR #553, opened 2018-07-12
 Reorder and improve terminology section

Samze:
Just general cleanup for this section.
- Putting terms in logical order
- Being more explicit about what a service instance is. `An instantiation of a Service Offering.` becomes `An instantiation of a Service Offering and Service Plan.`


duglin:
LGTM

PR #555, opened 2018-07-12
Use tags instead of branches for releases

mattmcneeney:


duglin:
Let's just make sure we understand how hot-fixes will work -e.g. do we need to create a branch for that or does git just do it all for us

mattmcneeney:
Sounds good @duglin. I'm pretty sure it's fine to just checkout the tag, add a new commit, add a new tag for the patch, and then push it. @MHBauer did you manage to check if this worked?

MHBauer:
It's going to want you to be on a branch before it allows you to push the tag. So checkout with `git checkout -b update-vWhatever tag-name` then push both the branch and tag, then delete the branch remotely and locally. Silly, but it works.

MHBauer:
Actually, you don't need to push the branch. You just need to be on it locally.

mattmcneeney:
Sounds good. We very very rarely make changes to previous releases anyway, so I think this is OK. WDYT @duglin ?

MHBauer:
edited, please reread the above.

duglin:
So to be clear it would be:
```
$ git checkout -b v2.12.1-tmp v2.12
$ <edit stuff>
$ git add <stuff>
$ git commit -s -m "..."
$ git tag v2.12.1
$ git push upstream v2.12.1
$ git checkout master
$ git branch -D v2.12.1-tmp
```

I just tried this and I think it worked

CONTRIBUTING.md
```diff
@@ -167,13 +167,13 @@ any PMC member:
 1. Merge the release proposal PR into the master branch of the repository. There
    should not be any conflicts as the text in the files that have changed should
    only be changed during this release process.
-1. Create a new branch called **"v$major.$minor"** from the **Release SHA**.
 1. Cherry pick the commit in which the release proposal PR was merged, to pick
```
duglin: Do we still need to cherry pick it since we're still on master it is has the release PR already?

mattmcneeney: Wait I think I still need a line like this - to avoid merging any changes that went onto master after the RC was created. Update en route

CONTRIBUTING.md
```diff
@@ -167,13 +167,13 @@ any PMC member:
...
    up the file changes.
 1. Create a new commit updating [spec.md](spec.md), [profile](profile.md),
```
duglin: Is there any reason to not do this as part of the release PR?

CONTRIBUTING.md
duglin: Since we're just using tags now, shouldn't we just create a PR with all release changes and merge it into master? I know other things "could" get merged but I wonder if it would just be easier to just freeze master while we're in the release process. Otherwise, don't you need two commits for the version string updates (one for the tag and one for master)?

mattmcneeney:
@duglin I don't think you even need to create a local branch in your instructions, i.e.
```
$ git checkout v2.12
$ <edit stuff>
$ git add <stuff>
$ git commit -s -m "..."
$ git tag v2.12.1
$ git push upstream v2.12.1
$ git checkout master
```

duglin:
yep , you're right - just tried it!

CONTRIBUTING.md
mattmcneeney: This is removed!

CONTRIBUTING.md
mattmcneeney: There is still stuff to do after the release PR is merged into master though. Pushing an updated version now. LMK what you think

CONTRIBUTING.md
```diff
@@ -167,13 +180,17 @@ any PMC member:
 1. Merge the release proposal PR into the master branch of the repository. There
    should not be any conflicts as the text in the files that have changed should
    only be changed during this release process.
-1. Create a new branch called **"v$major.$minor"** from the **Release SHA**.
+1. Checkout the **Release SHA** (either to a local branch or in 'detached HEAD'
```
mattmcneeney: @duglin You're right that we could just lock master instead of doing this, but I'm easy either way

CONTRIBUTING.md
```diff
@@ -167,13 +180,17 @@ any PMC member:
...
+   state). This is done in order to ensure that changes that were merged into
+   master after the release candidate was created are not included in the
+   release.
 1. Cherry pick the commit in which the release proposal PR was merged, to pick
    up the file changes.
 1. Create a new commit updating [spec.md](spec.md), [profile](profile.md),
    [openapi.yaml](openapi.yaml) and [swagger.yaml](swagger.yaml) to include the
    version of the release `v$major.$minor`.
-1. Push the branch to the repository (`v$major.$minor`).
+1. Tag the commit with the name `v$major.$minor`.
```
duglin: Won't the SHA of this commit (which is the real commit for the release) be different from what you mention in the previous set of changes above?  Perhaps what's missing is that "Release SHA" should be "Release RC SHA" ?

mattmcneeney: Yes - that makes sense. I'll update!

mattmcneeney: Updated @duglin

duglin:
LGTM

fmui:
LGTM

mattmcneeney:
Minor changes made after running through this when releasing v2.14 and finding some instructions which needed clarification. @openservicebrokerapi/osbapi-pmc

duglin:
LGTM

zrob:
lgtm

PR #559, opened 2018-07-23
fixed Originating Identity example

fmui:
The decoded value of the Originating Identity example also contains a "user_name" entry.

### DIFF
--- a/profile.md
+++ b/profile.md
@@ -78,7 +78,7 @@ For example, a `value` of:
 ```
 would appear in the HTTP Header as:
 ```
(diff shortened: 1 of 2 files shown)


duglin:
LGTM

duglin:
hmm the example is a CF one, so perhaps we need to drop that user_name field from the encoding?  Does CF have a user_name field?

mattmcneeney:
@duglin Not according to the profiles (and not that I know of!)

fmui:
@mattmcneeney If you decode the base64 string that is in the spec today, the JSON contains the field "user_name". We could either don't touch the base64 string and use "user_name" or we change the  base64 string and JSON to something else.

mattmcneeney:
Interesting @fmui - I didn't know that! Can we use this PR to add that to the profile doc as well?

Samze:
It looks like the docs are wrong to me @fmui. At least in CC we only send `user_id`. So decodes to:
```
{
  "user_id": "683ea748-3092-4ff4-b656-39cacc4d5360",
}
```

So it looks like the base64 hash is infact wrong.

fmui:
I've changed the base64 strings.

duglin:
LGTM

zrob:
lgtm

PR #561, opened 2018-07-24
Brokers may return 200 for no-op update request

### LINKED ISSUE #539: What is the broker response for a no-op update?
When the platform sends an update request the broker can return a 200 OK (forgetting async for a moment) to state that has processed the proposed changes (either parameters or plan id). What is not defined is the broker's behaviour if the platform sends another request with the same proposed changes and the broker does nothing.

For parameters we cannot necessarily assume that the endpoint is idempotent because a broker could have a parameter like `"upgrade": true`. But plan changes can be more black or white.

Do we want to add any clarifying text in the spec to guide broker authors on what to do when an update results in a no-op?

Samze:
fixes #539

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1149,7 +1149,7 @@ $ curl http://username:password@service-broker-url/v2/service_instances/:instanc
 
 | Status Code | Description |
 | --- | --- |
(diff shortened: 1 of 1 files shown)


duglin:
LGTM

fmui:
LGTM

PR #562, opened 2018-07-25
GET instance/bindings MAY not return parameters

jberkhahn:
Just making this explicit.


mattmcneeney:
Hey @jberkhahn - what was the reasoning behind making this more explicit?

duglin:
We got some questions from our folks about whether or not they were allowed to exclude sensitive info passed as parameters. They didn't want to violate the spec by not returning all params if that's what the spec was mandating.

duglin:
LGTM

fmui:
LGTM

zrob:
lgtm

PR #567, opened 2018-08-06
Add v2.14 release to compatibility matrix

mattmcneeney:


duglin:
LGTM

Should we update the release process to remind people to do this as part of the normal release workflow?

mattmcneeney:
We could, but it needs to be done _after_ the release is cut so the correct date can be added.

zrob:
lgtm

duglin:
Just one more review needed on this one
@pmorie @n3wscott

Issue #566, opened 2018-08-03
Clarify the state of an instance after a failed update or delete


fmui:
Update and delete operations may fail for many different reasons. For example:
* A deletion may fail because other instances or resources depend on the instance.
* Resources to do the update are not available at this time.
* A legal hold does not allow an instance to be altered or deleted.

In many cases, when such an operation fails, the service instance is untouched and still usable. From an application point of view nothing changed. (In the legal hold case there MUST be no change.)

For the update operation, the spec contains this sentence:

> When the response includes a 4xx status code, the Service Broker MUST NOT apply any of the requested changes to the Service Instance.

That implies the service instance is still usable.

For the delete operation, we have this sentence:

> ...MUST be interpreted as a failure and the Platform MUST remember the Service Instance.

The word "remember” does not specify whether the instance is still operational or in an unusable state.
We should clarify what that means and maybe add another status code (409?) that let the broker tell the platform that the instance is untouched.

fmui:
See also issue #358.

nilebox:
It would also be nice to have similar feature for "create" (provisioning) operation.
Otherwise platform has to always perform orphan mitigation "just in case" after failed async provisioning.

Issue #508, opened 2018-04-20
Should platforms always send deprovision request when failed to update the service instance?


leonwanghui:
## Background
As the spec says:
`If the successful response includes a state of failed then the Platform MUST send a deprovision request to the Service Broker to prevent an orphan being created on the Service Broker`.

## Problem
Here is a use case below:
Service Broker A is published for providing volume service. After user B provisioned a volume service instance, he wants to resize the volume by calling `UpdateInstance` method. Then for some reasons A fails to complete update operation and it has to return `failed` state in the response of polling last operation. According to current spec definition, platform will deprovision the volume service instance automatically, but it is unacceptable for B because it would cause data loss.

## Proposed solution
For this specific scenario, I think both platform and service brokers could do some work to avoid the problem:
* Add a `state` of `rolled back` with the response of polling last operaion to indicate the repair work of service brokers
* Any service broker which supports rolling back feature MAY return this state to inform platforms everything is fine even though the update operation fails
* Platform should  perceive this state and keep the original parameter of service instance

leonwanghui:
@duglin @n3wscott Please take a look at this issue, thanks!

leonwanghui:
Hi @duglin , I got your point but IMO there is still some missing parts int the spec to indicate it. I can explain what I found in async update operation.

Firstly, there is no word to explain how platform would treat 5xx, from what I can see, the platform will try to deprovision the instance as long as it detects the `state` of `failed`.

Secondly, 5xx is normally returned for indicating server errors and it could be used for some internal errors occured in service broker. It's a bit odd to express it as rolled back meaning, one of the problems is those sevice brokers which doesn't support rolled back may return 5xx and then how can platform treat them differently?

I know it is kind of too big change to add another state, but it's the best option to fix the issue if the problem can not be solved the hole.

duglin:
On today's call we talked about removing:
`If the successful response includes a state of failed then the Platform MUST send a deprovision request to the Service Broker to prevent an orphan being created on the Service Broker.` from the spec since we believe this was done in error. This was added back when we only had async provision, and we missed that the text would then apply to async binding - that was a mistake.

Everyone ok with removing it and treating it as  "fix" to the spec even though technical its a breaking change?

leonwanghui:
@duglin It ok to me, thx!

mattmcneeney:
SGTM

duglin:
@leonwanghui did you want to open a PR for this?  I'm not hearing any objection

PR #570, opened 2018-08-07
Clarify the state of an instance after a failed update or delete

fmui:
See issue #566.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -29,6 +29,7 @@
   - [Fetching a Service Binding](#fetching-a-service-binding)
   - [Unbinding](#unbinding)
   - [Deprovisioning](#deprovisioning)
+  - [Corrupt Service Instances](#corrupt-service-instances)
   - [Orphans](#orphans)
 
 ## API Overview
@@ -1698,6 +1699,18 @@ For success responses, the following fields are defined:
   "operation": "task_10"
 }
 ```
+
+## Corrupt Service Instances
+
+When an update or delete operation fails, the Service Instance MAY be corrupt.
+A corrupt instance MAY be misconfigured, in an invalid state, not reachable, or
+not working at all.
+Platforms SHOULD not try to create bindings for this instance anymore.
+Whether or not a corrupt instance can be repaired by, for example, updating it
+again, is undefined.
+Deprovisioning a corrupt instance SHOULD still be possible. A Platform MUST
+remember the Service Instance until it is successfully deprovisioned or it has
+been cleaned up as an orphan.
 
 ## Orphans
 
--- a/spec.md
+++ b/spec.md
@@ -1705,7 +1705,7 @@ For success responses, the following fields are defined:
 When an update or delete operation fails, the Service Instance MAY be corrupt.
 A corrupt instance MAY be misconfigured, in an invalid state, not reachable, or
 not working at all.
-Platforms SHOULD not try to create bindings for this instance anymore.
+Platforms SHOULD NOT try to create bindings for this instance anymore.
 Whether or not a corrupt instance can be repaired by, for example, updating it
 again, is undefined.
 Deprovisioning a corrupt instance SHOULD still be possible. A Platform MUST
--- a/spec.md
+++ b/spec.md
@@ -782,6 +782,7 @@ For success responses, the following fields are defined:
 | --- | --- | --- |
 | state* | string | Valid values are `in progress`, `succeeded`, and `failed`. While `"state": "in progress"`, the Platform SHOULD continue polling. A response with `"state": "succeeded"` or `"state": "failed"` MUST cause the Platform to cease polling. |
 | description | string | A user-facing message that can be used to tell the user details about the status of the operation. If present, MUST be a non-empty string. |
+| instance_corrupt | boolean | For failed update and deprovisioning operations, this field indicates whether the instance is still usable or not. If the value is `true`, the Service Instance MUST be considered corrupt and the Platform SHOULD NOT allow the creation of new bindings. If the value is `false`, the Service Instance is in an unmodified and usable state. The default is true. |
 
 \* Fields with an asterisk are REQUIRED.
 
@@ -1190,9 +1191,13 @@ $ curl http://username:password@service-broker-url/v2/service_instances/:instanc
 | 400 Bad Request | MUST be returned if the request is malformed or missing mandatory data. |
 | 422 Unprocessable entity | MUST be returned if the requested change is not supported or if the request cannot currently be fulfilled due to the state of the Service Instance (e.g. Service Instance utilization is over the quota of the requested plan). Additionally, a `422 Unprocessable Entity` MUST be returned if the Service Broker only supports asynchronous update for the requested plan and the request did not include `?accepts_incomplete=true`; in this case the response body MUST contain a error code `"AsyncRequired"` (see [Service Broker Errors](#service-broker-errors)). The error response MAY include a helpful error message in the `description` field such as `"This Service Plan requires client support for asynchronous service operations."`. |
 
-Responses with any other status code MUST be interpreted as a failure.
 When the response includes a 4xx status code, the Service Broker MUST NOT
-apply any of the requested changes to the Service Instance.
+apply any of the requested changes to the Service Instance and the
+Service Instance MUST be in an unmodified and usable state.
+
+Responses with any other status code MUST be interpreted as a failure.
+The Service Instance MUST be considered corrupt and the Platform SHOULD NOT
+allow the creation of new bindings.
 
 #### Body
 
@@ -1672,8 +1677,13 @@ $ curl 'http://username:password@service-broker-url/v2/service_instances/:instan
 | 410 Gone | MUST be returned if the Service Instance does not exist. |
 | 422 Unprocessable Entity | MUST be returned if the Service Broker only supports asynchronous deprovisioning for the requested plan and the request did not include `?accepts_incomplete=true`. The response body MUST contain error code `"AsyncRequired"` (see [Service Broker Errors](#service-broker-errors)). The error response MAY include a helpful error message in the `description` field such as `"This Service Plan requires client support for asynchronous service operations."`. |
 
+When the response includes a 4xx status code other than 410 Gone, the
+Service Instance MUST be in an unmodified and usable state.
+
 Responses with any other status code MUST be interpreted as a failure and the
-Platform MUST remember the Service Instance.
+Platform MUST remember the Service Instance. The Service Instance MUST be
+considered corrupt and the Platform SHOULD NOT allow the creation of
+new bindings.
 
 #### Body
 
--- a/spec.md
+++ b/spec.md
@@ -782,7 +782,7 @@ For success responses, the following fields are defined:
 | --- | --- | --- |
 | state* | string | Valid values are `in progress`, `succeeded`, and `failed`. While `"state": "in progress"`, the Platform SHOULD continue polling. A response with `"state": "succeeded"` or `"state": "failed"` MUST cause the Platform to cease polling. |
 | description | string | A user-facing message that can be used to tell the user details about the status of the operation. If present, MUST be a non-empty string. |
-| instance_corrupt | boolean | For failed update and deprovisioning operations, this field indicates whether the instance is still usable or not. If the value is `true`, the Service Instance MUST be considered corrupt and the Platform SHOULD NOT allow the creation of new bindings. If the value is `false`, the Service Instance is in an unmodified and usable state. The default is true. |
+| instance_corrupt | boolean | For failed update and deprovisioning operations, this field indicates whether the instance is still usable or not. If the value is `true`, the Service Instance MUST be considered corrupt and the Platform SHOULD NOT allow the creation of new bindings. If the value is `false`, the Service Instance is in an unmodified and usable state. If not specified, the state of the resource is unspecified by this specification. |
 
 \* Fields with an asterisk are REQUIRED.
 
--- a/spec.md
+++ b/spec.md
@@ -783,7 +783,8 @@ For success responses, the following fields are defined:
 | --- | --- | --- |
 | state* | string | Valid values are `in progress`, `succeeded`, and `failed`. While `"state": "in progress"`, the Platform SHOULD continue polling. A response with `"state": "succeeded"` or `"state": "failed"` MUST cause the Platform to cease polling. |
 | description | string | A user-facing message that can be used to tell the user details about the status of the operation. If present, MUST be a non-empty string. |
-| instance_corrupt | boolean | For failed update and deprovisioning operations, this field indicates whether the instance is still usable or not. If the value is `true`, the Service Instance MUST be considered corrupt and the Platform SHOULD NOT allow the creation of new bindings. If the value is `false`, the Service Instance is in an unmodified and usable state. If not specified, the state of the resource is unspecified by this specification. |
+| status_code | number | The HTTP status code that would have been returned if the operation would have been executed synchronously. If the state is `failed` this field SHOULD be present and the value MUST be an integer in the range of 400 to 599. This field MUST NOT be present for any other state. |
+| error | string | An error code as described in the [Service Broker Errors](#service-broker-errors) section. If present, MUST be a non-empty string. If the state is `failed` and there is an error code this field SHOULD be present. This field MUST NOT be present for any other state. |
 
 \* Fields with an asterisk are REQUIRED.
 


spec.md
```diff
@@ -782,6 +782,7 @@ For success responses, the following fields are defined:
 | --- | --- | --- |
 | state* | string | Valid values are `in progress`, `succeeded`, and `failed`. While `"state": "in progress"`, the Platform SHOULD continue polling. A response with `"state": "succeeded"` or `"state": "failed"` MUST cause the Platform to cease polling. |
 | description | string | A user-facing message that can be used to tell the user details about the status of the operation. If present, MUST be a non-empty string. |
+| instance_corrupt | boolean | For failed update and deprovisioning operations, this field indicates whether the instance is still usable or not. If the value is `true`, the Service Instance MUST be considered corrupt and the Platform SHOULD NOT allow the creation of new bindings. If the value is `false`, the Service Instance is in an unmodified and usable state. The default is true. |
```
duglin: Last sentence: When not specified, the default value is `true`.

duglin: If we head down this path, it seems like we'd need to support returning this value in the sync case too, right?

duglin: While I would like to have the spec be precise about the state of the resource upon failure, for backwards compatibility can we? I'm wondering if the last sentence should be:
    If not specified, the state of the resource is unspecified by this specification.

fmui: The sync case is covered by HTTP status codes. 4xx means the instance is still usable, 5xx means the instance is corrupt.

fmui: You are right with the last sentence. But from a practical point of view, it makes no difference. The platform must treat the instance as corrupt.

duglin: ah I missed the correlation to the 400/500 response codes.  Makes sense. thanks

duglin:
LGTM

spec.md
```diff
@@ -782,6 +782,7 @@ For success responses, the following fields are defined:
 | --- | --- | --- |
 | state* | string | Valid values are `in progress`, `succeeded`, and `failed`. While `"state": "in progress"`, the Platform SHOULD continue polling. A response with `"state": "succeeded"` or `"state": "failed"` MUST cause the Platform to cease polling. |
 | description | string | A user-facing message that can be used to tell the user details about the status of the operation. If present, MUST be a non-empty string. |
+| instance_corrupt | boolean | For failed update and deprovisioning operations, this field indicates whether the instance is still usable or not. If the value is `true`, the Service Instance MUST be considered corrupt and the Platform SHOULD NOT allow the creation of new bindings. If the value is `false`, the Service Instance is in an unmodified and usable state. If not specified, the state of the resource is unspecified by this specification. |
```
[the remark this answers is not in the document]
n3wscott: or a new `instance_state` = string field?

pmorie: Are there invalid combinations of `instance_corrupt` and `state`? If so we should be explicit about them. Is it valid to see `state: succeeded` and `instance_corrupt: true`?

Can I update a corrupt instance?

spec.md
```diff
@@ -1190,9 +1191,13 @@ $ curl http://username:password@service-broker-url/v2/service_instances/:instanc
...
 
-Responses with any other status code MUST be interpreted as a failure.
 When the response includes a 4xx status code, the Service Broker MUST NOT
-apply any of the requested changes to the Service Instance.
+apply any of the requested changes to the Service Instance and the
+Service Instance MUST be in an unmodified and usable state.
+
+Responses with any other status code MUST be interpreted as a failure.
+The Service Instance MUST be considered corrupt and the Platform SHOULD NOT
```
pmorie: I think we should have a section on corrupt instances that explains:

- what the meaning of 'corrupt' is
- when you should set `instance_corrupt` as a broker author
- whether corrupt instances can be updated

fmui:
@pmorie Updates and deletes are not very different in this regard. In both cases, the operations fails if a pre-condition check fails. The broker may not have touched the instance at this point. It is still running as before. 

Think of a document management repository that contains documents that must not be deleted for legal reasons. The deletion of the repository must fail. If the instance becomes unusable after such a failure, there is no handle for the end-user to get access to the repository anymore. It would require some admin intervention to either fix this or provide an alternative way to access the repository. That doesn't scale.

A WebHook can minimize the chance of such a situation, but it cannot prevent it. We have to be able to handle that for update and delete operations.

spec.md
fmui: Agreed.

spec.md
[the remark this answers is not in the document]
mattmcneeney: What other states came to mind @n3wscott ?

duglin:
CI failure looks real

spec.md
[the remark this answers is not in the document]
duglin: I don't think we can add a new value to "state" because that would break existing implementations that are only expecting values from the list provided - unless we say that there are only 2 impls (CF and svccat).  TBH I'm not sure we can really make that claim given I know of at least one other implementation of a platform that's not CF or SVCCAT.

duglin: The `For failed update and deprovisioning operations...` text implies it only applies when state="failed" but I agree we should be more explicit about it.

spec.md
duglin: This "Responses with any other status code..." is kind of worrisome. If it said "Any other 5xx error..." then I agree. But what about some 3xx status code? If people agree I can open a new PR to address this since this isn't really part of this PR

spec.md
```diff
@@ -1689,6 +1700,18 @@ For success responses, the following fields are defined:
...
+
+When an update or delete operation fails, the Service Instance MAY be corrupt.
+A corrupt instance MAY be misconfigured, in an invalid state, not reachable, or
+not working at all.
+Platforms SHOULD not try to create bindings for this instance anymore.
+Whether or not a corrupt instance can be repaired by, for example, updating it
+again, is undefined.
+Deprovisioning a corrupt instance SHOULD still be possible. A Platform MUST
+remember the Service Instance until it is successfully deprovisioned or it has
+been cleaned up as an orphan.
```
duglin: It really feels like we need some text here about the Platform performing orphan mitigation. What if we combine #598 into this PR so we get the text just right.

spec.md
```diff
@@ -782,6 +783,7 @@ For success responses, the following fields are defined:
 | --- | --- | --- |
 | state* | string | Valid values are `in progress`, `succeeded`, and `failed`. While `"state": "in progress"`, the Platform SHOULD continue polling. A response with `"state": "succeeded"` or `"state": "failed"` MUST cause the Platform to cease polling. |
 | description | string | A user-facing message that can be used to tell the user details about the status of the operation. If present, MUST be a non-empty string. |
+| instance_corrupt | boolean | For failed update and deprovisioning operations, this field indicates whether the instance is still usable or not. If the value is `true`, the Service Instance MUST be considered corrupt and the Platform SHOULD NOT allow the creation of new bindings. If the value is `false`, the Service Instance is in an unmodified and usable state. If not specified, the state of the resource is unspecified by this specification. |
```
duglin: I know this might sound kind of odd, but I thought I'd toss it out there... this "instance_corrupt" field is sort of the equivalent of returning a 5xx status code in the synchronous case. However, its not as useful because in the state=failed+instance_corrupt=false case we don't know whether its the equivalent of a 400 or 402. All we have is the description text, which is nice for humans but computers can't do a lot with it if its trying to make some decision based on 400 vs 402. The same applies to 5xx error, we don't know which one it is - if that's meaningful.  Would it make more sense to make our sync and async flows more symmetrical and have a "status_code" field instead that's set to the HTTP status code that we would have returned if the flow was synchronous?  Then we don't need to have special spec language to deal with two possible ways of getting back the status code from the op.  The Platform side of it would be easier too because then there's just one type of field to check regardless of sync vs async.

spec.md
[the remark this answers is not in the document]
leonwanghui: @mattmcneeney `rolled_back` See #508

spec.md
fmui: (I assume you mean 422, not 402.)
If we go that route, we should also include the `error` field from the “Service Broker Errors” section.

mattmcneeney:
@fmui I think we've dropped the ball on this one a little. Anything we still need to do?

duglin:
I think @fmui is reworking the PR to return a "statusCode" type of property that matches what the HTTP status code would be if it were synchronous.  But I could be wrong.

fmui:
I have updated the PR as discussed in one of our calls.

duglin:
@fmui I'm sorry for missing this one for today's call - I'll add it to next week's.
All, please review when you get a chance.

spec.md
duglin: @fmui ^^

duglin: actually, does "any other" mean non-4xx or does it mean "not mentioned above in the table or previous paragraph" ?  I think we need to clarify this.

mattmcneeney:
@fmui I'm interested to hear what common status codes you'd expect to see returned (or if there even are common ones).

Also, I'm confused by the sentence `matches what the HTTP status code would be if it were synchronous`. We don't really mean that, right? This isn't to do with sync vs async; it's to do with allowing the broker to specify what error code it would return if it wasn't conforming to the `/v2/service_instances/:guid/last_operation` spec where it has to return a `200 OK` along with a `state` field.

duglin:
@fmui ^^^

fmui:
The idea is to keep the logic on the broker and the platform the same for synchronous and asynchronous calls in case of an error. For synchronous provision calls, the spec defines the HTTP status codes 400, 409, and 422 (not applicable here). Additionally, we have to consider 5xx status codes as well. If an asynchronous provisioning fails, the `status_code` field should provide the appropriate status code from the list above. For example, an invalid parameter combination should result in the status code 400.

duglin:
rebase needed

mattmcneeney:
A number of changes have now been made to this PR, and it has now been rebased, so please can we get another round of reviews from @openservicebrokerapi/osbapi-pmc Thank you!

spec.md
Samze: I also worry that if a Service Broker has an intermittent failure and returns 500s for a period of time then instances that receive update calls during this time will be marked as failed, leaving them unusable. 

I would prefer the broker be explicit about when an instance is corrupt rather than inferring it from a large range of status codes.

spec.md
```diff
@@ -1155,9 +1159,13 @@ $ curl http://username:password@service-broker-url/v2/service_instances/:instanc
 | 400 Bad Request | MUST be returned if the request is malformed or missing mandatory data. |
 | 422 Unprocessable entity | MUST be returned if the requested change is not supported or if the request cannot currently be fulfilled due to the state of the Service Instance (e.g. Service Instance utilization is over the quota of the requested plan). Additionally, a `422 Unprocessable Entity` MUST be returned if the Service Broker only supports asynchronous update for the requested plan and the request did not include `?accepts_incomplete=true`; in this case the response body MUST contain a error code `"AsyncRequired"` (see [Service Broker Errors](#service-broker-errors)). The error response MAY include a helpful error message in the `description` field such as `"This Service Plan requires client support for asynchronous service operations."`. |
 
-Responses with any other status code MUST be interpreted as a failure.
 When the response includes a 4xx status code, the Service Broker MUST NOT
```
Samze: I also worry that if a Service Broker has an intermittent failure and returns 500s for a period of time then instances that receive update calls during this time will be marked as failed, leaving them unusable. 

I would prefer the broker be explicit about when an instance is corrupt rather than inferring it from a large range of status codes.

spec.md
```diff
@@ -1155,9 +1159,13 @@ $ curl http://username:password@service-broker-url/v2/service_instances/:instanc
...
-apply any of the requested changes to the Service Instance.
+apply any of the requested changes to the Service Instance and the
+Service Instance MUST be in an unmodified and usable state.
+
+Responses with any other status code MUST be interpreted as a failure.
+The Service Instance MUST be considered corrupt and the Platform SHOULD NOT
```
Samze: I worry that if a Service Broker has an intermittent failure and returns 500s for a period of time then instances that receive update calls during this time will be marked as failed, leaving them unusable. 

I would prefer the broker be explicit about when an instance is corrupt rather than inferring it. It seems kinda risky.

spec.md
```diff
@@ -768,7 +769,10 @@ For success responses, the following fields are defined:
 | Response Field | Type | Description |
 | --- | --- | --- |
 | state* | string | Valid values are `in progress`, `succeeded`, and `failed`. While `"state": "in progress"`, the Platform SHOULD continue polling. A response with `"state": "succeeded"` or `"state": "failed"` MUST cause the Platform to cease polling. |
+
 | description | string | A user-facing message that can be used to tell the user details about the status of the operation. |
+| status_code | number | The HTTP status code that would have been returned if the operation would have been executed synchronously. If the state is `failed` this field SHOULD be present and the value MUST be an integer in the range of 400 to 599. This field MUST NOT be present for any other state. |
```
mattmcneeney: This is the line that I don't like: `The HTTP status code that would have been returned if the operation would have been executed synchronously.`. That suggests that only the HTTP codes defined for a synchronous response are permitted.

What about `An HTTP code that describes why the operation failed.`?

mattmcneeney: Or, how about we explicitly call out the status codes that can be returned? This would help platforms build a better UX around this:
`200 OK` / `400 Bad Request` / `409 Conflict` seem appropriate from the Provision section.

spec.md
```diff
@@ -768,7 +769,10 @@ For success responses, the following fields are defined:
...
+| error | string | An error code as described in the [Service Broker Errors](#service-broker-errors) section. If present, MUST be a non-empty string. If the state is `failed` and there is an error code this field SHOULD be present. This field MUST NOT be present for any other state. |
```
mattmcneeney: Which error code are we actually expecting to get back here?
`AsyncRequired` = not applicable
`ConcurrencyError` = should be returned to the original request with a `422` as outlined in the Blocking Operations section (surely the broker can decide at that time whether or not the instance is already being operated on?)
`RequiresApp` = not applicable

spec.md
mattmcneeney: I completely agree with @Samze. I'm also starting to question how prevalent this problem is in the wild. Do service brokers ever leave instances as corrupt today? Do they allow this to happen or have they built preventative measures?

mattmcneeney:
Sorry it took me so long to review this again @fmui. I've just added some comments, interested to get your feedback!

spec.md
nilebox: The "instance_corrupt" for failed update concerns me a lot. The expectations of the failed update is that the state of resource will be rolled back to the previous successful state.

nilebox:
@fmui another thing missing in the last operation response is whether the failure is retryable or not, see https://github.com/kubernetes-incubator/service-catalog/issues/2505

Service Catalog currently indefinitely retries after getting 200 response with `state: failed` response. It also performs orphan mitigation every time.

spec.md
duglin: > That suggests that only the HTTP codes defined for a synchronous response are permitted.

The sync case's status codes can be anything, not just what's listed in the table - I don't think we need to list all possible values since we don't in the sync case. As for your proposed alternative text in the first comment, I'm ok with it but that's because I don't see the difference  :-)

spec.md
duglin: Doesn't this call into question the entire orphan mitigation strategy?

duglin:
@fmui what's the status of this? I've been assuming we're waiting for an update, is that true?

spec.md
fmui: @Samze  I agree with you, but that would be an incompatible change. The PR just clarifies the current spec, it doesn't change it. The current spec says, a status code 500 must be interpreted as a failure. Following the spec, orphan mitigation must kick in. Depending on the platform, there is shorter or longer period of time between the failure and orphan mitigation. Within this timeframe, the platform must already today consider the service instance as unusable. Otherwise, why should it trigger the orphan mitigation process later? So, the PR just spells out that platform should treat this instance as broken.

fmui: @mattmcneeney There are brokers that update in multiple steps. If one step fails, they can’t move forwards or backwards and then they wait for orphan mitigation. Sure, they could immediately clean up, but from a platform perspective this is not any different.

spec.md
fmui: You have listed the codes the spec defines and none of them might be applicable here. But a broker can make up its own error code if the spec doesn’t define one.
But the main reason for this field is for parity with an error messages of a synchronous call.

fmui:
@nilebox Interesting. I think create and delete operations should always be repeatable, but maybe not immediately because some human intervention may be required. Updates may not always be repeatable. The failed update might have damaged the instance beyond repair and the instance can only be deleted or has already been deleted by the broker.
So, do we need another flag `retryable` with the values `no`, `immediately`, and `later`?

nilebox:
@fmui to me both "immediately" and "later" are just "retryable". If the update cannot be retried immediately, the broker will just keep returning `state: failed` along with `retryable: true` during this period.

Otherwise it's not clear what does "later" mean, and then we would have to return something like "retryAfter" with a timestamp, which the broker may not actually know as it cannot predict the future :)

fmui:
@nilebox let me try different terms.  Do we need a flag `retryable` with the values `yes`, `no`, and `wait`?
`yes` means retrying is possible. `no` means there is no point in retrying it. `wait` means don’t retry until this operation has been triggered again manually (or something similar). 
My default example for the latter is a content repository with a legal hold. An unprovision request must fail because of the legal hold. Whether the repository should be eventually deleted or not is a business decision, which may take days. The is no point in retrying before the decision is made and it could also be decided not to delete the repository at all.
It is possible to merge `no` and `wait` into one value, but then we have to very clear in the spec that retrying this operation is not possible now, but retriggering the operation later (manually) may be successful.

nilebox:
@fmui thanks for clarification, it makes more sense now. I agree that having distinction for an operation that can be manually retried may be useful, but it would probably make sense to move it to a separate flag, i.e. manually retryable operation would have a boolean flag `retryable: true` and some additional field `retryCondition` with values `"auto"` / `"manual"`, or similar (i.e. some additional attribute that the marketplace/platform may choose to recognize or ignore).

spec.md
duglin: Thinking about this more... I do think it's important that the spec correlate the sync and async flows - meaning the `status_code` isn't just some random HTTP code - if the operation were invoked twice (once sync and once async) then typically people would see the same `status_code` in both cases, just in different locations/messages. W/o that I fear some people will not realize what we're trying to do and come up with different codes between the two flows.

duglin:
> Service Catalog currently indefinitely retries after getting 200 response with state: failed response. It also performs orphan mitigation every time.

FYI - We have customers that are complaining about this behavior. They have cases where they feel like their broker is being flooded with provision() requests even though the broker continually says "no" (state:failed). I believe that CF will stop on the first broker "no" - but perhaps @mattmcneeney can confirm.  Either way, I think this PR will really help because with the status code the Platfrom (svc-cat) should now be able to determine whether to retry or not. Meaning, perhaps something like, retry on 5xx but not on 4xx.

nilebox:
> FYI - We have customers that are complaining about this behavior. They have cases where they feel like their broker is being flooded with provision() requests even though the broker continually says "no" (state:failed). 

FYI, I have changed this behaviour recently, see https://github.com/kubernetes-incubator/service-catalog/pull/2520

But now we have an opposite problem: we treat all failed async operations as non-retryable, so having additional flag `retryable: true` would help there.

mattmcneeney:
Hey @nilebox 

I'm curious if you've had any feedback from service providers that they would like Platforms to retry operations after the last operation has returned `"state": "failed"`.

@duglin mentioned that they have had feedback that this could happen if a broker was being flooded with requests, and I'm wondering if there are any other use cases we know of?

Thanks!

fmui:
Replaced by PR #637.

PR #579, opened 2018-08-14
Fix an inconsistency introduced in v2.14

### LINKED ISSUE #508: Should platforms always send deprovision request when failed to update the service instance?

### LINKED ISSUE #563: Is the 2.14 spec contradictory around orphan mitigation?
Hey all,

I'm trying to understand the orphan mitigation sections of the specification and right now it seems to be contradictory. There may be more cases, but the sections that jumped out to me were 4** status code section in the [orphans](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#orphans) section and the instructions on what to do under certain response codes from the [provision operation](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#response-3)

The relevant snippets of text are:

| Status Code Of Service Broker Response | Platform Interpretation Of Response | Platform Initiates Orphan Mitigation? |
| --- | --- | --- |
| ... | ... | ... |
| All other 4xx | Request rejected | No |
| ... | ... | ... |

vs.

> Responses with any other status code MUST be interpreted as a failure and a deprovision request MUST be sent to the Service Broker to prevent an orphan being created on the Service Broker

where status codes defined in the 4** range are 400, 409, and 422.

Am I missing something here like the table is generic and can be overridden by specific endpoint instructions, or is this contradictory?

Thanks,

Will

duglin:
Inconsistency introduced by: https://github.com/openservicebrokerapi/servicebroker/pull/541

Fixes #563
Fixes #508 

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -973,9 +973,9 @@ $ curl http://username:password@service-broker-url/v2/service_instances/:instanc
 | 409 Conflict | MUST be returned if a Service Instance with the same id already exists but with different attributes. |
 | 422 Unprocessable Entity | MUST be returned if the Service Broker only supports asynchronous provisioning for the requested plan and the request did not include `?accepts_incomplete=true`. The response body MUST contain error code `"AsyncRequired"` (see [Service Broker Errors](#service-broker-errors)). The error response MAY include a helpful error message in the `description` field such as `"This Service Plan requires client support for asynchronous service operations."`. |
 
-Responses with any other status code MUST be interpreted as a failure and a
-deprovision request MUST be sent to the Service Broker to prevent an orphan
-being created on the Service Broker. However, while the platform will attempt
+Responses with any other status code MUST be interpreted as a failure. See
+the [Orphans](#orphans) section for more information related to whether
+orphan mitigation needs to be applied. While a Platform might attempt
 to send a deprovision request, Service Brokers MAY automatically delete
 any resources associated with the failed provisioning request on their own.
 


duglin:
@leonwanghui thanks, I updated the first comment

duglin:
@mattmcneeney @pmorie @n3wscott @zrob review?

gberche-orange:
@duglin I believe the text `If the response contains "state": "failed" then the Platform MUST send a deprovision request to the Service Broker to prevent an orphan being created on the Service Broker. However, while the Platform will attempt to send a deprovision request, Service Brokers MAY automatically delete any resources associated with the failed bind request on their own.`  introduced by #354 into the "Polling Last Operation for Service Instances" section still remains:

https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#body-1

I understand this text should be removed as this will negatively impacts service instance update failures as reported by @leonwanghui in #508 when broker `last operation for service instance` endpoint responds with a 200 status with a `failed` body. I'd then suggest to reopen #508

duglin:
@gberche-orange nice catch!! I opened https://github.com/openservicebrokerapi/servicebroker/pull/591 to try to address it

Issue #432, opened 2018-01-27
Doesn't GET request need authentication?


leonwanghui:
When I had a look at our [OpenAPI doc](http://petstore.swagger.io/?url=https://raw.githubusercontent.com/openservicebrokerapi/servicebroker/v2.13/openapi.yaml#/), I found that not all requests contain ```X-Broker-API-Originating-Identity``` parameter (specifically GET request doesn't contain). So it confuses me if it is safe to expose GET request (currently we have two GET requests) to the public, and IMO authentication SHOULD be guranteed in all user requests to accomplish access controller.

Just challenge me if you have any doubts : )

duglin:
This is actually something we're running into as well. The problem we're running into is that some requests are more system-generated than user-initiated - like GET /v2/catalog.  We're still working on a proposal but our current thinking is that some system ID should be used in those cases.

Keep in mind though, even w/o this header the request is still required to have some kind of auth (even if just basic auth).

leonwanghui:
Sure, please ping me if you finish the proposal, thx!发自我的iPhone------------------ Original ------------------From: Doug Davis  <notifications@github.com>Date: 周六,1月 27,2018 09:16To: openservicebrokerapi/servicebroker <servicebroker@noreply.github.com>Cc: Leon Wang <wanghui71leon@gmail.com>, Author <author@noreply.github.com>Subject: Re: [openservicebrokerapi/servicebroker] Doesn't GET request needauthentication? (#432)This is actually something we're running into as well. The problem we're running into is that some requests are more system-generated than user-initiated - like GET /v2/catalog.  We're still working on a proposal but our current thinking is that some system ID should be used in those cases.
Keep in mind though, even w/o this header the request is still required to have some kind of auth (even if just basic auth).

—You are receiving this because you authored the thread.Reply to this email directly, view it on GitHub, or mute the thread.


  
  




[...]

PR #580, opened 2018-08-14
Recommend that platform initiated request have an identity too

duglin:
Fixes: #432 

See what y'all think of this.  Some minor line-wrapping included in there too.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -257,14 +257,13 @@ will need to provide this identification information to the Service Broker on
 each request. Platforms MAY support this feature, and if they do, they MUST
 adhere to the following:
 - For any OSBAPI request that is the result of an action taken by a Platform's
-  user, there MUST be an associated `X-Broker-API-Originating-Identity` header on that HTTP
-  request.
-- Any OSBAPI request that is not associated with an action from a Platform's
+  user, there MUST be an associated `X-Broker-API-Originating-Identity` header
+  on that HTTP request and its value MUST contain the identity of that user.
+- Any request that is not associated with an action from a Platform's
   user, such as the Platform refetching the catalog, MAY exclude the header from
-  that HTTP request.
-- If present on a request, the `X-Broker-API-Originating-Identity` header MUST contain the
-  identify information for the Platform's user that took the action to cause the
-  request to be sent.
+  that HTTP request. However, it is RECOMMENDED that even Platform initiated
+  requests be associated with some identity for authorization or tracking
+  purposes.
 
 The format of the header MUST be:
 


spec.md
```diff
@@ -259,14 +259,16 @@ will need to provide this identification information to the Service Broker on
...
-  that HTTP request.
-- If present on a request, the `X-Broker-API-Originating-Identity` header MUST contain the
-  identify information for the Platform's user that took the action to cause the
-  request to be sent.
+  that HTTP request. However, it is RECOMMENDED that even Platform initiated
+  requests be associated with some identity for authorization or tracking
+  purposes.
+- If present on a request, the `X-Broker-API-Originating-Identity` header MUST
+  contain the identify information for the Platform's user that took the
+  action to cause the request to be sent.
```
duglin: Matt points out that his sentence is now inconsistent

duglin:
On today's call I was asked to find out which ops IBM needed this info on... we basically want an ID on all operations into the Broker - mainly for auth purposes.

spec.md
```diff
@@ -259,14 +259,15 @@ will need to provide this identification information to the Service Broker on
 each request. Platforms MAY support this feature, and if they do, they MUST
 adhere to the following:
 - For any OSBAPI request that is the result of an action taken by a Platform's
-  user, there MUST be an associated `X-Broker-API-Originating-Identity` header on that HTTP
-  request.
-- Any OSBAPI request that is not associated with an action from a Platform's
+  user, there MUST be an associated `X-Broker-API-Originating-Identity` header
+  on that HTTP request.
+- When a request is associated with an action from a Platform's user, the
```
zrob: @duglin I was curious what this sentence adds to the statement above that operations by a Platform's user must include the header

mattmcneeney:
@duglin Let me know if you get any feedback on what you think a sensible value would be for CF for things like polling last operation:
> However, it is RECOMMENDED that even Platform initiated requests be associated with some identity for authorization or tracking purposes.

spec.md
duglin: I finally remember why I did that! :-)  Its because the first bullet just said the header had to be there, but it didn't mandate that the value had to be associated with the user. I merged that thought into the previous bullet to avoid this confusion in the future.

duglin:
@mattmcneeney we created an admin type of ID and used that for Platform initiated requests

duglin:
While we'd like for platform initiated requests to include an ID, we're ok with closing this PR for now.

However, it would be really nice if CF did include the original user ID on the last_operation() message flows since those really are (originally) initiated by the user.

Issue #278, opened 2017-07-25
Define expected behavior for deprovisioning an instance in a "provisioning" state


krancour:
Suppose provisioning is (asynchronously) ongoing for service instance `x`. Expected behavior is undefined if an attempt is made to deprovision an instance in such a state.

The easy answer is to say this equates to a conflict-- 409. However, I do believe there's probably room here to permit individual brokers to optimize. i.e. Some brokers may be capable of aborting provisioning.

Maybe we could say that brokers _may_ return a 409 in this case, or _may_ return a 200 or 202 if they are able to abort provisioning and immediately handle or accept, respectively, the deprovisioning request? This is just a though-- this scenario may require some discussion.

duglin:
related: https://github.com/openservicebrokerapi/servicebroker/issues/196

Issue #196, opened 2017-05-10
Stopping an async, or any, request


duglin:
While working on https://github.com/openservicebrokerapi/servicebroker/pull/193 and https://github.com/openservicebrokerapi/servicebroker/issues/195 it dawned on me that the spec doesn't really talk much about what to do when things get stuck. For example, what if during a createInstance() call things just hang?  Regardless of whether its an async or sync request.  What is the platform to do?  The current spec doesn't provide any help because it mandates that the platform can not send ANY other request while there's an outstanding request being processed. 

That's part of the reason I worded #193 the way I did - I wanted to loosen it a bit. I wanted to allow for the platform to send a DELETE at any time in order to recover from a hung request. The broker is always free to reject the DELETE request, but this at least gives the platform a way to try to kick/stop things.

This only helps in the createXXX() cases. For ops that can/should not be fixed via a DELETE we'll have to find some other solution, but I wanted to start with the easier problem first... the creates.

So, before I opened a PR I wanted to know what people thought about adding some text to make it clear that:
- a platform is free to send a DELETE at any time - not just because its done but also to stop the processing of an outstanding create() op
- a broker is free to reject the DELETE if its busy with a related op (point to text in #193)

Thoughts?

SimyungYang:
I also ran into such a case. In my case, SB provision operation is processed in an async way, and I didn't implement the last_operation command in a proper way. The service creation was in progress forever. I eventually delete the service instance by "purge-service-instance".

LGTM on the assumption that cf platform will not send a DELETE after createXXX() command is successfully executed. If it does, SB cannot distinguish whether the command is called by cf platform or the user.

In case SB is working on a related job and it doesn't have a proper info(OrderId or VMId) to perform DELETE, it can reject the command.

duglin:
> LGTM on the assumption that cf platform will not send a DELETE after createXXX() command is successfully executed. If it does, SB cannot distinguish whether the command is called by cf platform or the user.

But that's how it deletes it when it wants to remove it.  Also, in the spec there is no distinction between a platform and a user - its really just a platform (or broker client).

SimyungYang:
I thought the cf platform, itself will send a DELETE at anytime after a certain period of time for the management reason to an unsuccessful service instance creation. Not by user's "cf delete-service" execution. That's what I meant. 

> a platform is free to send a DELETE at any time - not just because its done but also to stop the processing of an outstanding create() op

[...]

PR #581, opened 2018-08-15
Clarify what happens when deleting during a provision/bind request

duglin:
Fixes: #196 
Fixes: #278 

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -695,9 +695,10 @@ operation.
 
 ## Blocking Operations
 
-Service Brokers do not have to support concurrent requests that mutate the same
-resource. If a Service Broker receives a request that it is not able to process
-due to other activity being done on that resource then the Service Broker MUST
+Service Brokers MAY choose the degree to which they support concurrent
+requests, ranging from not supporting them at all to only supporting them
+in selective situations. If a Service Broker receives a request that it is
+not able to process due a concurrency issue then the Service Broker MUST
 reject the request with a HTTP `422 Unprocessable Entity` and a response body
 containing error code `"ConcurrencyError"` (see
 [Service Broker Errors](#service-broker-errors)). The error response MAY include
@@ -707,11 +708,6 @@ for this Service Instance is in progress."`.
 Note that per the [Orphans](#orphans) section, this error response does not
 cause orphan mitigation to be initiated. Therefore, Platforms receiving this
 error response SHOULD resend the request at a later time.
-
-Service Broker MAY choose to only support concurrent requests for certain
-operations. For example, a specification compliant Service Broker can choose
-to reject concurrent requests in general but accept a deprovision request
-to halt the processing of a long running provision request.
 
 Brokers MAY choose to treat the creation of a Service Binding as a mutation of
 the corresponding Service Instance - it is an implementation choice. Doing so
@@ -1545,7 +1541,10 @@ Instance failing to authenticate.
 If a Service Broker accepts the request to delete a Service Binding during
 the process of it being created, then it MUST have the net effect of halting
 the current creation process and deleting any resources associated with
-the Service Binding.
+the Service Binding. If the request to create the Service Binding is
+asynchronous, then its `last_operation` response SHOULD return an HTTP
+status code of `200`, a `state` value of `failed` and a `description`
+that indicates the create failed due to a concurrent delete request.
 
 ### Request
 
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -710,6 +710,11 @@ Note that per the [Orphans](#orphans) section, this error response does not
 cause orphan mitigation to be initiated. Therefore, Platforms receiving this
 error response SHOULD resend the request at a later time.
 
+Service Broker MAY choose to only support concurrent requests for certain
```
zrob: this paragraph seems to be stating what is already true in the spec.  is it necessary?

mattmcneeney: I agree. The spec already says `Service Brokers do not have to support concurrent requests that mutate the same resource.`. I like the addition of this behaviour clarification in general, but maybe this paragraph is redundant?

spec.md
```diff
@@ -1544,6 +1549,11 @@ delete any resources associated with the Service Binding. In the case where
 credentials were generated, this might result in requests to the Service
 Instance failing to authenticate.
 
+If a Service Broker accepts the request to delete a Service Binding during
+the process of it being created, then it MUST have the net effect of halting
+the current creation process and deleting any resources associated with
+the Service Binding.
```
gberche-orange: Does this imply that the broker MUST consistently return consistent `410 Gone` responses to the `Polling Last Operation for Service Bindings` endpoint calls (i.e. prevent race conditions among binding creation and binding deletion)? Instead could a compliant broker be allowed to serialize delete binding execution after the binding creation completes ? 

In other words, should a client `Polling Last Operation for Service Bindings` endpoint potentially expect to receive a `200 OK` and then when polled again receive a `410 Gone` response?

spec.md
duglin: True, but I wasn't sure if it was clear that a broker could be selective about it - meaning only support it on certain resources or even just for certain ops. Let me try to morph the existing sentence to add what I think is missing and see what you think...

spec.md
duglin: hmmm great question. I think a 410 might be appropriate but it might also really confuse existing brokers. I wonder if we should require the last_op to return a 200 + a status of "failed", and then the "description" can include something like "failed due to a delete request".  Let me try that and so you guys can see how it looks.

spec.md
```diff
@@ -695,9 +695,10 @@ operation.
 
 ## Blocking Operations
 
-Service Brokers do not have to support concurrent requests that mutate the same
-resource. If a Service Broker receives a request that it is not able to process
-due to other activity being done on that resource then the Service Broker MUST
+Service Brokers MAY choose the degree to which they support concurrent
+requests, ranging from not supporing them at all to only supporting them
```
mattmcneeney: Pointed out by @leonwanghui:
s/supporing/supporting

(Thanks!)

mattmcneeney:
@duglin If a platform made a `DELETE` call during an async provision, what behaviour should the platform see for a broker which supports this functionality?
1. If the platform gets a `200 OK` or `410 Gone`, can it assume the create operation has been cancelled and no resources remain?
2. If the platform gets a `202 Accepted`, and hits the `/last_operation` endpoint, should it expect to see `delete in progress` until the delete has finished? From reading the PR now, it sounds like the `/last_operation` endpoint should immediately show it has been deleted. Are we only supporting synchronous cancellation?

duglin:
@mattmcneeney 
1 - yes
2 - why do you think that async delete isn't supported? I think an async-concurrent-delete would have the LastOp return either a 410 if the resource is already deleted, or a 200 + state="in progress" or a 200 + state="success" after its been deleted.

spec.md
```diff
@@ -1616,6 +1625,14 @@ deprovision the service. This specification does not specify what a Service
 Broker is to do if it receives a deprovision request while there are still
 Service Bindings associated with it.
 
+If a Service Broker accepts the request to delete a Service Instance during
+the process of it being created, then it MUST have the net effect of halting
+the current creation process and deleting any resources associated with
+the Service Instance. If the request to create the Service Instance is
+asynchronous, then its `last_operation` response SHOULD return an HTTP
```
kibbles-n-bytes: If the broker is doing a bind asynchronously, and then it accepts an unbind asynchronously, I'd expect the `last_operation` endpoint to show the most recent operation, which is the unbind. So it seems wrong to have the `last_operation` then reflect the bind rather than the unbind. If there is an operation string to pass in, then the bind operation should definitely have that status, though.

duglin:
ping @kibbles-n-bytes @tinygrasshopper @zrob for a review

spec.md
duglin: @kibbles-n-bytes I always assumed that the "operation" would be message-exchange specific - meaning the bind() and unbind() would have different "operation" values. The spec says: "A Service Broker-provided identifier for the operation."  And I *think* "operation" in that sentence means bind() vs unbind().  Or am I not following your question?

duglin:
Needs 2 more reviews

duglin:
@kibbles-n-bytes @n3wscott @zrob needs just one more

zrob:
lgtm

PR #586, opened 2018-08-20
Added network metadata

fmui:
As discussed at the last F2F meeting, here is the PR to include network metadata and adapt credentials, if necessary.

This document provides the background: https://docs.google.com/document/d/1XONqDEA-zIGlT6NnM1nTPd0ZmUeqm2cbRnRk8E386eE

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -863,7 +863,7 @@ This specification does not describe any of them. Instead, it provides
 extension points that allows the service side and the application side to
 negotiate the connection type and exchange the required configuration data.
 
-It is not excepted that the Service Broker deals with the network
+It is not expected that the Service Broker deals with the network
 configuration. A network management component on the Service Broker side SHOULD
 intercept the OSBAPI calls and SHOULD provide and act on these extensions,
 if the service does not expose public endpoints.
@@ -1380,7 +1380,7 @@ The following HTTP Headers are defined for this operation:
 | app_guid | string | Deprecated in favor of `bind_resource.app_guid`. GUID of an application associated with the Service Binding to be created. If present, MUST be a non-empty string. |
 | bind_resource | [BindResource](#bind-resource-object) | A JSON object that contains data for Platform resources associated with the Service Binding to be created. See [Bind Resource Object](#bind-resource-object) for more information. |
 | parameters | object | Configuration parameters for the Service Binding. Service Brokers SHOULD ensure that the client has provided valid configuration parameters and values for the operation. |
-| network_data | object | Network configuration data depending on the Network Profile. This object MUST contain a field `network_profile_id`. The value of this field MUST be the ID of the Network Profile. |
+| network_data | array of [NetworkData](#network-data-object) objects | Network configuration data depending on the Network Profile. |
 
 \* Fields with an asterisk are REQUIRED.
 
@@ -1406,6 +1406,15 @@ while in Kubernetes it might map to a "namespace". The scope of what a
 Platform maps the `app_guid` to is Platform specific and MAY vary across
 binding requests.
 
+##### Network Data Object 
+
+| Request Field | Type | Description |
+| --- | --- | --- |
+| id* | string | MUST be the ID of the Network Profile. |
+| data | object | Network configuration data. |
+
+\* Fields with an asterisk are REQUIRED.
+
 ```
 {
   "context": {
@@ -1422,7 +1431,7 @@ binding requests.
     "parameter2-name-here": "parameter2-value-here"
   },
   "network_data": {
-    "network_profile_id": "urn:something:vpn"
+    "id": "urn:something:vpn"
   }
 }
 ```
@@ -1445,7 +1454,7 @@ $ curl http://username:password@service-broker-url/v2/service_instances/:instanc
     "parameter2-name-here": "parameter2-value-here"
   },
   "network_data": {
-    "network_profile_id": "urn:something:vpn"
+    "id": "urn:something:vpn"
   }
 }' -X PUT -H "X-Broker-API-Version: 2.13" -H "Content-Type: application/json"
 ```
@@ -1484,7 +1493,7 @@ For `200 OK` and `201 Created` response codes, the following fields are defined:
 | route_service_url | string | A URL to which the Platform MUST proxy requests for the address sent with `bind_resource.route` in the request body. `"requires":["route_forwarding"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
 | volume_mounts | array of [VolumeMount](#volume-mount-object) objects | An array of configuration for remote storage devices to be mounted into an application container filesystem. `"requires":["volume_mount"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
 | endpoints | array of [Endpoint](#endpoint-object) objects | A list of network endpoints. The endpoints need not to be reachable and the host names need not to resolvable from outside the service network. |
-| network_data | object | Network configuration data depending on the Network Profile. This object MUST contain a field `network_profile_id`. The value of this field MUST be the ID of the Network Profile. |
+| network_data | array of [NetworkData](#network-data-object) objects | Network configuration data depending on the Network Profile. |
 
 ##### Volume Mount Object
 
@@ -1515,7 +1524,7 @@ can be mounted on all app instances simultaneously.
 | Response Field | Type | Description |
 | --- | --- | --- |
 | host* | string | A host name, a single IP address, an IP address range like 192.0.2.0-192.0.2.50, or a CIDR block. |
-| ports* | string | A single port, multiple comma-separated ports, or a single range of ports.  |
+| ports* | array of strings | A non-empty array. Each element is either a single port (for example "443") or a port range (for example "9000-9010"). |
 | protocol | string | The protocol. Valid values are `tcp`, `udp`, or `all`. The default value is `tcp`. |
 
 \* Fields with an asterisk are REQUIRED.
@@ -1533,12 +1542,14 @@ can be mounted on all app instances simultaneously.
   "endpoints": [
     {
       "host": "mysqlhost",
-      "ports:" "3306"
+      "ports:" ["3306"]
     }
   ],
   "network_data": {
-    "network_profile_id": "urn:something:vpn", 
-    "gateway": "example.org:12345"
+    "id": "urn:something:vpn", 
+    "data": { 
+      "gateway": "example.org:12345"
+    }
   }
 }
 ```
@@ -1552,7 +1563,7 @@ can be mounted on all app instances simultaneously.
     "device_type": "shared",
     "device": {
       "volume_id": "bc2c1eab-05b9-482d-b0cf-750ee07de311",
-      "mount_config": {
+      "mount_config": { 
         "key": "value"
       }
     }
@@ -1649,16 +1660,6 @@ endpoints to the new endpoints.
 `:binding_id` MUST be the ID of a previously provisioned Service Binding for that
 instance.
 
-#### Headers
-
-The following HTTP Headers are defined for this operation:
-
-| Header | Type | Description |
-| --- | --- | --- |
-| X-Broker-API-Version* | string | See [API Version Header](#api-version-header). |
-
-\* Headers with an asterisk are REQUIRED.
-
 #### Body
 | Request Field | Type | Description |
 | --- | --- | --- |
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -1373,6 +1482,8 @@ For `200 OK` and `201 Created` response codes, the following fields are defined:
 | syslog_drain_url | string | A URL to which logs MUST be streamed. `"requires":["syslog_drain"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
 | route_service_url | string | A URL to which the Platform MUST proxy requests for the address sent with `bind_resource.route` in the request body. `"requires":["route_forwarding"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
 | volume_mounts | array of [VolumeMount](#volume-mount-object) objects | An array of configuration for remote storage devices to be mounted into an application container filesystem. `"requires":["volume_mount"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
+| endpoints | array of strings | A list of network endpoints. The format of the strings is `<hostname or IP>:<port>`. The endpoints need not to be reachable and the host names need not to resolvable from outside the service network. |
```
holgero: We would propose to have a struct of host and port instead of the concatenated strings.

spec.md
```diff
@@ -1407,6 +1518,13 @@ can be mounted on all app instances simultaneously.
     "host": "mysqlhost",
     "port": 3306,
     "database": "dbname"
+  },
+  "endpoints": [
+    "mysqlhost:3306"
```
holgero: Following the proposal to use a struct instead of a concatenated string this would be: `{ "host": "mysqlhost", "port": 3306 }`

spec.md
```diff
@@ -1495,6 +1613,106 @@ For success responses, the following fields are defined:
...
+| credentials* | object | The free-form hash of credentials that the Platform has received from the Service Broker. |
+| endpoint_mappings* | array of [EndpointMapping](#endpoint-mapping-object) objects | A list of endpoint mapping. |
+
+\* Headers with an asterisk are REQUIRED.
+
+##### Endpoint Mapping Object
+
+| Request Field | Type | Description |
+| --- | --- | --- |
+| source* | string | The endpoint on the service side. The format of the strings is `<hostname or IP>:<port>`. |
```
holgero: We would propose to have a struct of host and port instead of the concatenated strings also in the endpoint mapping object.

spec.md
```diff
@@ -1495,6 +1613,106 @@ For success responses, the following fields are defined:
...
+    "uri": "mysql://mysqluser:pass@mysqlhost:3306/dbname",
+    "username": "mysqluser",
+    "password": "pass",
+    "host": "mysqlhost",
+    "port": 3306,
+    "database": "dbname"
+  },
+  "endpoint_mappings": [
+    {
+      "source": "mysqlhost:3306",
```
holgero: With the proposed struct this would look like this: `{ "source": { "host": "mysqlhost", "port": 3306 }, "target": { "host": "appnethost", "port": 9876 } }`

spec.md
```diff
@@ -1495,6 +1613,106 @@ For success responses, the following fields are defined:
...
+
+The Platform requests the corrected credentials from the Service Broker by
+providing the original credentials and a table that maps the original
+endpoints to the new endpoints.
+
+### Request
+
+##### Route
+
+`PUT /v2/service_instances/:instance_id/service_bindings/:binding_id/credentials`
```
holgero: Wouldn't a `POST` to the path `/v2/service_instances/:instance_id/service_bindings/:binding_id/update_credentials` be preferable?
`PUT` to the path `/v2/service_instances/:instance_id/service_bindings/:binding_id/credentials` might imply an update of the resource credentials (which exists in the binding) and also the possibility to do a `GET` `/v2/service_instances/:instance_id/service_bindings/:binding_id/credentials`.

spec.md
```diff
@@ -21,12 +21,14 @@
...
   - [Polling Last Operation for Service Bindings](#polling-last-operation-for-service-bindings)
   - [Polling Interval and Duration](#polling-interval-and-duration)
+  - [Network Metadata](#network-metadata)
   - [Provisioning](#provisioning)
   - [Fetching a Service Instance](#fetching-a-service-instance)
   - [Updating a Service Instance](#updating-a-service-instance)
   - [Binding](#binding)
     - [Types of Binding](#types-of-binding)
   - [Fetching a Service Binding](#fetching-a-service-binding)
+  - [Adapting Service Binding Credentials](adapting-service-binding-credentials)
```
holgero: The hash sign is missing in the anchor, this seems to make the travis verification job fail. (`#adapting-service-binding-credentials`).

spec.md
```diff
@@ -1495,6 +1626,112 @@ For success responses, the following fields are defined:
...
+## Adapting Service Binding Credentials
+
+If the endpoints of a Service Instance are from an Application point of view
+different to the endpoints provided in the binding credentials, then the
+binding credentials MUST be adapted before the Platform provides them to
+the Application.
+
+The Platform requests the corrected credentials from the Service Broker by
+providing the original credentials and a table that maps the original
+endpoints to the new endpoints.
```
liorokman: The platform itself is not necessarily always able to create this table. Depending on the underlying transport mechanism, sometimes the platform is only able to control the correct target hostname, while the service broker controls the correct target port. In these cases, the platform would be unable to request an adaptation as specified here, since it would be unable to create a valid mapping table.

spec.md
```diff
@@ -1495,6 +1626,112 @@ For success responses, the following fields are defined:
...
+    "password": "pass",
+    "host": "appnethost",
+    "port": 9876,
+    "database": "dbname"
+  }
+}
+```
+
+
 ## Unbinding
```
liorokman: Unbinding operations should also contain enough information in the request and response to undo any network operations which occurred during the bind operations.

spec.md
```diff
@@ -954,11 +1049,18 @@ For success responses, the following fields are defined:
 | --- | --- | --- |
 | dashboard_url | string | The URL of a web-based management user interface for the Service Instance; we refer to this as a service dashboard. The URL MUST contain enough information for the dashboard to identify the resource being accessed (`9189kdfsk0vfnku` in the example below). Note: a Service Broker that wishes to return `dashboard_url` for a Service Instance MUST return it with the initial response to the provision request, even if the service is provisioned asynchronously. If present, MUST be a non-empty string. |
```
liorokman: Is there any requirement that a provided dashboard-url be also reachable over any negotiated network connectivity? If so, how is this accomplished?

spec.md
```diff
@@ -1398,6 +1509,16 @@ can be mounted on all app instances simultaneously.
 
 \* Fields with an asterisk are REQUIRED.
 
+##### Endpoint Object
+
+| Response Field | Type | Description |
+| --- | --- | --- |
+| host* | string | The host name or IP. |
+| port* | string | The port. |
+| proto | string | The protocol. Valid values are `tcp` and `udp`. The default value is `tcp`. |
```
liorokman: Why not require that the proto field be filled with a protocol as mapped in the IANA protocol registry ( https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml , ``/etc/protocols`` )

spec.md
```diff
@@ -1398,6 +1510,16 @@ can be mounted on all app instances simultaneously.
 
 \* Fields with an asterisk are REQUIRED.
 
+##### Endpoint Object
+
+| Response Field | Type | Description |
+| --- | --- | --- |
+| host* | string | A host name, a single IP address, an IP address range like 192.0.2.0-192.0.2.50, or a CIDR block. |
+| ports* | string | A single port, multiple comma-separated ports, or a single range of ports.  |
```
loewenstein: Is there a reason to rule out multiple ranges? I would make this an array (as indicated by the plural key `ports`) of either port or port range.
E.g.
```
...
"ports": [80,443,"9000-9005"]
...
```
That should allow all cases.

spec.md
```diff
@@ -1277,6 +1380,7 @@ The following HTTP Headers are defined for this operation:
 | app_guid | string | Deprecated in favor of `bind_resource.app_guid`. GUID of an application associated with the Service Binding to be created. If present, MUST be a non-empty string. |
 | bind_resource | [BindResource](#bind-resource-object) | A JSON object that contains data for Platform resources associated with the Service Binding to be created. See [Bind Resource Object](#bind-resource-object) for more information. |
 | parameters | object | Configuration parameters for the Service Binding. Service Brokers SHOULD ensure that the client has provided valid configuration parameters and values for the operation. |
+| network_data | object | Network configuration data depending on the Network Profile. This object MUST contain a field `network_profile_id`. The value of this field MUST be the ID of the Network Profile. |
```
loewenstein: We should separate opaque network data from spec data like network_profile_id.
Could we introduce a `data` key just like in the network_profiles? E.g.
```
...
"network_data": {
  "network_profile_id": "urn:something:vpn",
  "data": {
    "profile-specific-key": "value"
  }
}
...
```

spec.md
liorokman: The way this is specified right now creates an ambiguity. 

If the endpoint object contains: 
```{ "host": "192.168.50.1/24" , "port" : "10-20" } ```
Does this mean that all these 10 ports on all these hosts are part of the endpoint? How is this supposed to be wired on the network level? 

If the actual binding then contains only 1 host and 1 port from the range specified by a CIDR and port-range. I recommend disallowing more than a single endpoint at a time (single host:port), and allowing multiple of these. That would make this clearer, and also easier to work with.

File: spec.md [hunk 13]
```diff
@@ -847,6 +849,78 @@ The frequency and maximum duration of polling MAY vary by Platform client. If
...
+application side might be necessary to enable a network connection.
+
+There are many different types of connection configurations and network setups
+(firewalls, load balancers, security groups, VPNs, IaaS specific
+configurations, leased lines, etc.).
+This specification does not describe any of them. Instead, it provides
+extension points that allows the service side and the application side to
+negotiate the connection type and exchange the required configuration data.
+
+It is not excepted that the Service Broker deals with the network
```
williammartin: excepted -> expected

spec.md
fmui: @liorokman If your network profile describes, for example, the configuration of a firewall, opening a port range across a CIDR is a use case.

liorokman:
Hi Florian,

In the context this is written, the Endpoint object is used to describe in
a binding exactly which host:port combinations are being used by a specific
instance of a service. In this context, in my mind ranges are problematic.
A service doesn't describe a configuration of a firewall, and it shouldn't
be allowed to specify huge ranges which have to be opened. Services specify
specific endpoints, allowing consumers to connect to well-known established
host:port combinations.

The example of a firewall configuration section makes sense in the
network_data object, not in the Endpoints array.

--
Lior

spec.md
```diff
@@ -847,6 +849,78 @@ The frequency and maximum duration of polling MAY vary by Platform client. If
...
+
+The Platform MAY need to open new endpoints for the Service, which are
+reachable from the Application. Therefore, the Platform has to know from the
+Service Broker, which endpoints have been opened by the Service Instance.
+This list of endpoints is also part of the binding response.
+
+### Adaptation of the Binding Credentials
+
+If the Platform established new endpoints, then the binding credentials have to
+adapted before they are handed over to the Application. Because only the
```
gogolok: adapted -> be adapted ?

wolfoo2931:
Hello @fmui, 

for better understanding I've added an additional drawing into the google docs. Does it describe a correct use case? 

Best Regards,
Oliver

duglin:
Spoke with the IBM Cloud team and they are interested in this so we'll be doing some in-depth reviews.

ping @rboykin @phanziii

spec.md
```diff
@@ -1495,6 +1638,108 @@ For success responses, the following fields are defined:
...
+\* Headers with an asterisk are REQUIRED.
+
+##### Endpoint Mapping Object
+
+| Request Field | Type | Description |
+| --- | --- | --- |
+| source* | [Endpoint](#endpoint-object) object | The endpoint on the service side. |
+| target* | [Endpoint](#endpoint-object) object | The mapped endpoint on the application side. |
+
+The `ports` arrays in both endpoints MUST be of the same size and the order of
```
htrmeira: Do you prefer to have this same index for host and port or another object, like: `[{ "host": "appnethost", "port": 9876}]`?

fmui:
I'm splitting this PR into multiple pieces:
1. Add binding endpoints #632

PR #587, opened 2018-08-21
Fix example response for fetch binding

tinygrasshopper:
* fix the example response to be valid json
* remove the operation example response, as fetching binding is not async


duglin:
easy review: @fmui @mattmcneeney @n3wscott @pmorie @zrob

PR #589, opened 2018-08-21
Remove breaking TLS text

### LINKED ISSUE #571: Is the spec contradictory around TLS requirements?
In the [platform to service broker auth section](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#platform-to-service-broker-authentication), there seems to be some contradictory text:

> While the communication between a Platform and Service Broker MAY be unsecure, it is RECOMMENDED that all communications between a Platform and a Service Broker are secured via TLS and authenticated.

vs.

> Additionally, the Service Broker MUST secure communications with TLS. The Platform and Service Broker SHOULD agree whether the Service Broker will use a root-signed certificate or a self-signed certificate.

So, is it recommended or required?

williammartin:
There was originally no requirement for TLS when using Basic
Authentication.

There is a requirement for TLS when using the out of
spec Bearer Token scheme, but this requrement is defined
in the RFC and does not need to be called out explicitly.

[fixes #571]

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -169,7 +169,9 @@ no obligation to understand or process unknown extension fields.
 
 While the communication between a Platform and Service Broker MAY be unsecure,
 it is RECOMMENDED that all communications between a Platform and a Service
-Broker are secured via TLS and authenticated.
+Broker are secured via TLS and authenticated. If communications are secured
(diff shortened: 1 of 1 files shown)


duglin:
LGTM

Issue #584, opened 2018-08-16
Add a unique-request-id header


jberkhahn:
Would it be useful to have an optional header that is unique per-request, similar to Cloud Foundry's VCAP-request-ID? Basically an ID that is unique per request for logging and tracing purposes. In CF this is a unique ID used from the time the user submits a request to the front facing api server all the way through. It would be up to the platform to generate and track these IDs how they wanted.

Thoughts?

duglin:
+1
There are other HTTP headers that we've talked about being useful ( like etag ), so it might nice to add a section to the spec to mention features (probably http features) that people may want to consider using. While we're not defining them as part of the spec, I think its useful to call them out (w/o duplicating their definition) so that people know that they're potentially good ideas to consider in their implementations.

mattmcneeney:
This sounds sensible to me.

duglin:
no concerns from people on the 8/21 call - so PR should come soon for further discussions

mattmcneeney:
Great. Thanks @jberkhahn !

Issue #585, opened 2018-08-20
Section about HTTP headers is missing for a few GET requests


fmui:
The spec does not define the HTTP headers for "Polling Last Operation for Service Bindings" and "Fetching a Service Instance".  They should match the headers definition of in the "Catalog Management" section.

duglin:
It appears that all of the operation "Headers" sections are basically the same, or should be - some are missing the Originating Identity header. What do people think about creating a global HTTP Headers section where we talk about the headers in general, for all ops. And remove the op-specific header sections. Then we can include other headers (optional HTTP-defined headers) that people might consider using.  If, in the future, some operation has its own HTTP headers (e.g. perhaps etag for `get /catalog`) then we can have an op-specific Headers section.

Also see: #584

fmui:
+1 for having a global headers section.

mattmcneeney:
@duglin is going to put a PR together for this!

mattmcneeney:
This is done and can be closed!

PR #593, opened 2018-08-28
Request Identity Header

jberkhahn:
per #584 

I also rejiggered some of the header sections to try to make them more consistently in the same spot for each operation entry.


jberkhahn:
Actually, closing this and going to do #585 first

Issue #582, opened 2018-08-15
Add plan_updateable to plans


duglin:
Right now `plan_updateable` is a flag on the Service in the Catalog, and it tells the Platform whether or not people can switch plans for that Service. However, that doesn't really help w.r.t. knowing if people can switch from any particular plan.

Proposal: add an "updateable" (yes with misspelling) to the Plan struct so that Platforms can know if users will be allowed to update services of *this* plan to another. Granted it won't tell them which ones they can migrate to, but its at least better than nothing like they have today. This also would not remove the need for the Broker to reject an upgrade if the Platform ignore this new field.

What do people think?

duglin:
correct

mattmcneeney:
How many use cases do we know of where only some plan migrations are allowed (i.e. `small` to `large` is allowed, but `large` to `small` isn't)? I'll do some digging on my side to see if this is a problem for any of the brokers I know of

duglin:
I'm not sure how often it does happen but it happens enough for us that our product guys are asking for it :-)

jmrodri:
This is related to https://github.com/openservicebrokerapi/servicebroker/issues/217

MHBauer:
Why with misspelling?

mattmcneeney:
Closing this issue as it was solved in #595

PR #595, opened 2018-08-29
Add updateable field to plan object

jberkhahn:
per #582

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -447,7 +447,7 @@ It is therefore RECOMMENDED that implementations avoid such strings.
 | bindings_retrievable | boolean | Specifies whether the [Fetching a Service Binding](#fetching-a-service-binding) endpoint is supported for all plans. |
 | metadata | object | An opaque object of metadata for a Service Offering. It is expected that Platforms will treat this as a blob. Note that there are [conventions](profile.md#service-metadata) in existing Service Brokers and Platforms for fields that aid in the display of catalog data. |
 | dashboard_client | [DashboardClient](profile.md#dashboard-client-object) | A Cloud Foundry extension described in [Catalog Extensions](profile.md#catalog-extensions). Contains the data necessary to activate the Dashboard SSO feature for this service. |
-| plan_updateable | boolean | Whether the service supports upgrade/downgrade for some plans. Please note that the misspelling of the attribute `plan_updatable` as `plan_updateable` was done by mistake. We have opted to keep that misspelling instead of fixing it and thus breaking backward compatibility. Defaults to false. |
+| plan_updateable | boolean | Whether the Service Offering supports upgrade/downgrade for Service Plans by default. Service Plans can override this field (see [Service Plan](#plan-object)). Please note that the misspelling of the attribute `plan_updatable` as `plan_updateable` was done by mistake. We have opted to keep that misspelling instead of fixing it and thus breaking backward compatibility. Defaults to false. |
 | plans* | array of [Plan](#plan-object) objects | A list of plans for this service, schema is defined below. MUST contain at least one plan. |
 
 \* Fields with an asterisk are REQUIRED.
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -471,6 +471,7 @@ how Platforms might expose these values to their users.
 | metadata | object | An opaque object of metadata for a Service Plan. It is expected that Platforms will treat this as a blob. Note that there are [conventions](profile.md#service-metadata) in existing Service Brokers and Platforms for fields that aid in the display of catalog data. |
 | free | boolean | When false, Service Instances of this plan have a cost. The default is true. |
 | bindable | boolean | Specifies whether Service Instances of the Service Plan can be bound to applications. This field is OPTIONAL. If specified, this takes precedence over the `bindable` attribute of the service. If not specified, the default is derived from the service. |
+| updateable | boolean | Whether the service supports upgrade/downgrade for some Service Plans. Please note that the attribute is intentionally misspelled as `updateable` for legacy reasons. Defaults to false. |
```
MHBauer: Two things:
1. Is it really upgrade/downgrade, or is it 'different'?
2. Is matching an existing fields spelling for legacy? That seems strange, but also correct to me.

duglin: Thinking more about this.... I think the "Defaults to false" might need to be something like "If not specified then the service's 'updateable' field applies". Since we don't want this to change the current behavior if its not present.

spec.md
```diff
@@ -471,6 +471,7 @@ how Platforms might expose these values to their users.
 | metadata | object | An opaque object of metadata for a Service Plan. It is expected that Platforms will treat this as a blob. Note that there are [conventions](profile.md#service-metadata) in existing Service Brokers and Platforms for fields that aid in the display of catalog data. |
 | free | boolean | When false, Service Instances of this plan have a cost. The default is true. |
 | bindable | boolean | Specifies whether Service Instances of the Service Plan can be bound to applications. This field is OPTIONAL. If specified, this takes precedence over the `bindable` attribute of the service. If not specified, the default is derived from the service. |
+| updateable | boolean | Whether the Plan supports upgrade/downgrade/sidegrade to another version. Please note that the attribute is intentionally misspelled as `updateable` for legacy reasons. If not specificed the Service Offering's `updateable` field applies. |
```
Samze: We should add precdence behaviour if updateable is specified at the service level too. We did this for bindable and added the text:

`If specified, this takes precedence over the `bindable` attribute of the service. If not specified, the default is derived from the service.`

Samze:
What is the semantic relationship between the service level updatable field and plan level updatable field? Does a service updatable effect the plan updatable. For example is it valid to have service updatable false and plan updatable true?

duglin:
I'm leaning towards saying that the plan can be "false" if the service is "true", but the plan can not be "true" if the service is "false".  ie. a plan can be more restrictive than the service, but not less.

duglin:
I take that back... Russell just pointed out to me that "bindable" for services and plans allows for a plan to override the service regardless of the service's value.  So we should probably be consistent with that.

jberkhahn:
So leave it as is then? Independent of the service offering's field?

rboykin:
@jberkhahn @duglin I suggest adding in the default/precedence clauses like are done in the services.bindable and the plan.bindable fields. For services.bindable: 
```
This specifies the default for all plans of this service. Plans can override this field (see Plan Object).
```

For plan.bindable (changing `bindable` to `plan_updateable`): 
```
This field is OPTIONAL. If specified, this takes precedence over the `bindable` attribute of the service. If not specified, the default is derived from the service. 
```

jberkhahn:
borrowed the wording from bindable per Russel's suggestion

jberkhahn:
Fixed

Samze:
I wonder if having `plan_updateable` on the service offering and `updateable` on the service plan is confusing? 

Perhaps we should just use `plan_updateable` in both places for consistency?

tinygrasshopper:
There is some text in the `Updating a Service Instance` section which talks about enabling support for updating the service instance(`To enable support for the update of the plan, a Service Broker MUST declare support per service by including "plan_updateable": true in its catalog endpoint.`) I think that needs to be enhanced with the implications of the `updateable` flag on the plan.

mattmcneeney:
@jberkhahn to address comments and then we will re-review (hopefully asynchronously before next week).

duglin:
LGTM

spec.md
```diff
@@ -447,7 +447,7 @@ It is therefore RECOMMENDED that implementations avoid such strings.
...
 | dashboard_client | [DashboardClient](profile.md#dashboard-client-object) | A Cloud Foundry extension described in [Catalog Extensions](profile.md#catalog-extensions). Contains the data necessary to activate the Dashboard SSO feature for this service. |
-| plan_updateable | boolean | Whether the service supports upgrade/downgrade for some plans. Please note that the misspelling of the attribute `plan_updatable` as `plan_updateable` was done by mistake. We have opted to keep that misspelling instead of fixing it and thus breaking backward compatibility. Defaults to false. |
+| plan_updateable | boolean | Whether the Service Offering supports upgrade/downgrade for Service Plans by default. Service Plans can override this field (see Service Plan Object). Please note that the misspelling of the attribute `plan_updatable` as `plan_updateable` was done by mistake. We have opted to keep that misspelling instead of fixing it and thus breaking backward compatibility. Defaults to false. |
```
mattmcneeney: We should probably link to the Service Plan Object here (https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#plan-object)

duglin: ping @jberkhahn

jberkhahn:
Added the link.

spec.md
```diff
@@ -1059,8 +1060,8 @@ Service Instance to other plans. By modifying parameters, users can change
 configuration options that are specific to a service or plan.
 
 To enable support for the update of the plan, a Service Broker MUST declare
-support per service by including `"plan_updateable": true` in its [catalog
-endpoint](#catalog-management).
+support per service by including `"plan_updateable": true` in either the Service Offering
```
mattmcneeney: technically this should be `per Service Offering` but I'll let it slide :)

PR #596, opened 2018-09-03
Document `metadata.shareable` type as boolean

sslavic:
Before this change `metadata.shareable` was wrongly documented as
`string` typed attribute.

This change fixes the documented type of the attribute.


sslavic:
@duglin not to drag this one further, please do it in a new PR.

PR #607, opened 2018-09-25
Remove redundant text

### LINKED ISSUE #576: Remove "adding a service broker to the platform" section?
The text in [this section](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#adding-a-service-broker-to-the-platform) seems irrelevant and describes platform-specific behaviour:
> After implementing the first endpoint GET /v2/catalog documented above, the Service Broker will need to be registered with your Platform to make your services and plans available to end users.

Should we remove it?

mattmcneeney:
Resolves #576


duglin:
LGTM

Issue #603, opened 2018-09-14
Align our "valid chars in names/fields" with CF


duglin:
We're running into a situation where the spec says that service names "MUST only contain alphanumeric characters, periods, and hyphens (no spaces)", but we have existing service brokers that use underscores and CF allows this. While we could ask CF to change to match the spec I think its more important to support the existing brokers and change the spec to align with what CF does today.

We should:
1 - ask a CF person to provide the exact checks (if any) that it does for each of the fields - mainly the human readable fields, but technically we should check all of them if the spec tries to restrict it
2 - modify the spec to match

duglin:
Assign to Will

williammartin:
Hi @duglin, when you say:

> ask a CF person to provide the exact checks (if any) that it does for each of the fields - mainly the human readable fields, but technically we should check all of them if the spec tries to restrict it

I just want to get some clarification. You would like us to go through the specification, enumerate all the places the specification places restriction on naming, and cross reference these with any validation that CF does on these fields?

duglin:
@williammartin correct - then we can make the spec match what CF does

duglin:
@williammartin any update on this?

duglin:
were you going to open a PR?

williammartin:
Whoever gets to the story next will open a PR.

duglin:
@williammartin looking at https://www.pivotaltracker.com/story/show/160591651 - I see where it talks about what the spec says, but I don't see where it talks about what CF does. Or is that one of the comments and I'm just missing it?

williammartin:
Look for "CC, if present," italic comments

duglin:
Ah, thanks!

duglin:
@williammartin this can be closed now, right?

williammartin:
@henryaj I think you PRed some changes right?

henryaj:
Yes, https://github.com/openservicebrokerapi/servicebroker/pull/609 has been merged.

duglin:
great thanks- closing

PR #609, opened 2018-09-26
Harmonize field requirements in spec to those in CC

henryaj:
Currently the OSB API spec stipulates various requirements on the responses brokers should send back, e.g. a service name "MUST only contain alphanumeric characters, periods, and hyphens (no spaces)".

In practice, the CF Cloud Controller is more lenient than the spec (in the case of service names, non-alphanumeric characters are permitted). This PR brings the requirements in the spec into line with those enforced by CC.

[#160591651]

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -443,7 +443,7 @@ It is therefore RECOMMENDED that implementations avoid such strings.
 
 | Response Field | Type | Description |
 | --- | --- | --- |
-| name* | string | A CLI-friendly name of the service. MUST be unique across all service objects returned in this response. MUST be a non-empty string. |
+| name* | string | The name of the service. MUST be unique across all service objects returned in this response. MUST be a non-empty string. Using a CLI-friendly name is RECOMMENDED. |
 | id* | string | An identifier used to correlate this service in future requests to the Service Broker. This MUST be globally unique such that Platforms (and their users) MUST be able to assume that seeing the same value (no matter what Service Broker uses it) will always refer to this service. MUST be a non-empty string. Using a GUID is RECOMMENDED. |
 | description* | string | A short description of the service. MUST be a non-empty string. |
 | tags | array of strings | Tags provide a flexible mechanism to expose a classification, attribute, or base technology of a service, enabling equivalent services to be swapped out without changes to dependent logic in applications, buildpacks, or other services. E.g. mysql, relational, redis, key-value, caching, messaging, amqp. |
@@ -969,7 +969,7 @@ For success responses, the following fields are defined:
 | Response Field | Type | Description |
 | --- | --- | --- |
 | dashboard_url | string | The URL of a web-based management user interface for the Service Instance; we refer to this as a service dashboard. The URL MUST contain enough information for the dashboard to identify the resource being accessed (`9189kdfsk0vfnku` in the example below). Note: a Service Broker that wishes to return `dashboard_url` for a Service Instance MUST return it with the initial response to the provision request, even if the service is provisioned asynchronously. If present, MUST be a string or null. |
-| operation | string | For asynchronous responses, Service Brokers MAY return an identifier representing the operation. The value of this field MUST be provided by the Platform with requests to the [Polling Last Operation for Service Instances](#polling-last-operation-for-service-instances) endpoint in a percent-encoded query parameter. If present, MUST be a string or null. |
+| operation | string | For asynchronous responses, Service Brokers MAY return an identifier representing the operation. The value of this field MUST be provided by the Platform with requests to the [Polling Last Operation for Service Instances](#polling-last-operation-for-service-instances) endpoint in a percent-encoded query parameter. If present, MAY be null. |
 
 ```
 {
@@ -1172,7 +1172,7 @@ For success responses, the following fields are defined:
 
 | Response Field | Type | Description |
 | --- | --- | --- |
-| dashboard_url | string | The updated URL of a web-based management user interface for the Service Instance; we refer to this as a service dashboard. The URL MUST contain enough information for the dashboard to identify the resource being accessed (`0129d920a083838` in the example below). Note: a Service Broker that wishes to return `dashboard_url` for a Service Instance MUST return it with the initial response to the update request, even if the Service Instance is being updated asynchronously. If present, MUST be a string or null. If not present, the Platform MUST retain the previous value of the `dashboard_url`. |
+| dashboard_url | string | The updated URL of a web-based management user interface for the Service Instance; we refer to this as a service dashboard. The URL MUST contain enough information for the dashboard to identify the resource being accessed (`0129d920a083838` in the example below). Note: a Service Broker that wishes to return `dashboard_url` for a Service Instance MUST return it with the initial response to the update request, even if the Service Instance is being updated asynchronously. If not present or null, the Platform MUST retain the previous value of the `dashboard_url`. |
 | operation | string | For asynchronous responses, Service Brokers MAY return an identifier representing the operation. The value of this field MUST be provided by the Platform with requests to the [Polling Last Operation for Service Instances](#polling-last-operation-for-service-instances) endpoint in a percent-encoded query parameter. If present, MUST be a non-empty string. |
 
 ```
--- a/spec.md
+++ b/spec.md
@@ -1615,7 +1615,7 @@ For success responses, the following fields are defined:
 
 | Response Field | Type | Description |
 | --- | --- | --- |
-| operation | string | For asynchronous responses, Service Brokers MAY return an identifier representing the operation. The value of this field MUST be provided by the Platform with requests to the [Polling Last Operation for Service Instances](#polling-last-operation-for-service-instances) endpoint in a percent-encoded query parameter. If present, must not contain more than 10,000 characters. |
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -443,7 +443,7 @@ It is therefore RECOMMENDED that implementations avoid such strings.
 
 | Response Field | Type | Description |
 | --- | --- | --- |
-| name* | string | A CLI-friendly name of the service. MUST only contain alphanumeric characters, periods, and hyphens (no spaces). MUST be unique across all service objects returned in this response. MUST be a non-empty string. |
+| name* | string | A CLI-friendly name of the service. MUST be unique across all service objects returned in this response. MUST be a non-empty string. |
```
duglin: We have a bit of an inconsistency here because the definition of CLI-friendly say:
`... MUST only use alphanumeric characters, periods, and hyphens, with no spaces`
I'm assuming, based on the proposed change, that we should probably drop the CLI-friendly part of this. Or, perhaps say it SHOULD be a CLI-friendly name.

spec.md
```diff
@@ -775,7 +775,7 @@ For success responses, the following fields are defined:
 | Response Field | Type | Description |
 | --- | --- | --- |
 | state* | string | Valid values are `in progress`, `succeeded`, and `failed`. While `"state": "in progress"`, the Platform SHOULD continue polling. A response with `"state": "succeeded"` or `"state": "failed"` MUST cause the Platform to cease polling. |
-| description | string | A user-facing message that can be used to tell the user details about the status of the operation. If present, MUST be a non-empty string. |
+| description | string | A user-facing message that can be used to tell the user details about the status of the operation. |
```
duglin: Does this mean that CF allows for `null` and "" values?

spec.md
```diff
@@ -968,8 +968,8 @@ For success responses, the following fields are defined:
...
+| dashboard_url | string | The URL of a web-based management user interface for the Service Instance; we refer to this as a service dashboard. The URL MUST contain enough information for the dashboard to identify the resource being accessed (`9189kdfsk0vfnku` in the example below). Note: a Service Broker that wishes to return `dashboard_url` for a Service Instance MUST return it with the initial response to the provision request, even if the service is provisioned asynchronously. If present, MUST be a string or null. |
+| operation | string | For asynchronous responses, Service Brokers MAY return an identifier representing the operation. The value of this field MUST be provided by the Platform with requests to the [Polling Last Operation for Service Instances](#polling-last-operation-for-service-instances) endpoint in a percent-encoded query parameter. If present, MUST be a string or null. |
```
duglin: saying its a string is redundant with the definition of the property itself so we can drop that. I guess the real question is whether we want to be clear and say "This field MAY be null" - but I'm not sure what that means in the async case. What does CF do?

spec.md
```diff
@@ -1615,7 +1615,7 @@ For success responses, the following fields are defined:
 
 | Response Field | Type | Description |
 | --- | --- | --- |
-| operation | string | For asynchronous responses, Service Brokers MAY return an identifier representing the operation. The value of this field MUST be provided by the Platform with requests to the [Polling Last Operation for Service Instances](#polling-last-operation-for-service-instances) endpoint in a percent-encoded query parameter. If present, MUST be a non-empty string. |
+| operation | string | For asynchronous responses, Service Brokers MAY return an identifier representing the operation. The value of this field MUST be provided by the Platform with requests to the [Polling Last Operation for Service Instances](#polling-last-operation-for-service-instances) endpoint in a percent-encoded query parameter. If present, must not contain more than 10,000 characters. |
```
duglin: MUST NOT

spec.md
```diff
@@ -1547,7 +1547,7 @@ For success responses, the following fields are defined:
 
 | Response field | Type | Description |
 | --- | --- | --- |
-| operation | string | For asynchronous responses, Service Brokers MAY return an identifier representing the operation. The value of this field MUST be provided by the Platform with requests to the [Polling Last Operation for Service Bindings](#polling-last-operation-for-service-bindings) endpoint in a percent-encoded query parameter. If present, MUST be a non-empty string. |
+| operation | string | For asynchronous responses, Service Brokers MAY return an identifier representing the operation. The value of this field MUST be provided by the Platform with requests to the [Polling Last Operation for Service Bindings](#polling-last-operation-for-service-bindings) endpoint in a percent-encoded query parameter. If present, MUST be a string containing no more than 10,000 characters. |
```
duglin: let's remove the "MUST be a string" part since that's unnecessary

spec.md
```diff
@@ -1172,7 +1172,7 @@ For success responses, the following fields are defined:
...
+| dashboard_url | string | The updated URL of a web-based management user interface for the Service Instance; we refer to this as a service dashboard. The URL MUST contain enough information for the dashboard to identify the resource being accessed (`0129d920a083838` in the example below). Note: a Service Broker that wishes to return `dashboard_url` for a Service Instance MUST return it with the initial response to the update request, even if the Service Instance is being updated asynchronously. If present, MUST be a string or null. If not present, the Platform MUST retain the previous value of the `dashboard_url`. |
```
duglin: the last 2 sentences are awkward. Let's remove the part about it being a string since that's unnecessary. 
Does CF treat "null" the same as not present? Meaning the last sentence applies? If so, we should make it clear that null==not present.

spec.md
henryaj: Agreed - we'll drop the "CLI-friendly" from the first sentence.

spec.md
henryaj: That's correct.

spec.md
henryaj: Yes, CC treats 'null' the same as not present for this specific field. We'll update the text to clarify.

spec.md
```diff
@@ -472,7 +472,7 @@ how Platforms might expose these values to their users.
 | Response Field | Type | Description |
 | --- | --- | --- |
 | id* | string | An identifier used to correlate this plan in future requests to the Service Broker. This MUST be globally unique such that Platforms (and their users) MUST be able to assume that seeing the same value (no matter what Service Broker uses it) will always refer to this plan and for the same service. MUST be a non-empty string. Using a GUID is RECOMMENDED. |
-| name* | string | The CLI-friendly name of the plan. MUST only contain alphanumeric characters, periods, and hyphens (no spaces). MUST be unique within the service. MUST be a non-empty string. |
+| name* | string | The CLI-friendly name of the plan. MUST be unique within the service. MUST be a non-empty string. |
```
duglin: I like how you did it above..... "Using a CLI-friendly name is RECOMMENDED".  Removes the ambiguity around the first sentence.

spec.md
```diff
@@ -968,8 +968,8 @@ For success responses, the following fields are defined:
...
+| dashboard_url | string | The URL of a web-based management user interface for the Service Instance; we refer to this as a service dashboard. The URL MUST contain enough information for the dashboard to identify the resource being accessed (`9189kdfsk0vfnku` in the example below). Note: a Service Broker that wishes to return `dashboard_url` for a Service Instance MUST return it with the initial response to the provision request, even if the service is provisioned asynchronously. If present, MUST be a string or null. |
+| operation | string | For asynchronous responses, Service Brokers MAY return an identifier representing the operation. The value of this field MUST be provided by the Platform with requests to the [Polling Last Operation for Service Instances](#polling-last-operation-for-service-instances) endpoint in a percent-encoded query parameter. If present, MAY be null. |
```
duglin: When operation is null, does CF send any operation at all on the last_op call?

spec.md
duglin: can we use it here too?

spec.md
henryaj: I don't believe it does.

(I think [this](https://github.com/cloudfoundry/cloud_controller_ng/blob/master/lib/services/service_brokers/v2/client.rb#L243) is the relevant piece of code;  if the broker doesn't provide an operation (i.e. `parsed_response['operation']` is `nil`) then CC will return `nil` for `broker_provided_operation`.)

duglin: For bindings there's a 10k limit on the size of this field - but that limit isn't mentioned here. Is CF inconsistent on this or does the text here need to include that size limit as well?

henryaj: @duglin sorry for the delay - you're quite right, CC enforces a 10k char limit. I'll add it to the PR.

duglin:
LGTM

duglin:
ping @fmui @kibbles-n-bytes for a review

spec.md
```diff
@@ -443,7 +443,7 @@ It is therefore RECOMMENDED that implementations avoid such strings.
 
 | Response Field | Type | Description |
 | --- | --- | --- |
-| name* | string | A CLI-friendly name of the service. MUST only contain alphanumeric characters, periods, and hyphens (no spaces). MUST be unique across all service objects returned in this response. MUST be a non-empty string. |
+| name* | string | The name of the service. MUST be unique across all service objects returned in this response. MUST be a non-empty string. Using a CLI-friendly name is RECOMMENDED. |
```
fmui: Can we define here what a CLI-friendly name is? 
For example something like: "Using a CLI-friendly name, which only contains alphanumeric characters, periods, hyphens and no spaces is RECOMMENDED."

duglin: We define it here: https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#body
But it might not stand-out enough. Should we consider adding it to our "terminology" section?  Then we can just add an href to that section from here.

duglin: @fmui is this still an issue? a blocker?

zrob:
lgtm

henryaj:
Fixes https://github.com/openservicebrokerapi/servicebroker/pull/609.

henryaj:
Fixes https://github.com/openservicebrokerapi/servicebroker/issues/603.

PR #616, opened 2018-10-30
Adding the ability for the broker to publish maintenance and for the platform to trigger maintenance

tinygrasshopper:
There are cases in which the service provider would need to perform maintenance on the service instances that have been deployed by the users of the platform. We have been observing that for some services it is not possible to perform maintenance on the instance transparently(i.e. without effecting the workload the service is attached to). The intent of this PR is to give the platform the ability to understand when service instances need maintenance and for the platform user to trigger maintenance on instances that have been already provisioned. 

There are two problems being tackled with the PR: 
1. How does the broker inform the platform that maintenance is possible for instances 
1. How can the Platform trigger maintenance on service instance that need it

[Original proposal](https://docs.google.com/document/d/1ql0pNDGdxD1qlD1gYRytz7ZfGJT0XlO-Ol-hr-h7g-Y/edit#)(might be slightly out of date, but conveys intent)

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -477,6 +477,7 @@ how Platforms might expose these values to their users.
 | free | boolean | When false, Service Instances of this plan have a cost. The default is true. |
 | bindable | boolean | Specifies whether Service Instances of the Service Plan can be bound to applications. This field is OPTIONAL. If specified, this takes precedence over the `bindable` attribute of the service. If not specified, the default is derived from the service. |
 | plan_updateable | boolean | Whether the Plan supports upgrade/downgrade/sidegrade to another version. This field is OPTIONAL. If specificed, this takes precedence over the Service Offering's `plan_updateable` field. If not specified, the default is derived from the Service Offering. Please note that the attribute is intentionally misspelled as `plan_updateable` for legacy reasons. |
+| latest_instance_maintenance_information | object | An opaque object of the latest instance maintenance information for the plan. This field is OPTIONAL. It is expected that Platforms will treat this as a blob. Platforms MAY use changes in the latest maintenance information to derive if maintenance is possible on already provisioned Service Instances. Platforms MAY use the this to trigger maintenance on Service Instances using [Update Service Instance](#updating-a-service-instance). |
 | schemas | [Schemas](#schemas-object) | Schema definitions for Service Instances and Service Bindings for the plan. |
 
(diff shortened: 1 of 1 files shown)


duglin:
I checked with the IBM team and while this is an interesting feature, we don't have a need for this at this time.

mattmcneeney:
@duglin I was to dig into why this isn't a problem for IBM today. I imagine you must have scenarios whereby the software a service instance is using (anything from a library to the underlying OS) needs to be upgraded, and that will incur downtime for the service instance and any bound applications. Do you just transparently handle that with no impact or notification to the user? Do you use maintenance windows?

duglin:
I chatted with our product guys and got more info... lots of info... :-)
- today any maintenance that needs to happen is more of a global thing and not instance specific. So the idea of allowing an individual instance owner to choose when to kick it off isn't as critical to us right now
- having said that, conceptually as I said above, this is an interesting idea and might be useful in the future and used by people if made available
- however, while returning this information on the /v2/catalog response can work, there was a preference for allowing the broker to call back into the platform to signal this information. But, this is actually part of a bigger need. We're seeing cases where Brokers need to call back into the platform for a variety of reasons - for example, to see what instances the Platform thinks are there. So, orphan mitigation from the reverse side. We've seen cases where Brokers believe an instance exists and charges customers for it, but the platform didn't know about it. So, allowing the broker to do this reconciliation is important to us. We also talked about how having the call-back on the async ops would be useful too. So, we're now leaning more towards wanting to define a set of Platform APIs to allow for broker initiated interactions. But there is an acknowledgement that not all environments can support brokers calling back into platforms, so we may need to have two ways to solve these issues (platform initiated vs broker initiated).
- finally, does it make sense to overload the PATCH operation (which today means update my plan or parameters) with a new set of semantics (perform maintenance)? maybe? not sure. I guess if we allow for updating of the Instance "name" then this isn't much different.  Just another thought that came up in our talks.

kieron-pivotal:
We have been discussing the situation when a broker receives an update request containing maintenance information, but the maintenance information in the request does not match the current maintenance information in the catalog.

To be in the spirit of the spec, we suggest returning `422 Unprocessable Entity` with a body containing "MaintenanceInfoConflict".

cc @jacknewberry

mattmcneeney:
Closing in favour of @laurzca's new and improved PR: #628

PR #620, opened 2018-11-20
Introduce Platform metadata that can be updated by the platform

tinygrasshopper:
Alternative to the implementation proposed in #544 

Based on the hypothesis:
- Set of service instances that will consume information like name is less than service instances which will, so this PR tries to make updating of metadata optional 
- Platforms would have emergent requirements to update things other than name, like org/namespace when it is change for the service instance


duglin:
If we do this then I think we should do it in place of #544. We should also then define some of the properties that can appear in there - probably in the profile.md doc.

mattmcneeney:
Creating a new PR to cover this and #544

Issue #606, opened 2018-09-21
Proposal: Webhook callback when asynchronous create binding or provision completes


trevorlinton:
## Purpose

When making a provision or bind call to a OSB API that is asynchronous clients must then call the `last_operation` end point in a polling fashion to know when the request has succeeded or failed.  This creates a few un-necessary problems.  

The first is the client must retain state about what it was doing and what to do next which, should the client making the provision request not persist this information and state and crash, restart or encounter a redeploy the provisioned resource is orphaned.  Considering large windows for asynchronous provisioning can sometimes be up to 30 minutes this provides a large burden for clients to retain considerable state.

Second, if the broker service receives quite a large amount of client requests for provisions it may be  overwhelmed with polling operations to `last_operation` end points.   

This could be alleviated (or at least given alternate workflows for clients) by supporting the concept of an optional url callback where the client during a async provision or async create binding request provides a url where results of this operation should be sent.  When the operation (successful or not) completes the url is called with the information that would have been normally returned had the caller made the provision or create binding call synchronously (NOT the `last_operation` end point). 

## Rationale

Not all systems or brokers are truly platform aware, consider a broker which is designed to issue databases via REST interface, itself may not do the binding operations (but internally keep track of who is using the provisioned database), a client may be the calling platform to a broker to create a new resource, then capture those credentials during the binding phase and use them to support applications on its platform.   A broker may infact support multiple platforms (think an Azure/AWS/GCloud type system that may have a generic OSB database provisioner).

Since brokers in these scenarios (during binding or provision requests) have no real knowledge of the application they cannot reliably clean up an operation that may have been abandoned due to an intermediate failure by the client during the provision or binding window, they also rely on the client to continue to poll which can lead to wasteful operations or significant delays in provisioning should a clients poll interval be misconfigured and considerably too long. 

## Design

On `PUT /v2/service_instances/:instance_id` or `PUT /v2/service_instances/:instance_id/service_bindings/binding_id` if `accepts_incomplete=true` is passed, an optional `webhook` and `secret` may be provided.  The webhook parameter MUST be a valid URI-encoded `http` or `https` URI.  The `secret` MUST be provided if the `webhook` parameter is provided and MUST be text no more than 32 bytes used by the client to validate requests coming from the broker to the client. The secret SHOULD be unique to each provision or create binding call.

[...]

PR #621, opened 2018-11-24
Add a delay to polling response

duglin:
First part of #606

Adds a `poll_delay` field to the `last_operation` response to allow a
Broker to tell a Platform to back-off on its polling

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -281,6 +281,8 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/LastOperationResource'
+          headers:
+            - $ref: '#/components/parameters/RetryAfter'
         '400':
           description: Bad Request
           content:
@@ -336,6 +338,8 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/LastOperationResource'
+          headers:
+            - $ref: '#/components/parameters/RetryAfter'
         '400':
           description: Bad Request
           content:
@@ -530,6 +534,14 @@ components:
       name: X-Broker-API-Originating-Identity
       in: header
       description: identity of the user that initiated the request from the Platform
+      schema:
+        type: string
+
+    RetryAfter:
+      name: Retry-After
+      in: header
+      description: Indicates when to retry the request
+      require: false
       schema:
         type: string
 
--- a/spec.md
+++ b/spec.md
@@ -770,6 +770,7 @@ For success responses, the following fields are defined:
 | --- | --- | --- |
 | state* | string | Valid values are `in progress`, `succeeded`, and `failed`. While `"state": "in progress"`, the Platform SHOULD continue polling. A response with `"state": "succeeded"` or `"state": "failed"` MUST cause the Platform to cease polling. |
 | description | string | A user-facing message that can be used to tell the user details about the status of the operation. |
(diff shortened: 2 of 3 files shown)


tinygrasshopper:
I think we should leverage the HTTP caching mechanism, to optimize the polling. If the broker sets `Cache-Control: max-age=` header on the last operation resource, it would convey to the platform that this resource is not needed to be `GET`ted again for the specified time. That would make the platform implementation simpler as well by leveraging a HTTP client extension which can respect the cache control header.

mattmcneeney:
What are we going to do about this one @duglin ?

duglin:
On a previous call I believe it was @fmui who pointed out that the Retry-After header exists. The docs for that header (https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html) only talk about it being used for 503's and 3xx's, and we'd want to use it on a 200. So I did some digging and found this: https://tools.ietf.org/html/rfc6585  which talks about it being used on  a 429. This addresses something I was wondering.... can it be used for non 503's and 3xx's ?  The original spec doesn't say it can't - and RFC6585 proves that other people are interpreting it the same way.  :-)

I've modified this PR to use that header instead.  What do other people think?

mattmcneeney:
@duglin I think you may need to update `openapi.yaml` too

mattmcneeney:
I think I'm happy with this proposal, but I just wanted to highlight this text (from [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After)):
```
The Retry-After response HTTP header indicates how long the user agent should wait before making a follow-up request. There are three main cases this header is used:

- When sent with a 503 (Service Unavailable) response, this indicates how long the service is expected to be unavailable.
- When sent with a 429 (Too Many Requests) response, this indicates how long to wait before making a new request.
- When sent with a redirect response, such as 301 (Moved Permanently), this indicates the minimum time that the user agent is asked to wait before issuing the redirected request.
```

It doesn't say that we _can't_ use the header for this situation, but doesn't call out this case either.

duglin:
@mattmcneeney I updated `openapi.yaml` now - I also fixed the swagger file, not sure what I was thinking putting it on the request instead of the response!

mattmcneeney:
Trying to retrigger the Travis CI build now...

mattmcneeney:
Travis just passed, but doesn't seem to have updated this PR... https://travis-ci.org/openservicebrokerapi/servicebroker/builds/469807918

mattmcneeney:
Travis is working again at last! Phew!

georgi-lozev:
What is supposed to be the value? Milliseconds, seconds, minutes? 
I presume "an integer number of seconds (in decimal)" based on https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html, but should the spec have an opinion on that?

What should the platform do in case the broker provides date and not duration as per the recommendation? Error or ignore and stick to some default.

CC @laurzca @mattmcneeney @fmui

mattmcneeney:
My assumption was that Platforms would try to handle both a `HTTP-date` or `delta-seconds` as outlined in the spec. But we could be more restrictive in here (e.g. enforce a time interval in seconds) if we want to make it easier for Platforms?

PR #624, opened 2018-11-30
Dashboard client ID and secret are required in CF

laurzca:
We checked the behavior of `dashboard_client` in CF, and it turns out both the `dashboard_client.id` and `dashboard_client.secret` fields are required. This is technically a breaking change, but we want to bring the spec in line with actual behavior.


duglin:
LGTM

mattmcneeney:
Hopefully an easy one for @openservicebrokerapi/osbapi-pmc

fmui:
LGTM

PR #626, opened 2018-12-06
Add org and space name to CF context object

mattmcneeney:
This would be useful for Service Broker authors who want to, for example, display the org/space name where a Service Instance resides in a Cloud Foundry environment. This provides a better UX than using the org/space GUID, with the caveat that the name may be changed.

### DIFF
--- a/profile.md
+++ b/profile.md
@@ -174,6 +174,20 @@ deployment:
   ```
   "organization_guid": "1113aa0-124e-4af2-1526-6bfacf61b111"
   ```
+  - `organization_name`
+
+    Version: 2.15
+
+    The name of the organization that a Service Instance is associated with.
+    Note that the name of an organization in Cloud Foundry MAY be changed.
+    This property MUST be a non-empty string serialized as follows:
+    ```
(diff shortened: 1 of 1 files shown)


duglin:
LGTM

question: if an org, or space, name changes then the brokers will only know about when there's some OSBAPI specific action that needs to be taken, right? Meaning, CF will not send a message to the broker **just** because of the org/space name change - right?

duglin:
@kibbles-n-bytes @zrob @fmui @tinygrasshopper needs a review

mattmcneeney:
@duglin I've had to update this as I'd forgotten to add the field to the table of requests. Please can you re-review!

mattmcneeney:
Closed and reopened to trigger Travis to run again after I've fixed one of the test scripts which was failing

duglin:
LGTM

zrob:
lgtm

duglin:
@zrob can you use github's review mechanism - we're planning on moving away from pullApprove

mattmcneeney:
Closing this as it will be covered by #633

PR #628, opened 2018-12-18
Allow app developers to perform maintenance on a service instance

laurzca:
There are situations when a developer may want to upgrade a service instance, but that upgrade cannot be represented by either a plan change or through a configuration parameter. An example of this is upgrading the underlying operating system that a service instance is running on. This will often cause downtime for the service instance, so platform operators do not want to trigger this themselves or for it to happen automatically.

The intent of this change is to provide three new features:
1. Platforms can notify a developer when some maintenance can be performed on a service instance (through the `maintenance_info` field in the catalog)
2. This maintenance information can be provided in the request to update a service instance (alongside providing a plan or configuration parameters) in order to trigger maintenance on the service instance
3. This maintenance information can also be provided by a platform when provisioning a service instance to handle edge cases in which a service broker has been updated but the platform has not fetched the latest catalog, which will ensure that the end-user is provisioning what they are expecting (and the service broker can reject the request if otherwise)


jacknewberry:
Line 342 - error when maintenance_info does not match:
We also use "MaintenanceInfoConflict" in the case when the platform passes maintenance_info, but the broker catalog contains none. I think this use is consistent, but did trip us up.

Lines 1056 - update:
The ODB will probably not allow plan or parameters to be changed during maintenance, but I suspect you are correct that it is safer to leave individual brokers to define this behaviour.

spec.md
```diff
@@ -1032,27 +1043,32 @@ if it contains sensitive information.
 ## Updating a Service Instance
 
 By implementing this endpoint, Service Broker authors can enable users to
-modify two attributes of an existing Service Instance: the Service Plan and
-parameters. By changing the Service Plan, users can upgrade or downgrade their
-Service Instance to other Service Plans. By modifying parameters, users can change
-configuration options that are specific to a Service Offering or Service Plan.
+modify the following attributes of an existing Service Instance:
+* The Service Plan (allowing users to upgrade of downgrade their Service
```
AartiKriplani: typo here, should be `upgrade **or** downgrade`

mattmcneeney: Thank you @AartiKriplani !

mattmcneeney:
Added `maintenance_info` to `previous_values` for update operations too. This is for consistency with `plan_id` and `parameters`, the other two fields that can be changed via an update.

duglin:
How are these fields different from optional parameters that might be specified?

mattmcneeney:
The `maintenance_info` doesn't specify anything that can be _configured_, like a plan or parameters. Instead, it is additional information that the platform is saying it _expects_ the service instance it provisions to adhere to. Does that make sense?

duglin:
Both contain info about what kind of aspects we expect of the service.
In both cases we expect the broker to reject the request if those values are not supported.

The `os_version` example is an interesting one because I can very easily imagine that being viewed as a 'parameter' since the broker might actually allow people to choose from one than one value.

What if the platform sends just some of the fields? 

If this data is really 'opaque' to the platform, and we don't expect platforms or users to look at that data, then I would suggest we don't allow people to see it. Turn this field into a single value (int or uuid) and keep the actual meaning of that value hidden within the broker. The minute you expose the values people will want to see and understand them, and eventually control them - or offer people choices among them.... at that point they're pretty much "parameters".

mattmcneeney:
I think I understand where you're coming from. But what about things that can't be configured? In our world, things like the stemcell version, versions of libraries, etc, all cannot be configured but are potentially useful information for users to understand and trigger upgrades of.

In CF, we are actually planning on making this information visible to end users. Our current thinking is something like this:
```
$ cf services

name   service           plan              last operation
s1*    example-service   example-plan      create succeeded

* A newer version of the service instance is available.


$ cf service s1

name:            s1
service:         example-service
tags:
plan:            example-plan
description:     ...

Showing status of last operation from service s1...

status:    create succeeded
message:
started:   2019-01-02T14:10:47Z
updated:   2019-01-02T14:10:47Z

An upgrade is available for this service instance:
   stemcell: 224.6
   os: 12.4

To upgrade, run `cf upgrade-service s1`


$ cf upgrade-service s1

The service instance may be unavailable during this operation. Do you want to continue?
yes

Check cf service s1 to get the status of the upgrade.
```

duglin:
That helps me understand how you want to use it - thanks.  Couple of questions:
1 - is it an all-or-nothing type of thing? Meaning, do you ever envision people being able to specify just **some** of the properties?
2 - do you ever envision a broker wanting to allow users to choose from a collection of maintenance levels?  You're already offering them a choice between "current" and "new", it's not very hard for me to envision someone offering more than one (e.g. newer than current, but not the latest due to a new bug that's been found) ?  e.g. I'm on ubuntu 12.x, and I want to upgrade to 14.x but not 16.x yet because I haven't fully tested on 16.x but I know I don't want to stay on 12.04.
3 - doesn't exposing this much info to the user (stuff that you seem to be implying they have no real control over, aside from the big "upgrade or not" option) violate some basic tenant of OSBAPI? Isn't things like stemcell something that should be hidden from the user? Sure, a CF user might know what "stemcell" means, but I bet most Kube folks would not - nor should they.  And this goes to my previous questions.... you expose the "OS" in that data, which means people might now need to care - which might be a good or bad thing. But, the minute they have the option to think about it it's not long before they ask to have influence over it - thus it becomes a "parameter".

I'm still not convinced that this isn't a parameter, however, if we do treat it as something different I can't help but think of it more like the appStore. Meaning, tell people when a new version is available (along with some text saying what's changed), but then allow people to choose when to upgrade. IOW, just have a version string and some descriptive text. Encouraging brokers to expose the details via properties feels like we're headed down a slippery slope of turning them into another variant of parameters or exposing things that users should not really care about.

mattmcneeney:
1 - It's an all type of thing. We don't expect users (or platforms) to themselves choose what `maintenance_info` will look like. They will simply provide the same set of fields that were in the brokers catalog.
2 - We did some research with broker authors and don't think this will become a requirement. If something was designed to be chosen by an end user, it should be a parameter.
3 - I agree with this. Essentially we want to use this field to expose some 'real life' information through the black box that is OSBAPI. One of the biggest pain points we hear is that users don't totally understand what they're getting. They don't want to configure everything (and broker authors don't want to make everything configurable), but we want a way of exposing some of the things that are actually happening behind the scenes.

I see this very similar to having some kind of upgrade string, but a more detailed object may allow platforms to provide a better UX, and could make it easier for platforms to compare old and new to tell when an upgrade is available (how many times do you just see _"Bug fixes"_ in the app store; how could we ensure broker authors don't write that and then inadvertently make it seem like no upgrade is available)

duglin:
A version string would solve that.

mattmcneeney:
But it would be more difficult for service authors to detail complex information about what is now available. From the example above:
```
An upgrade is available for this service instance:
   stemcell: 224.6
   os: 12.4
```

We could obviously have a version string like `"stemcell: 224.6, os: 12.4"`, but I don't think that provides any benefit and makes it harder for UIs to use the information?

duglin:
Sorry - wasn't clear, what I meant was to make it:
```
      "maintenance_info": {
        "version": "1.1",
        "details": "random text which could include info like: OS Version: 2.1"
      }
```

I'm trying to avoid the specifics of what's new from being machine readable because I think if we do that people will keep asking for more things. Such as: 
- how do I know what each property mean?
- what are the possible properties that might appear?
- what are the possible values of each property?
- why can't I offer my users a choice for each property? e.g. they can choose their OS
- why can't I offer my users more than one maintenance level? They may support more than one variant at a time

All of these push these fields towards becoming parameters - which is what I think they should be in the first place (just optional) - and I'd like to avoid two types of parameters.

gberche-orange:
> This maintenance information can also be provided by a platform when provisioning a service instance to handle edge cases in which a service broker has been updated but the platform has not fetched the latest catalog, which will ensure that the end-user is provisioning what they are expecting (and the service broker can reject the request if otherwise)

I believe the problem of "stale catalog" cached by platform is wider than security patches to service instances that this PR seems to target. When a platform exposes a "stale catalog version" to its users, violated user expectations will be wider:
- stale catalog condition also need to be detected in other phases, such as service instance provisioning, service binding
- stale catalog items can be wider than stemcell/os version and include other items such as pricing changes, documentation changes, metadata changes, behavior changes

I would therefore suggest to look for a more generic behavior for handling state catalog conditions such as:
- mandate use of CacheControl headers by platforms to the `v2/catalog` endpoint to ensure platform poll brokers regularly enough to reduce stale catalog.
- generalize the `maintenance_info` concept to enable broker to refuse other stale catalog conditions that CVEs. 
   - May be rename `maintenance_info` into `catalog_update`

``` 
      "catalog_update": {
        "version": "1.1",
        "details": "Stemcell/os version X. "
        "implies_downtime": true
      }
``` 

or

``` 
      "catalog_update": {
        "version": "1.1",
        "details": "Pricing v2. SLA v3"
        "implies_downtime": false
      }
```
   - Extend to other steps than update service instance.

tinygrasshopper:
In my mind, this proposal is trying to achieve:
- Enable brokers to signal to the platform that an update is available
- Enable platforms to trigger maintenance updates on the service instances

With the constraints:
- Services can have multiple aspects about them that can be changed, and might need to be communicated to the user. 
- Service brokers can remain "stateless" around which service instances need maintenance, the platform remains the source of truth of this information. 

The UI that the current changes to the spec would allow are:
```
$ cf services

name   service           plan              last operation
s1*    example-service   example-plan      create succeeded

* A newer version of the service instance is available.


$ cf service s1

name:            s1
service:         example-service
tags:
plan:            example-plan
description:     ...

Showing status of last operation from service s1...

status:    create succeeded
message:
started:   2019-01-02T14:10:47Z
updated:   2019-01-02T14:10:47Z

An update is available for this service instance
To upgrade, run `cf update-service s1`


$ cf update-service s1

The service instance may be unavailable during this operation. Do you want to continue?
yes

Check cf service s1 to get the status of the update. 
```

@duglin:
> ```
>       "maintenance_info": {
>         "version": "1.1",
>         "details": "random text which could include info like: OS Version: 2.1"
>       }
> ```

The motivation for the maintenance info is to communicate the changes in aspects of the service to the platform, i.e. combination of all things that make up a SI. The reason that I would like this to be a free form object is, the broker is best placed to control how many versions/properties make up a service instance. 

You are right in saying that with the information currently provided, the platform has no way of interpreting what the keys in the maintenance information mean. The platform can detect which keys have changed i.e. which aspects of the service instance have changed, but that might not be meaningful to the user. I like the suggestion of the `details` field, that allows the broker to craft messages that it thinks users might be interested in, but I think that not be in the `maintenance_info` block, rather a top level field. The intent of the maintenance_info is to communicate the latest maintenance information, rather than user facing messages, as the `maintenance_info` can change without the user facing messages changing. 

This PR allows for further extensions that can be added to the spec that describe how information in `maintenance_info` can be interpreted to build better UIs.

spec.md
```diff
@@ -916,6 +922,7 @@ Service Broker will use it to correlate the resource it creates.
...
 | parameters | object | Configuration parameters for the Service Instance. Service Brokers SHOULD ensure that the client has provided valid configuration parameters and values for the operation. |
+| maintenance_info | object | If a Service Broker provides maintenance information for a Service Plan in its [Catalog](#catalog-management), a Platform MAY provide the same maintenance information when provisioning a Service Instance. This field can be used to ensure that the end-user of a Platform are provisioning what they are expecting since maintenance information can be used to describe important information (such as the version of the operating system the Service Instance will run on). If a Service Broker's catalog has changed and new maintenance information is available for the Service Plan being provisioned, then the Service Broker MUST reject the request with a `422 Unprocessable Entity` as described below. |
```
fmui: This implies that the end-user has seen and understood the maintenance info. Without semantics, the platform can only present the plain JSON object. I guess many end-users don't understand what they are looking at.
And how should an end-user judge whether the OS version is good or a bad for a service instance? The OS a database is running on, for example, is an implementation detail and should be hidden from an end-user.

jacknewberry: I don't think this implies that the end user has always completely understood the maintenance info. Service authors can select the info that they think might be useful to end users, but not every end user needs to understand the whole object. 
At the lowest level, a basic user needs to know that there is maintenance available. More experienced users might be interested in the details of the maintenance.

Showing maintenance information allows users to work towards better understanding of the services they use, and allows already accomplished users to make good decisions about applying maintenance operations. Hiding information about maintenance disempowers both new and experienced users.

The platform can create a range of possible user experiences when displaying maintenance info. Examples could include: displaying the whole JSON as you suggest; displaying an itemised diff; displaying only a notification that the maintenance info has changed, and perhaps others. If this does not provide enough flexibility for platforms, the spec could be made more restrictive in the future.

fmui: Most services provided via OSB are managed services. There is an inherent disempowerment. The only choice the user has is when to apply the update and not if the update should be deployed at all. If the update fixes a security issue, the service instance will be updated anyway automatically at some point.
Eventually, this boils down to a hint that there is an update available with some high-level description. Providing service internals is a bad design, is unnecessary, and in the worst case can be a security issue.

jacknewberry: I agree there is an inherent disempowerment with managed services. On the one hand they free users of the burden of maintaining their service, but on the other there is a risk that users are not able to use the service effectively.

Providing some information about the internals of a service need not represent bad design. Often some information is useful to make decisions about how a service is used, or to help make judgements about how it will respond. 

Importantly, providing a reasonable amount of information to the end user will not harm them.  There are times when it will provide a real benefit.

mattmcneeney:
I believe this can now be closed since we have a replacement with #642

PR #632, opened 2019-01-11
Add binding endpoints

fmui:
I'm splitting the network metadata PR (#586) into multiple pieces to simplify the discussion.
This is part 1 - binding endpoint.

This PR defines how brokers can provide the network endpoints of a service instance to the platform. The endpoint data is sent along with the binding credentials and can be used by the platform (or other intermediate components) to configure network components.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1182,7 +1182,8 @@ For success responses, the following fields are defined:
 If `"bindable": true` is declared for a Service Offering or Service Plan in the
 [Catalog](#catalog-management) endpoint, the Platform MAY request generation
 of a Service Binding. Otherwise, Platforms MUST NOT make a binding request to
-the Service Broker for any Service Instance using the given Service Offering or Service Plan.
+the Service Broker for any Service Instance using the given Service Offering
+or Service Plan.
 
 Note: Not all services need to be bindable --- some deliver value just from
 being provisioned. Service Brokers that offer services that are bindable MUST
@@ -1205,6 +1206,10 @@ Credentials are a set of information used by an Application or a user to
 utilize the Service Instance. Credentials SHOULD be unique whenever possible, so
 access can be revoked for each Service Binding without affecting consumers of other
 Service Bindings for the Service Instance.
+
+Service Brokers SHOULD also provide all network hosts and ports that the
+Application uses to connect to the Service Instance via this Service Binding.
+This data allows the Platform to adjust network configurations, if necessary. 
 
 #### Log Drain
 
@@ -1364,6 +1369,7 @@ For `200 OK` and `201 Created` response codes, the following fields are defined:
 | syslog_drain_url | string | A URL to which logs MUST be streamed. `"requires":["syslog_drain"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
 | route_service_url | string | A URL to which the Platform MUST proxy requests for the address sent with `bind_resource.route` in the request body. `"requires":["route_forwarding"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
 | volume_mounts | array of [VolumeMount](#volume-mount-object) objects | An array of configuration for remote storage devices to be mounted into an application container filesystem. `"requires":["volume_mount"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
+| endpoints | array of [Endpoint](#endpoint-object) objects | The network endpoints that the Application uses to connect to the Service Instance. The endpoints need not to be reachable and the host names need not to resolvable from outside the service network. |
 
 ##### Volume Mount Object
 
@@ -1382,10 +1388,20 @@ For `200 OK` and `201 Created` response codes, the following fields are defined:
 Currently only shared devices are supported; a distributed file system which
 can be mounted on all app instances simultaneously.
 
-| Field | Type | Description |
+| Response Field | Type | Description |
 | --- | --- | --- |
 | volume_id* | string | ID of the shared volume to mount on every app instance. |
 | mount_config | object | Configuration object to be passed to the driver when the volume is mounted. |
+
+\* Fields with an asterisk are REQUIRED.
+
+##### Endpoint Object
+
+| Response Field | Type | Description |
+| --- | --- | --- |
+| host* | string | A host name or a single IP address. |
+| ports* | array of strings | A non-empty array. Each element is either a single port (for example "443") or a port range (for example "9000-9010"). |
+| protocol | string | The protocol. Valid values are `tcp`, `udp`, or `all`. The default value is `tcp`. |
 
 \* Fields with an asterisk are REQUIRED.
 
@@ -1398,7 +1414,13 @@ can be mounted on all app instances simultaneously.
     "host": "mysqlhost",
     "port": 3306,
     "database": "dbname"
-  }
+  },
+  "endpoints": [
+    {
+      "host": "mysqlhost",
+      "ports:" ["3306"]
+    }
+  ]
 }
 ```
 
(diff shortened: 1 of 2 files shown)


spec.md
```diff
@@ -1364,6 +1369,7 @@ For `200 OK` and `201 Created` response codes, the following fields are defined:
 | syslog_drain_url | string | A URL to which logs MUST be streamed. `"requires":["syslog_drain"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
 | route_service_url | string | A URL to which the Platform MUST proxy requests for the address sent with `bind_resource.route` in the request body. `"requires":["route_forwarding"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
 | volume_mounts | array of [VolumeMount](#volume-mount-object) objects | An array of configuration for remote storage devices to be mounted into an application container filesystem. `"requires":["volume_mount"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
+| endpoints | array of [Endpoint](#endpoint-object) objects | The network endpoints that the Application uses to connect to the Service Instance. The endpoints need not to be reachable and the host names need not to resolvable from outside the service network. |
```
mattmcneeney: "need not to be" reads a little funny to me, but I'm not sure if I can think of a better phrase.

Also, s/the host names need not to resolvable/the host names need not be resolvable/

shalako: "Endpoints and hostnames will not necessarily be accessible or resolvable from outside the service network."

shalako: nvm, I see wording was updated

shalako: There are a couple of minor typos in the new wording. Suggested fixes in bold below.

The network endpoints that the Application uses to connect to the Service Instance. All Service Instance endpoints that **are** relevant for the Application MUST be in this list, even if the endpoints are not reachable or host names are not resolvable **from** outside the service network.

leonwanghui:
I think the `requires` (shown below) should also be updated if we want to add `endpoints` as binding option.

mattmcneeney:
Hey @leonwanghui 

The `requires` field is used to specify what parameters a Platform MUST include when making a binding request to a Service Broker. For example, if a Service declares `"requires": ["route_forwarding"]` in the catalog, then the binding request MUST include `"bind_resource": {"route":"..."}}`.

In this case, we're only describing what the binding _response_ may contain.

Does that make sense?

leonwanghui:
Hi @mattmcneeney , I think I got your point, but what I mean is not set `endpoints` as the required parameter, rather it means the only permissions currently supported are `syslog_drain`, `route_forwarding`, `volume_mount` and `endpoints`. Just my thought : )

spec.md
```diff
@@ -1464,6 +1486,7 @@ For success responses, the following fields are defined:
 | route_service_url | string | A URL to which the Platform MUST proxy requests for the address sent with `bind_resource.route` in the request body. `"requires":["route_forwarding"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
 | volume_mounts | array of [VolumeMount](#volume-mount-object) objects | An array of configuration for mounting volumes. `"requires":["volume_mount"]` MUST be declared in the [Catalog](#catalog-management) endpoint or the Platform can consider the response invalid. |
 | parameters | object | Configuration parameters for the Service Binding. |
+| endpoints | array of [Endpoint](#endpoint-object) objects | The network endpoints that the Application uses to connect to the Service Instance. All Service Instance endpoints that are relevant for the Application MUST be in this list, even if endpoints are not reachable or host names are not resolvable from outside the service network. |
```
mattmcneeney: It's possible to misinterpret this MUST. @fmui to fix this.

shalako:
This appears to standardize on the `endpoints` object to indicate the host, port, and protocol by which a consumer can reach the hosts of a marketplace service instances. Seems reasonable; `credentials` was generic to begin with but our expectation was that common fields would be identified and made more explicit.

tinygrasshopper:
Here is a thought, does the adding of the `endpoints` top level field overlap with the generic problem that we have in the spec around communication/interpretation of meaning about the things in the credentials block. The `endpoints` block proposed is extraction of the information that the broker will send thru in the binding block

One of the ideas we had floating around for solving the generic problem of lack of structure in the credentials block was with the declaration of a "kind" for the binding credentials returned, so that platforms can interpret and potentially take action on the credentials block, I am not sure if this proposal is a special case of a "kind".

So the way it might look like as a kind is:

```json
{
  "credentials": {
    "uri": "mysql://mysqluser:pass@mysqlhost:3306/dbname",
    "username": "mysqluser",
    "password": "pass",
    "host": "mysqlhost",
    "port": 3306,
    "database": "dbname",
  },
  "kind": ["single-host", "mysql"]
}
```

or
```json
{
  "credentials": {
    "uri": "mysql://mysqluser:pass@mysqlhost1:3306,mysqlhost1:3305/dbname",
    "username": "mysqluser",
    "password": "pass",
    "host": "mysqlhost",
    "port": 3306,
    "database": "dbname",
    "hosts": [
      {
        "host": "mysqlhost1",
        "ports": ["3306"]
      },
      {
        "host": "mysqlhost2",
        "ports": ["3305"]
      }
     ],
  },
  "kind": ["multi-host", "mysql"]
}
```

Where kind definitions can look like:

```
Kind mysql: Must have a uri

Kind single-host: Must have host and port in the credentials block 

Kind multi-host: Must have a hosts object
```

The networking extension mechanism in my mind, can kick in in reaction to the kinds `single-host` or `multi-host` being sent in the response.

I think there are downsides thinking about these things as kinds rather than a new endpoint:
- there might be a name conflict with existing things in the credentials block for the broker
- I am not sure if the kinds proposal can fit the use-case of the next part of the overall networking PR, which is the hosts mapping

The upside is, its part of the larger extensibility mechanism around bindings. 

Thoughts? 

cc: @Samze

tinygrasshopper:
mermaid-diagram-20190219132828

me and @fmui chatted about this PR offline. We used the above example to work through how this information would be used in the future. 

The main constraints on using the 'kinds' approach:
1. Each of the proxies would have to be aware of kinds. We want to reduce maintenance overhead for proxies 
1. Existing creds blocks would have to change in case of conflict with kind definitions

loewenstein:
@tinygrasshopper you can find endpoints (, adapt_credentials and network_profile/network_data) being used in https://github.com/Peripli/istio-broker-proxy/

Diagram in readme is quite similar to the above.

spec.md
```diff
@@ -1382,13 +1388,23 @@ For `200 OK` and `201 Created` response codes, the following fields are defined:
...
 
 \* Fields with an asterisk are REQUIRED.
 
+##### Endpoint Object
+
+| Response Field | Type | Description |
+| --- | --- | --- |
+| host* | string | A host name or a single IP address. |
+| ports* | array of strings | A non-empty array. Each element is either a single port (for example "443") or a port range (for example "9000-9010"). |
+| protocol | string | The protocol. Valid values are `tcp`, `udp`, or `all`. The default value is `tcp`. |
```
loewenstein: Would it make sense to allow L7 protocols like HTTP in this place? An implementation like the https://github.com/Peripli/istio-broker-proxy could decide on specific Istio rules based on that protocol information. The alternative would be to configure this out-of band and put it into Network Profile Data.

fmui: I would rather keep this PR as simple as it is now and move additional topics to another PR.

spec.md
```diff
@@ -1198,6 +1198,14 @@ Service Binding](#fetching-a-service-binding) endpoint. However, in order for
 the Platform to successfully use the Service Binding, the information MUST be
 returned at least once.
 
+In some cases, the creation of a Service Binding has to be accompanied by
+additional network configuration. This could be the configuration of a firewall
+or a load balancer or it could be the setup of a network tunnel or a VPN
+connection.
+To enable the Platform or another central component to do this in a service
```
tinygrasshopper: I think `another central component` is confusing here, as it introduces a term which is not mentioned in the spec before. I think it would read better as `To enable the Platform to do this in a service agnostic way`

mattmcneeney:
We did it!

PR #633, opened 2019-01-14
Allow context to be updated and add org name, space name and instance name

mattmcneeney:
PRs #620, #544 and #626 were all dealing with the same problem:
**It is useful for Service Brokers to understand Platform-side information for Service Instances**.

This was proving difficult, especially for the name of the service instance, and the org/space names in CF, because Platforms may allow this information to be changed at any time.

#620 considered adding a new `platform_metadata` object, but this looked very similar to the `context` object that already exists.
#544 considered adding a top-level `name` field to a Service Instance.
#626 is now included as part of this PR, and added `organization_name` and `space_name` to `context`, but with no way for Platforms to signal to Service Brokers if any of those values change

This PR therefore allows for the following:
* Service Broker authors can _opt-in_ to receive update service instance requests when contextual information in a Platform changes (i.e. the name of the service instance) by specifying `"allow_context_updates": true` in its catalog
* Platforms can send a `PATCH /v2/service_instances/:guid` request to Service Brokers that opt-in to this, sending updated Platform-side contextual information
* The `context` object for Cloud Foundry now includes `instance_name`, `organization_name` and `space_name`

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -450,6 +450,7 @@ It is therefore RECOMMENDED that implementations avoid such strings.
 | bindable* | boolean | Specifies whether Service Instances of the service can be bound to applications. This specifies the default for all Service Plans of this Service Offering. Service Plans can override this field (see [Service Plan Object](#service-plan-object)). |
 | instances_retrievable | boolean | Specifies whether the [Fetching a Service Instance](#fetching-a-service-instance) endpoint is supported for all Service Plans. |
 | bindings_retrievable | boolean | Specifies whether the [Fetching a Service Binding](#fetching-a-service-binding) endpoint is supported for all Service Plans. |
+| allow_context_updates | boolean | Specifies whether a Service Instance supports [Update](#updating-a-service-instance) requests when contextual data for the Service Instance in the Platform changes. |
 | metadata | object | An opaque object of metadata for a Service Offering. It is expected that Platforms will treat this as a blob. Note that there are [conventions](profile.md#service-metadata) in existing Service Brokers and Platforms for fields that aid in the display of catalog data. |
 | dashboard_client | [DashboardClient](profile.md#dashboard-client-object) | A Cloud Foundry extension described in [Catalog Extensions](profile.md#catalog-extensions). Contains the data necessary to activate the Dashboard SSO feature for this service. |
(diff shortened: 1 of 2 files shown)


spec.md
```diff
@@ -450,6 +450,7 @@ It is therefore RECOMMENDED that implementations avoid such strings.
 | bindable* | boolean | Specifies whether Service Instances of the service can be bound to applications. This specifies the default for all Service Plans of this Service Offering. Service Plans can override this field (see [Service Plan Object](#service-plan-object)). |
 | instances_retrievable | boolean | Specifies whether the [Fetching a Service Instance](#fetching-a-service-instance) endpoint is supported for all Service Plans. |
 | bindings_retrievable | boolean | Specifies whether the [Fetching a Service Binding](#fetching-a-service-binding) endpoint is supported for all Service Plans. |
+| allow_context_updates | boolean | Specifies whether a Service Instance supports [Update](#updating-a-service-instance) requests when contextual data for the Service Instance in the Platform changes. |
```
tinygrasshopper: `allow_context_updates` -> `context_updateable` to be consistent with existing flags like `bindings_retrievable` ?

mattmcneeney:
But that sounds like context is a resource, like an instance or a binding,
which it isn’t!

jberkhahn:
On the Kubernetes side, I believe we use "namespace" for the namespace name field, so we might want to use "organization" and "space" here, or change kube to use "namespace_name".

spec.md
fmui: Can the context object also change during a normal instance update?

mattmcneeney: Yes, I think you can infer that from the spec today.

The aim with this flag is to allow service brokers to specify whether they would like to receive update requests when *only* context data changes.

mattmcneeney:
@jberkhahn That does make sense, but in my opinion `organization_guid` and `organization_name` looks more consistent than `organization_guid` and `organization`. Hmm.

AartiKriplani:
Currently, we(CF) also send the context info (incl platform, org guid, space guid) on `bind` action too. Does it make sense to add  `instance_name`, `organization_name` and `space_name` too? For consistency purpose? @mattmcneeney

PR #636, opened 2019-02-05
Clarify catalog schema requirements

jberkhahn:
Per the discussion today, make these requirements more explicit. It would be great if @Samze or @mattmcneeney could verify if this is the extant behavior in CF

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -516,6 +516,8 @@ The following rules apply if `parameters` is included anywhere in the catalog:
 * Platforms SHOULD be prepared to support later versions of JSON schema.
 * The `$schema` key MUST be present in the schema declaring the version of JSON
 schema being used.
+* The `type` key MUST be present in the schema declaring the type as `object`.
+* The `parameters` key MUST be present in the schema, but MAY be empty if the action accepts no parameters.
 * Schemas MUST NOT contain any external references.
 * Schemas MUST NOT be larger than 64kB.
 


mattmcneeney:
@jberkhahn 

```
* The `type` key MUST be present in the schema declaring the type as `object`.
```
In CF, we do not require `type` to be present in the schema. If it is present, it must be a valid type, but it doesn't have to be an object.

```
* The `parameters` key MUST be present in the schema, but MAY be empty if the action accepts no parameters.
```
In CF, we do not require `parameters` to be present in the schema. If it is present, it may be empty.

Samze:
I disagree that we need more strict validation of the `type` or `properties` keywords. 

I think the only reason we enforced `$schema` at all is so that the platform knows whether they support parsing that particular version of JSON schema. Adding further restrictions beyond being valid JSON schema seems pointless to me. 

For broker authors who want zero properties they would have to understand the restrictions we've imposed and define:

```
{
   "type": "object",
   "properties": {},
   "additionalProperties": false
}
```

Rather than the more simple:
```
{
  "additionalProperties": false
}
```

So we'd be expecting Broker authors to understand both JSON schema and our increased set of rules. Is there a benefit here I'm missing?

jberkhahn:
how is a blank parameters to be interpreted then? Is it allow-anything?

PR #637, opened 2019-02-18
Allow Service Brokers to indicate the state of a Service Instance after a failed update or deprovisioning

fmui:
When an instance update or delete fails, the platform doesn't know the state of the instance and doesn't know if the operation can be repeated.
This PR allows brokers to provide this additional information to the platform in case of a failure.

This PR replaces PR #570.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -314,6 +314,9 @@ For error responses, the following fields are defined:
 | --- | --- | --- |
 | error | string | A single word in camel case that uniquely identifies the error condition. If present, MUST be a non-empty string. |
 | description | string | A user-facing error message explaining why the request failed. If present, MUST be a non-empty string. |
+| instance_usable | boolean | If an update or deprovisioning operation failed, this flag indicates whether or not the Service Instance is still usable. If `true`, the Service Instance can still be used, `false` otherwise. This field MUST NOT be present for errors of other operations. Defaults to true. |
+| update_repeatable | boolean | If an update operation failed, this flag indicates whether this update can be repeated or not. If `true`, the same update operation can be repeated and may succeed; if `false`, repeating the same update operation will fail again. This field MUST NOT be present for errors of other operations. Defaults to true. |
+| retry_delay | integer | This field suggests how long (in seconds) the Platform SHOULD wait until it repeats the operation. If this a negative number, the Platform SHOULD NOT automatically repeat the operation. Defaults to 0 seconds. |
 
 ### Error Codes
 
@@ -782,6 +785,9 @@ For success responses, the following fields are defined:
 | --- | --- | --- |
 | state* | string | Valid values are `in progress`, `succeeded`, and `failed`. While `"state": "in progress"`, the Platform SHOULD continue polling. A response with `"state": "succeeded"` or `"state": "failed"` MUST cause the Platform to cease polling. |
 | description | string | A user-facing message that can be used to tell the user details about the status of the operation. If present, MUST be a non-empty string. |
+| instance_usable | boolean | If an update or deprovisioning operation failed, this flag indicates whether or not the Service Instance is still usable. If `true`, the Service Instance can still be used, `false` otherwise. This field MUST NOT be present for errors of other operations. Defaults to true. |
+| update_repeatable | boolean | If an update operation failed, this flag indicates whether this update can be repeated or not. If `true`, the same update operation can be repeated and may succeed; if `false`, repeating the same update operation will fail again. This field MUST NOT be present for errors of other operations. Defaults to true. |
+| retry_delay | integer | If an operation failed, this field suggests how long (in seconds) the Platform SHOULD wait until it repeats the operation. If this a negative number, the Platform SHOULD NOT automatically repeat the operation. Defaults to 0 seconds. |
 
 \* Fields with an asterisk are REQUIRED.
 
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -328,6 +328,9 @@ For error responses, the following fields are defined:
 | --- | --- | --- |
 | error | string | A single word in camel case that uniquely identifies the error condition. If present, MUST be a non-empty string. |
 | description | string | A user-facing error message explaining why the request failed. If present, MUST be a non-empty string. |
+| instance_usable | boolean | If an update or deprovisioning operation failed, this flag indicates whether or not the Service Instance is still usable. If `true`, the Service Instance can still be used, `false` otherwise. This field MUST NOT be present for errors of other operations. Defaults to true. |
+| update_repeatable | boolean | If an update operation failed, this flag indicates whether this update can be repeated or not. If `true`, the same update operation MAY be repeated and MAY succeed; if `false`, repeating the same update operation will fail again. This field MUST NOT be present for errors of other operations. Defaults to true. |
+| retry_delay | integer | This field suggests how long (in seconds) the Platform SHOULD wait until it repeats the operation. If this a negative number, the Platform SHOULD NOT automatically repeat the operation. Defaults to 0 seconds. |
```
waterlink: We already have a `retry after` concept in the `last operation` endpoint. Do you think it makes sense to use the same name for this as well, or do you think it needs to be distinct?

spec.md
```diff
@@ -772,15 +775,21 @@ For success responses, the following fields are defined:
 | Response Field | Type | Description |
 | --- | --- | --- |
 | state* | string | Valid values are `in progress`, `succeeded`, and `failed`. While `"state": "in progress"`, the Platform SHOULD continue polling. A response with `"state": "succeeded"` or `"state": "failed"` MUST cause the Platform to cease polling. |
+| description | string | A user-facing message that can be used to tell the user details about the status of the operation. If present, MUST be a non-empty string. |
```
waterlink: The description field appears twice in this section. Conflict resolution artifact?

spec.md
```diff
@@ -772,15 +775,21 @@ For success responses, the following fields are defined:
...
+| instance_usable | boolean | If an update or deprovisioning operation failed, this flag indicates whether or not the Service Instance is still usable. If `true`, the Service Instance can still be used, `false` otherwise. This field MUST NOT be present for errors of other operations. Defaults to true. |
+| update_repeatable | boolean | If an update operation failed, this flag indicates whether this update can be repeated or not. If `true`, the same update operation MAY be repeated and MAY succeed; if `false`, repeating the same update operation will fail again. This field MUST NOT be present for errors of other operations. Defaults to true. |
+| retry_delay | integer | If an operation failed, this field suggests how long (in seconds) the Platform SHOULD wait until it repeats the operation. If this a negative number, the Platform SHOULD NOT automatically repeat the operation. Defaults to 0 seconds. |
 | description | string | A user-facing message that can be used to tell the user details about the status of the operation. |
```
waterlink: Duplicate field here.

spec.md
fmui: Thanks. Fixed.

waterlink: Great! Thank you!

spec.md
```diff
@@ -328,6 +328,9 @@ For error responses, the following fields are defined:
 | --- | --- | --- |
 | error | string | A single word in camel case that uniquely identifies the error condition. If present, MUST be a non-empty string. |
 | description | string | A user-facing error message explaining why the request failed. If present, MUST be a non-empty string. |
+| instance_usable | boolean | If an update or deprovisioning operation failed, this flag indicates whether or not the Service Instance is still usable. If `true`, the Service Instance can still be used, `false` otherwise. This field MUST NOT be present for errors of other operations. Defaults to true. |
+| update_repeatable | boolean | If an update operation failed, this flag indicates whether this update can be repeated or not. If `true`, the same update operation MAY be repeated and MAY succeed; if `false`, repeating the same update operation will fail again. This field MUST NOT be present for errors of other operations. Defaults to true. |
+| retry_after | integer | This field suggests how long (in seconds) the Platform SHOULD wait until it repeats the operation. If this a negative number, the Platform SHOULD NOT automatically repeat the operation. Defaults to 0 seconds. |
```
tinygrasshopper: I am not getting the negative number bit for `retry_after`, is that indicating the same thing as `update_repeatable`?

fmui: There is a difference. `update_repeatable` indicates wether a repeated update has a chance to succeed or not. A negative `retry_after` value indicates that even if a subsequent update could succeed, the platform should not automatically retry it.

fmui:
@georgi-lozev I agree. The right approach would be to return these states in the response of the fetch endpoint. But there are brokers that cannot provide a fetch endpoint. The only way I can think of that works for all brokers to transport this information is the error response. Do you have a better idea?

jberkhahn:
This seems reasonable from a k8s perspective. We'd have to plumb this through to our controller, but most of these values already have corresponding values in our reconciliation logic that these could be plugged into

spec.md
```diff
@@ -772,15 +775,20 @@ For success responses, the following fields are defined:
...
+| description | string | A user-facing message that can be used to tell the user details about the status of the operation. If present, MUST be a non-empty string. |
+| instance_usable | boolean | If an update or deprovisioning operation failed, this flag indicates whether or not the Service Instance is still usable. If `true`, the Service Instance can still be used, `false` otherwise. This field MUST NOT be present for errors of other operations. Defaults to true. |
+| update_repeatable | boolean | If an update operation failed, this flag indicates whether this update can be repeated or not. If `true`, the same update operation MAY be repeated and MAY succeed; if `false`, repeating the same update operation will fail again. This field MUST NOT be present for errors of other operations. Defaults to true. |
+| retry_after | integer | If an operation failed, this field suggests how long (in seconds) the Platform SHOULD wait until it repeats the operation. If this a negative number, the Platform SHOULD NOT automatically repeat the operation. Defaults to 0 seconds. |
```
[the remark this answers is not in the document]
tinygrasshopper: @georgi-lozev the `retry_after` field if for when the update operation itself has to be retried whereas the `Retry-After` is to retry the last operation polling.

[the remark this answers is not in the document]
georgi-lozev: Yes, I get the idea.

My question is why we won't re-use the same mechanism with the header to retry the update operation, but we introduce a new one, a field that is part of the body.
To me it looks like we can return a `Retry-After` header as part of the update operation to instruct the platform retrying after a given interval, same way as in last operation polling.

I don't think it's a showstopper, it's more kind of a nitpicking and aiming for some kind of consistency between implementing similar behaviors in the spec.

tinygrasshopper:
Thinking a bit more about this, I think this PR does 3 things:
- (Feature 1) Adds a mechanism to denote health/usability of a service instance.
- (Feature 2) Adds a mechanism to denote if subsequent updates are possible on a service instance
- (Feature 3) Adds a mechanism to inform the platform when the subsequent update can be trigged

A comment on Feature 3:
Are we are exposing the complexity of the retrying mechanism at two places in the spec, which I think is unnecessary. I think broker authors can leverage the existing `Retry-After` header to signify that the operation is still in progress today.

Today retry of an update can be encapsulated in the Last Operation invocation:
now

Where as after this PR, in addition to the above mechanism broker authors will also have the ability to retry the update itself:
after

What are the benefits of introducing an alternate mechanism for retry? Can we achieve the same result by adding the health/usability indicator to last operation?

mattmcneeney:
@fmui Do we still need this PR?

PR #648, opened 2019-03-05
Add request identity header

jberkhahn:
per #584

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -9,6 +9,7 @@
   - [Headers](#headers)
     - [API Version Header](#api-version-header)
     - [Originating Identity](#originating-identity)
+    - [Request Identity](#request-identity)
   - [Platform to Service Broker Authentication](#platform-to-service-broker-authentication)
   - [URL Properties](#url-properties)
   - [Service Broker Errors](#service-broker-errors)
@@ -177,6 +178,7 @@ The following HTTP Headers are defined for the operations detailed in this spec:
 | --- | --- | --- |
 | X-Broker-API-Version* | string | See [API Version Header](#api-version-header). |
 | X-Broker-API-Originating-Identity | string | See [Originating Identity](#originating-identity). |
+| X-Broker-API-Request-Identity | string | See [Request Identity](#request-identity). |
 
 \* Headers with an asterisk are REQUIRED.
 
@@ -255,6 +257,26 @@ end-user of the Platform. For example, during
 Broker's catalog, the Platform might not have an end-user with which to
 associate the request, therefore in those cases the originating identity header
 would not be included in those messages.
+
+### Request Identity
+
+A Platform might wish to uniquely identify a specific request as it flows throughout the system.
+For example, this might be used for logging for request tracing purposes. In order to facilitate
+this, Platforms will need to provide identification information to the Service Broker for each
+request. Platforms MAY support this feature, and if they do, they MUST adhere to the following:
+- For any OSBAPI request, there MUST be an associated `X-Broker-Request-Identity` header on
+  the HTTP request.
+- The Service Broker MAY include this value in log messages generated as a result of the request.
+- The Service Broker SHOULD include this header in the response to the request.
+The format of the header MUST be:
+
+```
+X-Broker-API-Request-Identity: value
+```
+
+`value` MUST be a non-empty string indicating the identity of the
+request being sent. The specific value MAY be unique for each request
+sent to the broker. Using a GUID is RECOMMENDED.
 
 ## Vendor Extension Fields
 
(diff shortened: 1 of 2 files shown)


spec.md
```diff
@@ -256,6 +258,26 @@ Broker's catalog, the Platform might not have an end-user with which to
...
+### Request Identity
+
+A platform might wish to uniquely identify a specific request as it flows throughout the system.
+For example, this might be used for logging for request tracing purposes. In order to facilitate
+this, platforms will need to provide identifacation information to the Service Broker for each
+request. Platforms MAY support this feature, and if they do, they MUST adhere to the following:
+- For any OSBAPI request, there MUST be an associated `X-Broker-Request-Identity` header on
+  the HTTP request
+- The Service Broker MAY include this value in log messages generated as a result of the request
+- The Service Broker MUST include this header in the response to the request
```
fmui: I guess this MUST has to be a SHOULD. Otherwise it's a breaking change.

fmui:
Could you please also update the swagger file.

williammartin:
What problem is this solving? I mean, I understand some reasons we might want this, but has there been any use case to drive this addition? I don't see it captured in the issue.

jberkhahn:
Changes updated.

jberkhahn:
Also, we discussed @williammartin's question in the meeting this morning, but for the record, this is to help debugging and tracing of a request as it flows through the systems. So, if  a request in the platform is marked with 'foo', that will show up in the platform logs. With this PR, that same request on the broker end will also be tagged with 'foo', so you know which action on the platform side links up with which request on the broker side.

Hope that helps.

mattmcneeney:
@williammartin @laurzca I see the benefit of using a spec-defined field for this ID. Today we use `x-vcap-request-id`, but this isn't documented anywhere for service broker authors so they don't know to expect it.

Thoughts on how we could make use of this:
1) Send both `X-Broker-API-Request-Identity` and `x-vcap-request-id`, with the value we send today in `x-vcap-request-id`
2) Deprecate `x-vcap-request-id` for service broker requests, and replace it with `X-Broker-API-Request-Identity`
3) Do nothing and add the `X-Broker-API-Request-Identity` field to the Kubernetes section of `profile.md`

spec.md
```diff
@@ -256,6 +258,26 @@ Broker's catalog, the Platform might not have an end-user with which to
 associate the request, therefore in those cases the originating identity header
 would not be included in those messages.
 
+### Request Identity
+
+A platform might wish to uniquely identify a specific request as it flows throughout the system.
+For example, this might be used for logging for request tracing purposes. In order to facilitate
+this, platforms will need to provide identifacation information to the Service Broker for each
+request. Platforms MAY support this feature, and if they do, they MUST adhere to the following:
```
waterlink: This is a good explanation on why request identity is needed, thank you!

waterlink:
@mattmcneeney I personally see `1.` option as the simplest one. Is there a reason we’d want to deprecate `x-vcap-request-id`?

@jberkhahn the PR looks good to me 👍

mattmcneeney:
> Is there a reason we’d want to deprecate x-vcap-request-id?

@waterlink I don't have a reason, other than not sending duplicate data, but that is almost definitely better than removing a header a service broker author is relying on.

spec.md
mattmcneeney: s/platform/Platform

spec.md
```diff
@@ -256,6 +258,26 @@ Broker's catalog, the Platform might not have an end-user with which to
...
+- For any OSBAPI request, there MUST be an associated `X-Broker-Request-Identity` header on
```
mattmcneeney: Please can we add periods to the end of these bullet points? :)

spec.md
```diff
@@ -256,6 +258,26 @@ Broker's catalog, the Platform might not have an end-user with which to
...
+- The Service Broker SHOULD include this header in the response to the request
+The format of the header MUST be:
+
+```
+X-Broker-API-Request-Identity: value
+```
+
+`value` MUST be a non-empty string indicating the identity of the
+request being sent. The specific value MAY be unique for each request
+sent to the broker. Using a GUID is RECOMMENDED
```
mattmcneeney: s/RECOMMENDED/RECOMMENDED.

swagger.yaml
```diff
@@ -352,6 +356,11 @@ parameters:
     in: header
     description: identity of the user that initiated the request from the Platform
     type: string
+  RequestIdentity:
+    name: X-Broker-API-Request-Identity
+    in: header
+    description: idenity of the request from the platform
```
mattmcneeney: s/platform/Platform

mattmcneeney:
Looks good, but there is a merge conflict @jberkhahn :(

Issue #654, opened 2019-04-08
Add K8S and CF annotations to context object and CF bind resource


gberche-orange:
**What is the problem?**

As an application developer, I need some implicit context from my platform to be provided to services I provision which includes:
* where to direct service-related alerts when the service is out of its SLA (e.g. a slack channel, or email address, or paging number)
* an application/organization unit attached to the application for associating the service with chargeback/inventories/cmdb 

As an application developer, I would like the annotations provided to my services to default to the ones I have set in my platform, so I don't have to explicitly define them on every service I request provisionning of (e.g. as explicit parameters)

Example of platform hosted annotations may set:
* [cloudfoundry annotations](http://v3-apidocs.cloudfoundry.org/version/3.68.0/index.html#metadata) which can be set or [orgs](http://v3-apidocs.cloudfoundry.org/version/3.68.0/index.html#organizations), [spaces](http://v3-apidocs.cloudfoundry.org/version/3.68.0/index.html#spaces), apps, and soon service-instances (currently [being implemented](https://www.pivotaltracker.com/n/projects/966314/stories/164144574))
* [K8S annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/) which can be set for instance on my namespace, or a [ServiceInstance](https://kubernetes.io/docs/concepts/extend-kubernetes/service-catalog/)

**Who does this affect?**

* Application developers: (implicitly) set annotations
* Service authors: their broker read and use annotations
* potentially platform operators: may operate intermediate layers such as "Peripli Service Manager" that would read/use such annotations, potentially without service authors being aware of the existence of such annotations.

**Do you have any proposed solutions?**

The [CF and K8S Context](https://github.com/openservicebrokerapi/servicebroker/blob/master/profile.md#context-object) get enriched with a set of annotations:
*  for CF, the following fields containing a subset of the corresponding entity annotations, formatted as a json hash with key/value string pairs
   *  `organization_annotations` 
   *  `space_annotations` 
   *  `instance_annotations`
*  for K8S, the following fields containing a subset of the corresponding entity annotations (i.e. [public ones](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set) starting with a non empty prefix), formatted as a json hash with key/value string pairs
   *  `namespace_annotations` 
   *  `serviceinstance_annotations`

[...]

PR #658, opened 2019-04-24
Add K8S and CF annotations to context object and CF bind resource

gberche-orange:

Addresses #654

**Checklist:**
- [x] The [swagger.yaml](swagger.yaml) doc has been updated with any required changes (no changes required: the Context object is not strongly typed)

https://github.com/openservicebrokerapi/servicebroker/blob/cc11bc5fccd65b36882b2c9c8ef9a56c53dbe41b/swagger.yaml#L676-L681

- [x] The [openapi.yaml](openapi.yaml) doc has been updated with any required changes (no changes required: the Context object is not strongly typed)

https://github.com/openservicebrokerapi/servicebroker/blob/d1e1273a636b793d81892a30f5ef8e5f833845b5/openapi.yaml#L854-L856

### DIFF
--- a/profile.md
+++ b/profile.md
@@ -204,11 +204,11 @@ deployment:
   This OPTIONAL property holds an object with the annotations key/value pairs. If present, this property MUST be an object, with zero or more properties as follows:
   
   ```
-  "organization_annotations": { "key-here":"value-here" }
+  "organization_annotations": { "prefix-here.org/name-here":"value-here" }
   ```
   For example:
   ```
-  "organization_annotations": { "send-alerts-to-email":"me@mycompany.com" }
+  "organization_annotations": { "myprovider.com/send-alerts-to-email":"me@mycompany.com" }
   ```
 
 
@@ -250,11 +250,11 @@ deployment:
   This OPTIONAL property holds an object with the annotations key/value pairs. If present, this property MUST be an object, with zero or more properties as follows:
   
   ```
-  "space_annotations": { "key-here":"value-here" }
+  "space_annotations": { "prefix-here.org/name-here":"value-here" }
   ```
   For example:
   ```
-  "space_annotations": { "send-alerts-to-email":"me@mycompany.com" }
+  "space_annotations": { "myprovider.com/send-alerts-to-email":"me@mycompany.com" }
   ```
 
 
@@ -304,11 +304,11 @@ Cloud Foundry API call:
   This OPTIONAL property holds an object with the annotations key/value pairs. If present, this property MUST be an object, with zero or more properties as follows:
   
   ```
-  "instance_annotations": { "key-here":"value-here" }
+  "instance_annotations": { "prefix-here.org/name-here":"value-here" }
   ```
   For example:
   ```
-  "instance_annotations": { "send-alerts-to-email":"me@mycompany.com" }
+  "instance_annotations": { "myprovider.com/send-alerts-to-email":"me@mycompany.com" }
   ```
 
 
@@ -343,11 +343,11 @@ The following properties are defined for usage within a Kubernetes deployment:
   This OPTIONAL property holds an object with the annotations key/value pairs. If present, this property MUST be an object, with zero or more properties as follows:
   
   ```
-  "namespace_annotations": { "key-here":"value-here" }
+  "namespace_annotations": { "prefix-here.org/name-here":"value-here" }
   ```
   For example:
   ```
-  "namespace_annotations": { "send-alerts-to-email":"me@mycompany.com" }
+  "namespace_annotations": { "myprovider.com/send-alerts-to-email":"me@mycompany.com" }
   ```
 
 - `serviceinstance_annotations`
@@ -359,11 +359,11 @@ The following properties are defined for usage within a Kubernetes deployment:
   This OPTIONAL property holds an object with the annotations key/value pairs. If present, this property MUST be an object, with zero or more properties as follows:
   
   ```
-  "serviceinstance_annotations": { "key-here":"value-here" }
+  "serviceinstance_annotations": { "prefix-here.org/name-here":"value-here" }
   ```
   For example:
   ```
-  "serviceinstance_annotations": { "send-alerts-to-email":"me@mycompany.com" }
+  "serviceinstance_annotations": { "myprovider.com/send-alerts-to-email":"me@mycompany.com" }
   ```
 
 
@@ -443,11 +443,11 @@ deployment:
   This OPTIONAL property holds an object with the annotations key/value pairs. If present, this property MUST be an object, with zero or more properties as follows:
   
   ```
-  "space_annotations": { "key-here":"value-here" }
+  "space_annotations": { "prefix-here.org/name-here":"value-here" }
   ```
   For example:
   ```
-  "space_annotations": { "send-alerts-to-email":"me@mycompany.com" }
+  "space_annotations": { "myprovider.com/send-alerts-to-email":"me@mycompany.com" }
   ```
 
 
--- a/profile.md
+++ b/profile.md
@@ -201,7 +201,7 @@ deployment:
 
   The [annotations](http://v3-apidocs.cloudfoundry.org/version/3.69.0/index.html#metadata) attached to the organization that a Service Instance is associated with.
   Note that the annotations of an organization in Cloud Foundry MAY be changed.
-  This OPTIONAL property holds an object with the annotations key/value pairs. If present, this property MUST be an object, with zero or more properties as follows:
+  This OPTIONAL property holds an object with the annotations key/value pairs. If present, this property MUST be an object, with one or more properties as follows:
   
   ```
   "organization_annotations": { "prefix-here.org/name-here":"value-here" }
@@ -247,7 +247,7 @@ deployment:
 
   The [annotations](http://v3-apidocs.cloudfoundry.org/version/3.69.0/index.html#metadata) attached to the space that a Service Instance is associated with.
   Note that the annotations of a space in Cloud Foundry MAY be changed.
-  This OPTIONAL property holds an object with the annotations key/value pairs. If present, this property MUST be an object, with zero or more properties as follows:
+  This OPTIONAL property holds an object with the annotations key/value pairs. If present, this property MUST be an object, with one or more properties as follows:
   
   ```
   "space_annotations": { "prefix-here.org/name-here":"value-here" }
@@ -301,7 +301,7 @@ Cloud Foundry API call:
 
   The [annotations](http://v3-apidocs.cloudfoundry.org/version/3.69.0/index.html#metadata) attached to the Service Instance.
   Note that the annotations of a Service Instance in Cloud Foundry MAY be changed.
-  This OPTIONAL property holds an object with the annotations key/value pairs. If present, this property MUST be an object, with zero or more properties as follows:
+  This OPTIONAL property holds an object with the annotations key/value pairs. If present, this property MUST be an object, with one or more properties as follows:
   
   ```
   "instance_annotations": { "prefix-here.org/name-here":"value-here" }
@@ -340,7 +340,7 @@ The following properties are defined for usage within a Kubernetes deployment:
 
   The [annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/) attached to the namespace in which the Service Instance will be visible.
   Note that the annotations of a namespace in Kubernetes MAY be changed.
-  This OPTIONAL property holds an object with the annotations key/value pairs. If present, this property MUST be an object, with zero or more properties as follows:
+  This OPTIONAL property holds an object with the annotations key/value pairs. If present, this property MUST be an object, with one or more properties as follows:
   
   ```
   "namespace_annotations": { "prefix-here.org/name-here":"value-here" }
@@ -356,7 +356,7 @@ The following properties are defined for usage within a Kubernetes deployment:
 
   The [annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/) attached to the Service Instance.
   Note that the annotations of a Service Instance in Kubernetes MAY be changed.
-  This OPTIONAL property holds an object with the annotations key/value pairs. If present, this property MUST be an object, with zero or more properties as follows:
+  This OPTIONAL property holds an object with the annotations key/value pairs. If present, this property MUST be an object, with one or more properties as follows:
   
   ```
   "serviceinstance_annotations": { "prefix-here.org/name-here":"value-here" }
@@ -440,7 +440,7 @@ deployment:
 
   The [annotations](http://v3-apidocs.cloudfoundry.org/version/3.69.0/index.html#metadata) attached to the application that a Service Binding is associated with.
   Note that the annotations of an application in Cloud Foundry MAY be changed.
-  This OPTIONAL property holds an object with the annotations key/value pairs. If present, this property MUST be an object, with zero or more properties as follows:
+  This OPTIONAL property holds an object with the annotations key/value pairs. If present, this property MUST be an object, with one or more properties as follows:
   
   ```
   "app_annotations": { "prefix-here.org/name-here":"value-here" }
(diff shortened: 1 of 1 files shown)


profile.md
```diff
@@ -261,6 +295,23 @@ Cloud Foundry API call:
...
+- `instance_annotations`
+
+  Version: 2.15
+
+  The [annotations](http://v3-apidocs.cloudfoundry.org/version/3.69.0/index.html#metadata) attached to the Service Instance.
+  Note that the annotations of a service instance in Cloud Foundry MAY be changed.
+  This OPTIONAL property holds an object with the annotations key/value pairs. If present, this property MUST be an object, with zero or more properties as follows:
+  
+  ```
+  "space_annotations": { "key-here":"value-here" }
```
waterlink: Did you mean `"instance_annotations"` here?

profile.md
```diff
@@ -283,6 +334,39 @@ The following properties are defined for usage within a Kubernetes deployment:
   "namespace": "testing"
   ```
 
+- `namespace_annotations`
+
+  Version: 2.15
+
+  The [annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/) attached to the namespace in which the Service Instance will be visible.
+  Note that the annotations of an namespace in Kubernetes MAY be changed.
```
waterlink: typo here: `a namespace`

waterlink:
I’m concerned about: what will happen if a service-broker wants to support a 3rd platform (not CF and not K8S), and that platform has a slightly different concept? Will we have to add one or two more `*_annotation`? Should then all service broker authors make changes to support these? Or should service brokers be able to handle `any_annotation` as long as they can recognise the keys inside (like your example `send-email-alert-to`)?

Another thing that I’ve noticed: `instance_annotations` and `serviceinstance_annotations` seem to be equivalent to me (one comes from CF, and another from K8S). Do you think it makes sense to merge them into a single key?

gberche-orange:
thanks @waterlink for your review

> I’m concerned about: what will happen if a service-broker wants to support a 3rd platform (not CF and not K8S), and that platform has a slightly different concept? Will we have to add one or two more *_annotation? 

The extension of the profile.md to additional platforms beyond CF and K8S does not seem specific to annotations. The current specs do not state whether new plaforms names can be sent without changing the specs to register their platform name.

> Should then all service broker authors make changes to support these? 

I would expect service broker to only parse/handle context objects for platforms they are aware of, and either ignore or reject unsupported platform types as they would not be sure of what context keys to expect.

> Or should service brokers be able to handle any_annotation as long as they can recognise the keys inside (like your example send-email-alert-to)?

Yes, generally brokers can't make any assumptions about the user provided annotations that they will receive in the context objects. They may choose to document the annotations they support/expect/require (e.g. in their documentation page url specified in catalog), and ignore annotations they are unaware of.

I can imagine that if a broker absolutely requires a mandatory inputs, it may offer its users to specify them at a higher level through annotations or explicitly through parameters that are documented in the catalog through associated Json schema.  

We'll have to see in the future if there is demand for brokers to publish their supported annotations directly in the catalog through additional json schemas (e.g. using a key `annotations` in place of the existing [parameters](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#input-parameters-schema-object))

gberche-orange:
@mattmcneeney the travis-ci build is apparently failing following your commit https://github.com/openservicebrokerapi/servicebroker/commit/275d5faa8f8f7ea77993d5cdc33d9b7b0cea4703 with the following message, 

> tools/../.github/pull_request_template.md: Can't find: tools/../.github/swagger.yaml
> tools/../.github/pull_request_template.md: Can't find: tools/../.github/openapi.yaml

See [build 524511401](https://travis-ci.org/openservicebrokerapi/servicebroker/builds/524511401?utm_source=github_status&utm_medium=notification) for full traces.

mattmcneeney:
Ah! Thanks @gberche-orange - I'm on the case now...

mattmcneeney:
I believe this is fixed @gberche-orange if you can rebase this!

waterlink:
@gberche-orange Thank you for addressing my concerns!

I still feel a bit weird about:
- CF `instance_annotations`, and
- K8S `serviceinstance_annotations`.

These seem to be terms for the same thing. So as a reader, my question: “Why they are not called the same thing (e.g. just `instance_annotations`)?”

Then my next thought is: “Oh, one comes from CF and another comes from K8S!” after reading a bit deeper.

And then my further line of thinking is: “Okay if we need to distinguish between CF and K8S annotations, then why not call them `cf_instance_annotations` and `k8s_instance_annotations`? This seems to be more clear, and deliver intent better.”

And when I’m asking this question to myself, I can come up with one hypothesis with my current knowledge: “Maybe we don’t want to mention `cf` and `k8s` in the JSON too much…”

Can you please help me understand if my line of thought is correct or not, and where did I take the wrong turn?

Other than that, the PR looks good to me! :+1:

gberche-orange:
sorry @waterlink I missed this question in my previous answers

My rationales for keeping CF `instance_annotations`, and K8S `serviceinstance_annotations` keys distinct and not prefixed by their related platforms were:
* I named CF `instance_annotations` to remain consistent with the `instance_name` in the [Cloud Foundry Context Object](https://github.com/openservicebrokerapi/servicebroker/blob/master/profile.md#cloud-foundry-context-object) and  K8S `serviceinstance_annotations` to remain consistent with the [ServiceInstance service catalog resource](https://kubernetes.io/docs/concepts/extend-kubernetes/service-catalog/) as I felt `instance` was ambiguous. I'm happy to revert that to `instance_annotations` if no one feels the same ambiguity and if consistency among K8S and CF context is felt more important.
* The Context object contains a property [platform](https://github.com/openservicebrokerapi/servicebroker/blob/master/profile.md#context-object-properties) which will be set as either `cloudfoundry` or `kubernetes`. I don't believe we need to disambiguate other properties in the context object since the `platform` key plays this role.

waterlink:
@gberche-orange if we do make this key the same, then `serviceinstance_annotations` for both will make more sense because: 

- this is less ambiguous, and
- even in CF Services API (sapi) team we call them "service instances" in our spoken language, user story writing, and the code: both backend and CLI.

gberche-orange:
@mattmcneeney any thoughts on CF `instance_annotations` key consistency with the `instance_name` key that you seem to have authored recently in https://github.com/openservicebrokerapi/servicebroker/commit/0430220efcf861c52cbb4ca80d65f5606f09058d#diff-3b9a1a89335a911c0c69b58b043febbeR227 ?

mattmcneeney:
I'm not sure I've quite followed this. Are we proposing that the `context` object has both `instance_annotations` and `serviceinstance_annotations`?

gberche-orange:
> Are we proposing that the context object has both instance_annotations and serviceinstance_annotations?

No. 

The CF context in this PR has an `instance_annotations` key to remain consistent with the `instance_name` you authored in the [Cloud Foundry Context Object](https://github.com/openservicebrokerapi/servicebroker/blob/master/profile.md#cloud-foundry-context-object) and  the K8S context has a `serviceinstance_annotations` to remain consistent with the [ServiceInstance service catalog resource](https://kubernetes.io/docs/concepts/extend-kubernetes/service-catalog/).

@waterlink is suggesting in https://github.com/openservicebrokerapi/servicebroker/pull/658#issuecomment-486832531 to instead name the both keys in CF context and K8S context as `serviceinstance_annotations`

This creates an inconsistency in the CF context between  `instance_annotations` and `instance_name` keys.

I was keen to get your opinion on whether you prefer :
- a) consistency across K8S and CF contexts (ie both keys named `serviceinstance_annotations`) 
- b) consistency in the CF context itself (as currently in this PR).

mattmcneeney:
Ah I've got it, thank you!

If possible, it would be good to use `instance_annotations` for both CF and K8S, to be consistent with the existing `instance_name` field.

However, if this isn't possible, I'd probably vote for consistency across K8S and CF contexts, and call them both `serviceinstance_annotations`. We would leave `instance_name` as it is.

gberche-orange:
thanks @mattmcneeney and @waterlink for inputs 

> If possible, it would be good to use instance_annotations for both CF and K8S, to be consistent with the existing instance_name field.

I've applied this in https://github.com/openservicebrokerapi/servicebroker/pull/658/commits/12b46c75890ce1eacdc7c469f80c40e34945b0ae

@waterlink 

please note that I also pushed further commits to:
* make example consistent with K8S annotations format, that will be hopefully soon also enforced by CF (see https://github.com/cloudfoundry/cloud_controller_ng/issues/1335)
* fix typo on Bind Resource Object annotation key example
* specify that when present the annotations must be non-empty. The rationale for this is to:
   * optimize the payload by avoiding that platforms send empty annotations keys
   * simplify service authors work that don't need to handle inconsistent platform behaviors

You may therefore need to approve again these changes.

gberche-orange:
Note that the following CF resources that do not yet have annotation support implemented might need to be exposed in a future annotation-related PR:
- route (see [story](https://www.pivotaltracker.com/story/show/164291350)). A new `route_annotations` key would be added in the CF Bind Resource object:  `annotations attached to the route that a Service Binding is associated with (i.e. the Binding is of type Route Services)`. Sample associated use-case: define implicit alerting target per route in a route service instance.
- isolation segment (see [story](https://www.pivotaltracker.com/story/show/163093413)). A new `isolation_segment_annotations` would be added to the CF context. Sample associated use-case: define an implicit alerting target for all orgs in an isolation segments.

waterlink:
Also, Travis failure seem to be unrelated and seems to be an internet connectivity problem that Travis’ VM/container had. Someone who has permissions should re-trigger that.

profile.md
```diff
@@ -195,6 +195,23 @@ deployment:
...
+- `organization_annotations`
+
+  Version: 2.15
+
+  The [annotations](http://v3-apidocs.cloudfoundry.org/version/3.69.0/index.html#metadata) attached to the organization that a Service Instance is associated with.
+  Note that the annotations of an organization in Cloud Foundry MAY be changed.
+  This OPTIONAL property holds an object with the annotations key/value pairs. If present, this property MUST be an object, with one or more properties as follows:
+  
+  ```
+  "organization_annotations": { "prefix-here.org/name-here":"value-here" }
```
waterlink: @gberche-orange In the OSBAPI call we’ve realised that CF won’t have the prefixes here.

Is there expectation that the annotation object will always need such prefixes?, or is it only for K8S example?

gberche-orange: @waterlink

I was hoping that the CF CAPI team would handle https://github.com/cloudfoundry/cloud_controller_ng/issues/1335 bringing planned Cf/K8S consistency and that prefixes could be used to filter out private annotation keys as mentionned in https://github.com/openservicebrokerapi/servicebroker/issues/654#issuecomment-485655814 in a consistent way between CF and K8S:

> As a 1st step, we can leverage K8S annotations naming conventions reproduced below to allow users to distinguish between public annotations (shared with all brokers) and private annotations (not shared with brokers):

I'm still waiting for CF CAPI response on this. The CF SAPI team could consider other alternative options for filtering private annotations though, and we could relax the CF examples in OSB API PR

fmui:
Not sure if it fits into this PR or should be another one, but if we add instance annotations for K8S, I think we should also have the instance name in the context object.

jberkhahn:
Looks like we're still waiting for the CF CAPI team to get back to us on this one?

georgi-lozev:
It looks like that to me. 

https://github.com/cloudfoundry/cloud_controller_ng/issues/1335#issuecomment-496116569

gberche-orange:
CF CAPI team just confirmed they will be prioritizing enforcing the K8S prefix conventions to CF annotations, see https://github.com/cloudfoundry/cloud_controller_ng/issues/1335#issuecomment-498734192

mattmcneeney:
@gberche-orange can this PR be merged before the associated CC change? @waterlink has kindly offered to take a look into this

gberche-orange:
@mattmcneeney Yes, I believe this PR can be merged before user input validation by CF CC enforces the K8S prefix into the user-provided annotations. The examples in this PR will conform to future enforced model by both CF and K8S. 

thanks @waterlink !

waterlink:
@mattmcneeney @gberche-orange Yes, this can go forward, since [this RFC](https://github.com/cloudfoundry/cloud_controller_ng/blob/master/decisions/0004-adding-key-prefix-to-annotations.md) was accepted in CAPI.

mattmcneeney:
@zrob Are you happy for us to merge this into the spec, knowing that the associated CC change is in the [backlog somewhere](https://github.com/cloudfoundry/cloud_controller_ng/issues/1335#issuecomment-498734192)?

zrob:
Sorry I'm behind the 8-ball on this proposal.  Is the intent of this PR that ALL annotations of the various resources org,space,app be sent to a broker for every bind request?

profile.md
```diff
@@ -350,6 +434,25 @@ deployment:
   "space_guid": "15823972-c216-4ba5-9f3f-e75b0b891297"
   ```
 
+- `app_annotations`
+
+  Version: 2.15
+
+  The [annotations](http://v3-apidocs.cloudfoundry.org/version/3.69.0/index.html#metadata) attached to the application that a Service Binding is associated with.
+  Note that the annotations of an application in Cloud Foundry MAY be changed.
+  This OPTIONAL property holds an object with the annotations key/value pairs. If present, this property MUST be an object, with one or more properties as follows:
```
waterlink: I’m writing a user story for this right now, and I’ve discovered an interesting scenario: when you want to remove all annotations (e.g. make public annotations -> private), then there is a problem with wording:

> If present, this property MUST be an object, with one or more properties

Because the only way I can delete all annotations is to send an empty object, and according to the line above that wouldn’t be valid.

@gberche-orange What do you think? should we fix that to `zero or more`?

profile.md
```diff
@@ -350,6 +434,25 @@ deployment:
   "space_guid": "15823972-c216-4ba5-9f3f-e75b0b891297"
   ```
 
+- `app_annotations`
+
+  Version: 2.15
```
waterlink: Probably all the versions will have to change to `2.16` now since the working group has cut a release.

waterlink:
User story in SAPI backlog to track the implementation of this on CF side: https://www.pivotaltracker.com/n/projects/2105761/stories/166935990

profile.md
gberche-orange: @waterlink thanks, good catch. I agree allowing for empty content will enable clearing previously set annotations during an update. I'll push new commits for fixing this.

gberche-orange:
> User story in SAPI backlog to track the implementation of this on CF side: https://www.pivotaltracker.com/n/projects/2105761/stories/166935990

thanks @waterlink !

Related updating scenarios 3 and 4, I wonder whether in the future changes to annotations (org/space/service-instance) should automatically trigger a PATCH  `/v2/service_instances/:instance_id` request with updated annotations (i.e. without user triggering a `cf update-service`, similar to https://www.pivotaltracker.com/story/show/163861852 (in which PATCH request was triggered by a `cf rename-service` and not an explicit `cf update-service`). 

Currently with cf cli 6.45 a `cf update-service` command without argument seems to not trigger calls to cloudcontroller to update the instance:

```
cf update-service mysql
OK
No changes were made

CF_TRACE=true cf update-service mysql | grep -A1 REQUEST
REQUEST: [2019-06-27T09:23:16+02:00]
GET /v2/info HTTP/1.1
--
REQUEST: [2019-06-27T09:23:16+02:00]
GET /login HTTP/1.1
```

profile.md
gberche-orange: thanks @waterlink, resolved in https://github.com/openservicebrokerapi/servicebroker/pull/658/commits/80268a5fc4d73d3778cb80424424cad8e9cd64a5

profile.md
gberche-orange: resolved in https://github.com/openservicebrokerapi/servicebroker/pull/658/commits/80268a5fc4d73d3778cb80424424cad8e9cd64a5

mattmcneeney:
@blgm Do you know if there is a story in the SAPI backlog for this? Happy to approve but just want to make sure this work is tracked somewhere

gberche-orange:
@mattmcneeney thanks for merging this PR!

I notice this PR does not appear in the roadmap project for 2.16 https://github.com/openservicebrokerapi/servicebroker/projects/1 

Would it make sense to associate this PR to the 2.16 milestone in the roadmap?

mattmcneeney:
Good spot, thanks @gberche-orange - project updated!

gberche-orange:
thanks @mattmcneeney !

Issue #641, opened 2019-02-25
Suggest mechanism to notice and trigger maintenance of a Service Instance


williammartin:
Given some disagreement on https://github.com/openservicebrokerapi/servicebroker/pull/628 about the need for arbitrary data to describe the changes in some particular maintenance, I'd like to propose we tackle this problem in two steps.

**What are the problem statements?**

1. As an application developer, I want to be able to update my own service instances when convenient to me, so that I can schedule downtime.
1. As a platform operator, I want to be able to update out of date service instances, so I can manage offerings reasonably.

**What is required to support this?**

1. There MUST be some indication that an update is available for a service instance
1. There MUST be some way to trigger an update of a service instance
1. There SHOULD be some indication of the impact of an update

**What changes are proposed to the spec?**

Since there is already a way to update a service instance via a `PATCH`:

1. There MUST be some indication that the service broker now provides an updated version of a service
1. There SHOULD be some indication of the impact of an update

I feel like the first point is uncontroversial, and the second is where there is some disagreement about whether an arbitrary data structure is required, or whether a string could suffice. In either case, we should have a structure we can extend to capture this:

```
maintenance_info: {
  version: "1.2.3+abcedf",
}
```

This should provide sufficient information for the Platform to determine (and provide this information to an Application Developer/Platform Operator), and trigger (as instructed by an Application Developer/Platform Operator) maintenance.

I can open a separate PR to deal with point 2, but the summary can be found below.

---

**Why should there be an indication of what is available in an update?**

One of our customers strongly desires the ability to show details of specific changes in a service, for example, changes in middleware. One way we could solve this is to have a `details` string as an absolute representation about what is contained in this version:

```
maintenance_info: {
  version: "1.2.3+abcedf",
  details: "stemcell: 224.6, os: 12.4",
}
```

Which, if the platform compared to a previously stored version might produce an experience like:

```
$ cf service s1

name:            s1
service:         example-service
tags:
plan:            example-plan
description:     ...

Showing status of last operation from service s1...

status:    create succeeded
message:
started:   2019-01-02T14:10:47Z
updated:   2019-01-02T14:10:47Z

An update is available for this service instance:
  Current: stemcell: 220.1, os: 12.4
  Update:  stemcell: 224.6, os: 12.4
```

**So why might this not be sufficient or unpleasant?**

1. Unstructured data in string format is not easily machine consumable
1. It is hard to computationally determine what has changed in an update
[...]

PR #659, opened 2019-05-07
Add maintenance info versions fetching endpoint for service plan.

waterlink:
This implements a solution for #641 , specifically the point about:
"There SHOULD be some indication of the impact of an update"

Solution:
To expose maintenance info versions and underlying details using a new endpoint 
`/v2/service_plans/:plan_id/maintenance_info_versions` that has a defined structure and info that the Platform can use to display detailed information about the update.

**Checklist:**
_(TODO: will do that after getting initial feedback on this)_
- [ ] The [swagger.yaml](swagger.yaml) doc has been updated with any required changes
- [ ] The [openapi.yaml](openapi.yaml) doc has been updated with any required changes

cc @AartiKriplani

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -675,6 +675,109 @@ schema being used.
 | --- | --- | --- |
 | version | string | This MUST be a string conforming to a semantic version. The Platform MAY use this field to determine whether an update is available for a Service Instance. |
 
+## Fetching Service Plan Maintenance Info Versions 
+
+The Service Broker SHOULD expose maintenance info versions and underlying details using this endpoint (/v2/service_plans/:plan_id/maintenance_info_versions) so that the Platform MAY show the user the difference between the current maintenance info version of the Service Instance and the one they could update to.
+
+### Request
+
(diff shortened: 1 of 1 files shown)


waterlink:
We have discussed this internally in the team. Given all the new feedback, we decided to:

1\. `description` field

- Settle down for a simple `description` field on the `maintenance_info` object in the plan.
- This field is likely to be provided by the operator who maintains the service broker installation (this is what most of our brokers will allow and encourage operators to do). For some other service brokers, this could be automatic, or could read text from the change log.
- The purpose of the `description` field is for operators (or authors) of the service broker installation to give any information to the developers, that they think they’ll need to be aware of. This could include whether there is a downtime, and whether there are any important configuration changes, etc. This should be written from humans and for humans.

2\. new endpoint described above

- we decided that there is no need for such endpoint. Addition of a single field on the plan in the catalog should allow for our use cases.

We’ll go ahead and create new PR. Closing this one.

PR #661, opened 2019-05-08
Indicate if a Service Instance is still usable after a failed update or deprovisioning and if an update can be repeated

fmui:
This PR is a subset of PR #637. It only contains the non-controversial `instance_usable` and `update_repeatable` flags.

**Checklist:**
- [X] The [swagger.yaml](swagger.yaml) doc has been updated with any required changes
- [X] The [openapi.yaml](openapi.yaml) doc has been updated with any required changes


spec.md
```diff
@@ -1734,6 +1760,14 @@ $ curl 'http://username:password@service-broker-url/v2/service_instances/:instan
 Responses with any other status code MUST be interpreted as a failure and the
 Platform MUST remember the Service Instance.
 
+When a deprovisioning fails, the Service Instance can still be usable or
+unusable or its state could be unknown. If a Service Instance became unusable,
+the Platform SHOULD NOT allow the creation of new bindings.
```
jberkhahn: I'm not sure if this is just me being pedantic, but don't we generally try to describe things in terms of the platform/broker relationship? As currently worded, this seems to be more describing the internal behavior of the platform itself. Wouldn't it make more sense if it was worded something like "If a Service Instance becomes unusable the Platform SHOULD NOT request new Service Bindings for that Service Instance."

fmui: Good point. I've changed that sentence.

georgi-lozev:
Looks good from SAPI point of view.

mattmcneeney:
@tinygrasshopper we need your vote!

PR #665, opened 2019-05-15
Clarify how maintenance_info objects should be compared.

waterlink:
- MaintenanceInfoConflict error should only be produced when
`maintenance_info.version` doesn't match. Other fields of Maintenance
Info object should be ignored in such checks.
- Adjust wording in the request/response body definitions to reflect the
above better.

This change is needed for the following reasons:
- The typo fix in the `description` of the Maintenance Info object
should not cause new `maintenance_info` version to be created.
- Platforms don't have to store or send back to the Service Broker all
the fields of the `maintenance_info`.
- In general, the `maintenance_info.version` is what matters from
perspective of checking whether two objects are same or not.

[#165995791](https://www.pivotaltracker.com/story/show/165995791)

A description of what the problem is, or a link to an issue which this is related to.

**Checklist:**
- [x] The [swagger.yaml](swagger.yaml) doc has been updated with any required changes (no changes required)
- [x] The [openapi.yaml](openapi.yaml) doc has been updated with any required changes (no changes required)


waterlink:
@mattmcneeney We’ve found out that there was an `*` (asterisk) symbol missing on the required property `version` of the `maintenance_info` object, so I pushed a new commit to fix it here. 

And GitHub decided to dismiss your review :(

waterlink:
Hey @fmui @jberkhahn @mattmcneeney 

I’ve fixed merge conflicts (because of this was merged #669). Could you please review this change again?

PR #663, opened 2019-05-14
Add 'description' field to the 'maintenance_info.'

waterlink:
This tackles the `There SHOULD be some indication of the impact of an update`
part of #641.

This is not necessarily about a description string containing versions changes
only, this may contain details about configuration changes, the default value
changes, etc.

The main use case for `maintenance_info` and `description` field is for the
plans that need maintenance done to them every 2-4 weeks or so, thus creating a
new plan every time is not an option.

[#165560405](https://www.pivotaltracker.com/story/show/165560405)

**What is the problem this PR solves?**
A description of what the problem is, or a link to an issue which this is related to.

**Checklist:**
- [x] The [swagger.yaml](swagger.yaml) doc has been updated with any required changes
- [x] The [openapi.yaml](openapi.yaml) doc has been updated with any required changes

PR #668, opened 2019-05-21
Make maintenance_info example valid json

georgi-lozev:
Fix the sample catalog to be a valid json.


georgi-lozev:
It's not an issue  anymore after https://github.com/openservicebrokerapi/servicebroker/pull/663 was merged.

PR #672, opened 2019-06-05
Rename ServiceInstanceProvision to ServiceInstance

leonwanghui:
When I use the [OpenAPITools](https://github.com/OpenAPITools/openapi-generator) tool to auto-generate Golang code from swagger.yaml (and openapi.yaml), it seems that the `ServiceInstanceProvision` defined is conflict with the generated `ServiceInstanceProvision` method. So this PR is proposed to rename `ServiceInstanceProvision` to `ServiceInstanceProvisionResponse` so as to avoid the name confusion. Besides, to keep all name consistent, `ServiceBinding` field would also been renamed to `ServiceBindingResponse`.

**Checklist:**

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -73,13 +73,13 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/ServiceInstanceProvision'
+                $ref: '#/components/schemas/ServiceInstanceProvisionResponse'
         '201':
           description: Created
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/ServiceInstanceProvision'
+                $ref: '#/components/schemas/ServiceInstanceProvisionResponse'
         '202':
           description: Accepted
           content:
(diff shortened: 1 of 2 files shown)


Samze:
I notice we already have:

```
    ServiceInstanceResource:
      type: object
      properties:
        service_id:
          type: string
        plan_id:
          type: string
        dashboard_url:
          type: string
        parameters:
          $ref: '#/components/schemas/Object'
```

I think that doing this rename may be confusing since we would have both a `ServiceInstance` object and a `ServiceInstanceResource` object. Really the `ServiceInstance` schema is specific to the provision operation.

However I notice that there is a `ServiceBinding` and `ServiceBindingResource`. 

I would be in favour for fixing both to be more explicit. Something like `ServiceInstanceProvisionResponse`.

leonwanghui:
@Samze I agree with your point, we need to make them all consistent.

leonwanghui:
@Samze @mattmcneeney @fmui PTAL, thanks!

tinygrasshopper:
I think the change makes sense, but it's a 'breaking' change for folks who are generating clients using the swagger spec. Do we need to add the breaking nature of this change to the readme

mattmcneeney:
Good point @tinygrasshopper. My guess is that there are very few folks doing that, but maybe we should try to advertise this change (emailing the mailing list?) just in case.

PR #680, opened 2019-06-24
Service Brokers SHOULD remember the state of an operation for a reasonable +amount of time

### LINKED ISSUE #359: "last_operation" behavior after having returned "succeeded"/"failed" once?
The spec says that when polling the `last_operation` endpoint for async operations:

```
A response with "state": "succeeded" or "state": "failed" MUST cause the platform to cease polling.
```

Does this mean the broker only guarantees returning a terminal state response once? If the platform has an internal issue and is unable to process that response after having hit the `last_operation` endpoint once, is it out of luck?

If so, what response should the platform expect? For everything but a 400, the platform is told to interpret it as a situation to keep polling, and our description of 400 does not encompass this situation. This means it could potentially be a long time before the platform realizes something is wrong, and thus a long time of the user being billed for a resource before the platform attempts orphan mitigation.

Personally, I feel like:
1) Brokers should be encouraged to keep the operation around for a reasonable amount of time even after returning a "succeeded"/"failed" response, in case it hasn't been fully processed.
2) We should include in the spec some behavior expectation for when the broker has removed an operation so that platforms wouldn't keep polling.
3) We should potentially give operations a broker-defined expiration date, or a way for platforms to confirm to the broker that the operation can be safely removed.

mattmcneeney:
Closes #359. Obviously, a _reasonable amount of time_ is up for debate. I'm not sure if this is a bad idea or not, especially for a specification!

**Checklist:**

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -809,6 +809,10 @@ For asynchronous provision operations, if the response contains
 `"state": "failed"` then the Platform might need to perform
 [Orphan Mitigation](#orphan-mitigation).
 
+Service Brokers SHOULD NOT forget the state of an operation immediately after
+returning a `"succeeded"` or `"failed"` state in case the Platform needs to ask
+again.
+
 ## Polling Last Operation for Service Bindings
 
 When a broker returns status code `202 Accepted` for [Binding](#binding) or
@@ -889,6 +893,10 @@ timestamp.
 
 If the response contains `"state": "failed"`, then the Platform might need to
 perform [Orphan Mitigation](#orphan-mitigation).
+
+Service Brokers SHOULD NOT forget the state of an operation immediately after
+returning a `"succeeded"` or `"failed"` state in case the Platform needs to ask
+again.
 
 ## Polling Interval and Duration
 


waterlink:
About “reasonable time.”

Right now, CF will poll the `last_operation` endpoint in the worst case scenario every 24h (when `Retry-After` header is 24h or more) when there is an operation in progress. Any of the successful polls will be recorded in CF DB.

It feels to me like “reasonable time” should be on the same order of magnitude minimum.

So a minimum of 72h perhaps makes sense.

Or did you have in mind an entirely different order of magnitude (like months or years)?

mattmcneeney:
@waterlink Good point. Since Platforms (CF/k8s) will likely have different magnitudes, my original thought was to leave this a little vague, but I'm guessing that you think this is a potentially bad idea!

@jberkhahn any thoughts from svc-cat?

jberkhahn:
seems fine. svc-cat is currently hardcoded to retry for 7 days with an exponential backoff.

mattmcneeney:
We discussed this on today's call. _'Reasonable'_ is too subjective. Let's change this to:

Service Brokers SHOULD NOT forget the state of an operation immediately after returning a `"succeeded"` or `"failed"` state in case the Platform needs to ask again.

mattmcneeney:
PR updated!

PR #681, opened 2019-06-25
Specify that Platforms SHOULD NOT reuse IDs

### LINKED ISSUE #447: Can instance_id be reused after orphan mitigation finished?
The OSB spec is not clear about whether `instance_id` is reusable after the initial provisioning request has failed and orphan mitigation (https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#orphans) has succeeded.

To give an example, let's say that the initial provisioning request has timed out. According to the OSB spec orphan mitigation is required in this case. But once it has been done, you want to try provisioning again (hoping that this time the request won't timeout). The question is whether it's okay to send a provisioning request with the same `instance_id` in that case, or you have to generate a new one.

Same applies to all `5xx` errors.

mattmcneeney:
Closes #447 

**Checklist:**

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -918,6 +918,8 @@ account on an multi-tenant SaaS application.
 This ID will be used for future requests (bind and deprovision), so the
 Service Broker will use it to correlate the resource it creates.
 
+Platforms SHOULD NOT reuse IDs, since Service Brokers MAY NOT support this.
+
 #### Parameters
 | Parameter Name | Type | Description |
 | --- | --- | --- |


fmui:
This PR conflicts with PR #680. If the ID can be reused, which state should `last_operation` return?

mattmcneeney:
I think that same logic could apply elsewhere in the spec (fetching an instance maybe?), but the mitigation here is that there is an `operation` field which could be used to reference a specific operation.

mattmcneeney:
We discussed this on today's call, and we concluded that we don't believe Platforms should reuse IDs. Some brokers wouldn't support this anyway, and in reality it seems very unlikely to happen since both CF and svc-cat generate a new GUID for each service instance.

Our options:
1. Close this PR and do nothing about this
1. Change the spec to say that platforms SHOULD NOT reuse IDs, and advise brokers what they should do in case that does happen

mattmcneeney:
Review please @jberkhahn @tinygrasshopper

PR #689, opened 2019-07-23
Add support for ETag and If-Modified-Since headers

### LINKED ISSUE #209: Expand spec to allow for "no changes" semantics to be returned by brokers
Currently when a catalog is fetched from the broker, the broker returns the entire catalog. There might be cases where generating the catalog is expensive, or the catalog returned is quite large. Some controllers might also refetch the catalog on regular basis (like kubernetes service catalog), and having a way for the controller to specify what the latest version (or the time it last fetched it) and having a broker respond with no changes would make it easier for controller writers to just check the response without doing diffs on existing vs. new catalog.
Seems like with the if-match and if-not-modified-since or if-not-modified-since as optionally being allowed to be returned by a broker would achieve this.

fmui:
Closes #209: "Expand spec to allow for "no changes" semantics to be returned by brokers"

edit: changed to closes so the issue gets closed automatically

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -397,7 +397,7 @@ likelihood of having compatibility with any Platform.
 
 Service Brokers and Platforms MAY support the
 [`ETag`](https://tools.ietf.org/html/rfc7232#section-2.3) and
-[`If-Modified-Since'](https://tools.ietf.org/html/rfc7232#section-3.3) 
+[`If-Modified-Since`](https://tools.ietf.org/html/rfc7232#section-3.3) 
 HTTP headers to enable caching of the catalog.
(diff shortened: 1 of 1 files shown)


mattmcneeney:
@jberkhahn This sounds like it might be useful for svc-cat. Do you think you'd add this to the backlog of things?

jberkhahn:
this sounds useful for us, yes

mattmcneeney:
I think this looks good, though I'm not sure when the CF folks would be able to get around to adding support for this. But it's more useful for svc-cat anyway since CF doesn't periodically poll catalog endpoints today.

fmui:
Review please @jberkhahn @tinygrasshopper

PR #701, opened 2020-03-05
Binding rotation

fmui:
This PR adds the binding rotation concept as discussed on the last OSB summit.

**Checklist:**
- [X] The [swagger.yaml](swagger.yaml) doc has been updated with any required changes
- [X] The [openapi.yaml](openapi.yaml) doc has been updated with any required changes

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -607,6 +607,8 @@ components:
           $ref: '#/components/schemas/Metadata'
         dashboard_client:
           $ref: '#/components/schemas/DashboardClient'
+        binding_rotatable:
+          type: boolean
         plan_updateable:
           type: boolean
         plans:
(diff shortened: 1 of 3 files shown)


mattmcneeney:
Thanks for writing this up @fmui. This looks like a great start!

Do we think we need two different request sections outlined for `PUT /v2/service_instances/:guid/service_bindings/:guid`? This feels a little funky for me, but I guess it may be clearer than having spec readers have to understand that it's one field OR the others.

lyubo1:
Thanks @fmui, LGTM.

rsampaio:
I gave this a try and implemented a `rotate` command in https://github.com/rsampaio/rocs in a branch and it provides a nice interface to request a binding rotation, I used a fake broker and the workflow is very similar to requesting a new binding, this made me wonder about extensions for bindings and I believe the trigger of rotation could be an extension in the future that can do more than just create a new binding and for now, this should cover the basic use cases of expiring a credential.

Issue #683, opened 2019-06-25
Proposal: Interpreting Bindings


tinygrasshopper:
**What is the problem?**
A description of what the problem is.

Currently the binding response from the service broker is not structured, brokers can returns anything in binding response, which is sometimes challenging for application developers to interpret. 

**Who does this affect?**
Broker authors and application developers

**Do you have any proposed solutions?**
Lets introduce a concept of a "Kind" for a service, which can help us predict/interpret the binding response 

**Additional context**
I have created a proposal here: https://docs.google.com/document/d/1VTDc_uAKOtIpYZCYMruGVHn2Gj7cKeyuTP9fmDwp46M/edit# please feel free to leave comments

Similar to 
https://github.com/openservicebrokerapi/servicebroker/issues/116

PR #704, opened 2020-03-13
Document existing CF credential_client_id bind ressource

gberche-orange:

credential_client_id is provided by cloudfoundry cloudcontroller in the binding resource since late 2017. See background at https://www.pivotaltracker.com/n/projects/966314/stories/151111701

**Checklist:**
- [ ] The [swagger.yaml](swagger.yaml) doc has been updated with any required changes
  - Existing profile fields seems not described in the current [swagger.yml ServiceBindingResourceObject](https://github.com/openservicebrokerapi/servicebroker/blob/765742d26dfa24680d1a62f7c7736cf48f1e4036/swagger.yaml#L729-L735)
- [ ] The [openapi.yaml](openapi.yaml) doc has been updated with any required changes
   - Existing profile fields seems not described in the current [openapi.yaml ServiceBindingResource](https://github.com/openservicebrokerapi/servicebroker/blob/765742d26dfa24680d1a62f7c7736cf48f1e4036/openapi.yaml#L787-L808)

### DIFF
--- a/profile.md
+++ b/profile.md
@@ -436,7 +436,7 @@ object in which Platforms MAY choose to add additional fields.
 
 ### Cloud Foundry Bind Resource Object
 
-The Cloud Foundry Bind Resource Object supports the Cloud Foundry [Service Binding](https://docs.cloudfoundry.org/devguide/services/application-binding.html) and the [Service Key](https://docs.cloudfoundry.org/devguide/services/service-keys.html) developer requests. The credentials returned by the Service Binding endpoints MAY be references to credentials stored in [CredHub](https://docs.cloudfoundry.org/credhub/index.html), the component designed for centralized credential management in Cloud Foundry. Service Brokers are then responsible for writing credentials into CredHub and granting read permissions to a bound application or a specified client. See more details into [Credhub documentation](https://github.com/cloudfoundry-incubator/credhub/blob/master/docs/secure-service-credentials.md#service-brokers). The provisioning of CredHub access for service brokers is performed out of band and is out of scope of this specification. 
+The Cloud Foundry Bind Resource Object supports the Cloud Foundry [Service Binding](https://docs.cloudfoundry.org/devguide/services/application-binding.html) and the [Service Key](https://docs.cloudfoundry.org/devguide/services/service-keys.html) developer requests. The credentials returned by the Service Binding endpoints MAY be references to credentials stored in [CredHub](https://docs.cloudfoundry.org/credhub/index.html), the component designed for centralized credential management in Cloud Foundry. Service Brokers are then responsible for writing credentials into CredHub and granting read permissions to a bound application or a specified client. See more details into [Credhub documentation](https://github.com/cloudfoundry-incubator/credhub/blob/master/docs/secure-service-credentials.md#service-brokers). The provisioning of CredHub access for Service Brokers is performed out of band and is out of scope of this specification.
 
 The following properties are defined for usage within a Cloud Foundry
 deployment:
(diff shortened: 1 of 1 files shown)


mattmcneeney:
Thanks for adding this to the spec @gberche-orange !

There seem to be some legitimate travis failures to investigate when you have the change: https://travis-ci.org/github/openservicebrokerapi/servicebroker/builds/661911797?utm_source=github_status&utm_medium=notification

gberche-orange:
thanks @mattmcneeney for your review and heads up on failed test.

I've fixed the formatting violations reported by the build.
Remains the following error in [build 663828758](https://travis-ci.org/github/openservicebrokerapi/servicebroker/builds/663828758?utm_source=github_status&utm_medium=notification) which I assume is false negative since the PR is not touching schemas
> Errors:
>	- Could not find components/schemas/# in contents of http://json-schema.org/draft-04/schema

mattmcneeney:
Yep you’re right. I’ve opened a PR to fix that so once that’s merged I’ll retrigger this build!

tinygrasshopper:
@gberche-orange the PR makes sense, a quick question how would the broker discover the url of the credhub its supposed to put credentials in. OR that does not come thru the spec, its configured on the broker

gberche-orange:
@tinygrasshopper thanks for the heads up. My reverse engineering of the presence of this field was incomplete (I haven't yet used this credhub ref feature but I'm keen to have it documented in the spec as I'm still observing it in OSB payloads my service brokers receive). 

Following a deeper read of https://www.pivotaltracker.com/n/projects/966314/stories/151111701 and credhub documentation, I've updated the PR to introduce the CloudFoundry service binding, service key and credhub reference concepts, and make it clear that the provisioning of credhub url and authN/authZ to Service Brokers is performed out of band.

It might make sense to add to the specs the credential format when using credhub reference, likely in #683 or #116

mattmcneeney:
@gberche-orange The way I read this change now is that Service Brokers may return a credhub reference for a 'service key' request, but not for a normal 'service binding'. Is that the correct reading or am I missing something?

gberche-orange:
@mattmcneeney 

> The way I read this change now is that Service Brokers may return a credhub reference for a 'service key' request, but not for a normal 'service binding'. Is that the correct reading or am I missing something?

Yes, this is correct.

For a service binding, service brokers should grant the bound app (through its provided appguid) permission to read the CredHub reference

For a service key, it is instead assumed that the credential will be looked up by the cloudcontroller (or another credhub client), and therefore the specified credhub client id should be granted permission to read the CredHub reference

See 
https://www.pivotaltracker.com/story/show/151111701/comments/180009415

> verified service key request contains client
>
>   2017-09-21T16:21:11.11-0700 [APP/PROC/WEB/0] OUT INFO: Request body: {"service_id":"f929f899-1818-4fc2-9454-f8def6c3f0e8","plan_id":"1faef310-2c21-4a1a-88ce-1fe06df77f76","bind_resource":{"credential_client_id":"cc_service_key_client"},"context":{"platform":"cloudfoundry","organization_guid":"d65a9c3f-0507-447c-bd4b-2e63c8088bfd","space_guid":"524df9a7-2236-4793-b0da-606066c6a214"}}
>
> verified bind request does not contain client
>
>   2017-09-21T16:22:15.18-0700 [APP/PROC/WEB/0] OUT INFO: Request body: {"service_id":"f929f899-1818-4fc2-9454-f8def6c3f0e8","plan_id":"1faef310-2c21-4a1

Does it make sense ?

Maybe @zrob which contributed this feature back in 2017 can provide more background if necessary

gberche-orange:
@mattmcneeney 

Sorry,credhub references can be returned in both SB and SK, but the client_id is included in the bind resource for SK. I'll double check the wording

gberche-orange:
@mattmcneeney 

Double reading the proposed changed, I did not spot where the could have an ambiguity about service binding responses not supporting credhub references. Can you please use github review feature to point the specific line and if possible suggest a better wording ?

mattmcneeney:
I mistook the `bind_resource` object being part of the response rather than the request. I think this actually makes sense to me now! Thanks for the explanation

mattmcneeney:
Trying to retrigger Travis

mattmcneeney:
@tinygrasshopper / @rsampaio if one of you can approve this one before Friday we can try to include it in the release!

PR #707, opened 2020-04-20
Service Instances can be labelled with information defined by the Broker

dlresende:

In this Pull Request, we introduce a new `ServiceInstanceMetadata` object that allows the Broker to associate specific `metadata`, more specifically `labels`, with Service Instances. By doing that Service Instances can be labelled with information defined by the Broker. The Platform can update its view of the `labels`. Service Instances with specific `labels` may be identified by Users via the Platform, in order to perform specific actions against those instances.

As an example, let's consider a MySQL Service Offering containing multiple Service Plans. A particular Service Plan combined with a specific set of parameters used in the Provision request could produce a Service Instance eligible for a [multi-source replication topology](https://dev.mysql.com/doc/refman/5.7/en/replication-multi-source.html). A User might want to take advantage of this feature and connect multiple MySQL Service Instances together. However, as not all existing MySQL Service Instances support multi-source replication, the User should be able to identify and select the eligible ones. This use case can be made possible by giving the Broker the ability to associate specific `labels` with Service Instances during a Provision or Update request, which will be returned to the Platform. The Platform can then store and make those `labels` available to the User, allowing them to identify and select the specific Service Instances.

The `ServiceInstanceMetadata` object follows a precedent established by the existing [BindingMetadata](https://github.com/dlresende/servicebroker/blob/master/spec.md#binding-metadata-object) object. We propose the creation of the `ServiceInstanceMetadata` object at the heart of the specification as opposed to a profile given the fact Platforms such as [Kubernetes](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/) and [Cloud Foundry](https://docs.cloudfoundry.org/adminguide/metadata.html#types) already have this concept.

**Checklist:**
- ✅ The [swagger.yaml](swagger.yaml) doc has been updated with any required changes
- ✅ The [openapi.yaml](openapi.yaml) doc has been updated with any required changes

@FelisiaM, @addytripathi & Diego

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -720,6 +720,8 @@ components:
       properties:
         dashboard_url:
           type: string
+        metadata:
+          type: object
 
     ServiceInstanceAsyncOperation:
       type: object
@@ -728,6 +730,8 @@ components:
           type: string
         operation:
           type: string # could be a link object to last operation
+        metadata:
+          type: object
 
     ServiceInstanceUpdateRequest:
       type: object
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -721,7 +721,7 @@ components:
         dashboard_url:
           type: string
         metadata:
-          type: object
+          $ref: '#/components/schemas/ServiceInstanceMetadata'
 
     ServiceInstanceAsyncOperation:
       type: object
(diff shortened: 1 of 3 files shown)


mattmcneeney:
Thanks @dlresende, this looks good to me!

ping @openservicebrokerapi/osbapi-pmc

mattmcneeney:
ping @fmui

Issue #709, opened 2020-05-14
Define a preferred async provision check


ransombriggs:
**What is the problem?**

We were discussing the semantics of 200 during provisioning and binding and were unsure why create would be called a second time since the :instance_id MUST be a globally unique non-empty string and we are using orphan deletion to mitigate creation issues.  I then found out why in a the 202 accepted definition.

> This triggers the Platform to poll the Last Operation for Service Instances endpoint for operation status. Note that a re-sent PUT request MUST return a 202 Accepted, not a 200 OK, if the Service Instance is not yet fully provisioned.

From the way that this reads there appears to be two methods for checking if something is successfully provisioned or bound, either pool `last_operation` until success or call put until 200.  Our provider implementation will support both for interoperability, but we are wondering which one we should prefer for our platform implementation.

**Who does this affect?**

platform authors

**Do you have any proposed solutions?**

The specification description should say whether `last_operation` or PUT until 200 is the preferred way to detect if something is fully provisioned.

ransombriggs:
Had some further discussions on our teams and realized that we are never retrying on failure, but instead cleaning up orphans, but that some platforms *could* retry instead of doing orphan removal.  Having a best practice around this for platforms re-using instance_id would be good as well.  Additionally an explanation of why a synchronous provision that returns a 201 may get a second PUT request would also be helpful for providers.

gberche-orange:
> From the way that this reads there appears to be two methods for checking if something is successfully provisioned or bound, either pool last_operation until success or call put until 200. Our provider implementation will support both for interoperability, but we are wondering which one we should prefer for our platform implementation.

I understand that the requirement to support duplicate PUT returning as a 200 is to account for OSB client platforms that use eventual consistency (instead of atomic transactional consistency) and therefore make multiple provisionning calls in parallel (namely the K8S svcat see https://github.com/kubernetes-sigs/service-catalog/issues/1639).

It is possible that brokers have optimized for the canonical case, and only support K8S svcat duplicate calls with some performance impact (additional calls and response latency). This is how I dealt with in in my [osb facade](https://github.com/orange-cloudfoundry/osb-cmdb), see details in section `Handle race conditions (including for K8S dups)` of [implemention notes](https://github.com/orange-cloudfoundry/osb-cmdb/blob/master/osb-cmdb/docs/impl-notes/v10-redesign.md).

[...]

PR #717, opened 2020-06-09
Remove ambiguity on provisioning status code

rsampaio:

The specification for synchronous provisioning says `201 Created` status code
should be returned for provisioning and bind and `200 OK` only for
the update, unbind, and deprovision requests. This change should remove ambiguity on the provisioning and binding requests and converge to a single way to verify if a service instance has completed the provisioning step by calling the `last_operation` endpoint.

This also reverts #561 and assumes duplicated requests will return `409
Conflict` for the same unique UUID and it should be advised to platforms
to start polling the `last_operation` endpoint to wait for completion.

**Checklist:**

Fixes Issue #709

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -676,11 +676,12 @@ Offering, even by Service Plan.
 To execute a request synchronously, the Service Broker need only return the
 usual status codes: `201 Created` for provision and bind, and `200 OK` for
 update, unbind, and deprovision. `200 OK` MAY be returned synchronously by
-Service Brokers that accept provisioning requests with unchanged parameters.
+Service Brokers that accept provisioning requests with unchanged parameters
+and synchronous operations.
 
 Platforms that rely on eventual consistency are RECOMMENDED to use `409 Conflict`
 as a no-op response and SHOULD use the `last_operation` endpoint to verify the status of
-the provisioning request instead.
+the provisioning request instead in asynchronous mode.
 
 Service Brokers that support synchronous responses for provision, update, and
 delete can ignore the `accepts_incomplete=true` query parameter if it is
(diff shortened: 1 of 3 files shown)


Samze:
I agree the spec looks inconsistent right now with this statement.

`
To execute a request synchronously, the Service Broker need only return the usual status codes: 201 Created for provision and bind, and 200 OK for update, unbind, and deprovision.
`

However I think replacing `200` with `409` for service brokers that receive a request for an instance that already exist would be a breaking change. For example if a newly written broken with this guidance returned a `409` then older Cloud Foundry platforms would interrupt that as an error and not a no-op but fail instead. 

I wonder if instead of removing `200`, we should change that statement to include `200` as a valid response for a synchronous provision/bind? And then perhaps this is something that can be cleaned up in a future version of the spec?

spec.md
```diff
@@ -692,6 +692,10 @@ To execute a request synchronously, the Service Broker need only return the
 usual status codes: `201 Created` for provision and bind, and `200 OK` for
 update, unbind, and deprovision.
 
+Platforms that rely on eventual consistency MAY consider `409 Conflict` a
+a no-op and SHOULD use the `last_operation` endpoint to verify the status of
```
fmui, line 696: Typo: "...MAY consider `409 Conflict` **a a** no-op and..."

fmui:
I agree with @Samze, this is probably a breaking change for v2.x. We cannot remove the `200` and with that a (probably unused) "feature". We can discourage its use, though.

rsampaio:
@Samze I added some clarifying points around the synchronous requests and the use of 200 code by existent brokers. I was also reflecting on the goal of this PR and I believe the intent is to encourage the use of `last_operation` for provisioning status and the use of `409 Conflict` for duplicated requests (with the same UUIDs and parameters) since it is more aligned with the HTTP semantics and ultimately encourage the adoption of asynchronous patterns instead of synchronous across the board. I made sure to mention that returning `200 OK` is still supported but not recommended.

I'm curious to hear your thoughts.

fmui (approved):

Samze (commented):
Sorry for the delayed response @rsampaio!

Just so I understand, the goal of this is to encourage platforms to use the broker response of 409 conflict mechanism instead of 200? I think maybe I'm not quite getting how this works for the synchronous case - I think `This response is deprecated in favor of the last_operation endpoint` is throwing me off.

The asynchronous case:
1. Platform provision to broker, broker responses 202
1. Platform provision to again broker with same request, broker returns 409
1. Platform hits last operation endpoint....continues until success

The synchronous case
1. Platform provision to broker, broker responses 201
1. Platform provision to again broker with same request, broker returns 409
1. Platform hits last operation endpoint.. ?
In this case the broker initially returned 201, so may not support the last operation endpoint.

Perhaps we should deprecated 200 in favour of 200 response? The only issue is that currently 409 and 200 are distinguished by 409 having different parameters and 200 having the same parameters.

rsampaio:
No worries @Samze!

> Just so I understand, the goal of this is to encourage platforms to use the broker response of 409 conflict mechanism instead of 200? I think maybe I'm not quite getting how this works for the synchronous case - I think `This response is deprecated in favor of the last_operation endpoint` is throwing me off.

That is correct and it only applies to the asynchronous case.

> The asynchronous case:
> 
> 1. Platform provision to broker, broker responses 202
> 2. Platform provision to again broker with same request, broker returns 409
> 3. Platform hits last operation endpoint....continues until success

Exactly!


> The synchronous case
> 
> 1. Platform provision to broker, broker responses 201
> 2. Platform provision to again broker with same request, broker returns 409
> 3. Platform hits last operation endpoint.. ?
>    In this case the broker initially returned 201, so may not support the last operation endpoint.

In this case, when the broker returns a 201 the service instance should be already created and there is no reason to call `last_operation` at all, the change is really intended to clarify the response of requests containing the same uuid for provisioning


> Perhaps we should deprecated 200 in favour of 200 response? The only issue is that currently 409 and 200 are distinguished by 409 having different parameters and 200 having the same parameters.

Maybe you mean 200 in favor of 201 and I agree, I believe the broker should return 409 for duplicated uuid and not only for duplicated parameters.

fmui (dismissed):

rsampaio:
@Samze just rebased and fixed the conflict, linter was complaining about a table and I fixed that as well, this is definitely ready to go.

@fmui thanks for the review!

fmui (approved):

tinygrasshopper (approved):

Samze (approved):

Issue #720, opened 2020-07-23
Service Instance response doesn't include context on fetch


rsampaio:
**What is the problem?**

The response of fetching information from Service Instance doesn't include the context field, context can be generated by the Service Broker and can contain useful information for the platform.

An example we have in our platform is a customer using a private plan:
* Customer provision a private plan
* Broker creates a VPC for the service instance and communicate the `service_name` back so customers can create a private endpoint
* Since this is context information generated by the broker and not provided by the User it doesn't make sense to return this value in `parameters`.

**Who does this affect?**
Service Broker Authors, Platforms and Developers

**Do you have any proposed solutions?**
A potential solution would be to include the field `context` on the response of fetching a Service Instance, I create a Pull Request with this change which I will link to this issue shortly.

**Additional context**
In addition to the `context` field, I wonder if the response should also include the `metadata` provided in the creation of the service instance as well as information about maintenance

gberche-orange:
@rsampaio should this issue be closed now that https://github.com/openservicebrokerapi/servicebroker/pull/721 got merged ? It seems the PR was missing the `close` keyword for closing it issue automatically

PR #721, opened 2020-07-23
Add attributes field to metadata for broker generated information

rsampaio:
This PR adds the attributes field to the response of fetching a service instance, more details can be found in this issue: https://github.com/openservicebrokerapi/servicebroker/issues/720

**Checklist:**


spec.md
```diff
@@ -1076,6 +1076,7 @@ For success responses, the following fields are defined:
 | plan_id | string | The ID of the Service Plan from the catalog that is associated with the Service Instance. |
 | dashboard_url | string | The URL of a web-based management user interface for the Service Instance; we refer to this as a service dashboard. The URL MUST contain enough information for the dashboard to identify the resource being accessed (`9189kdfsk0vfnku` in the example below). Note: a Service Broker that wishes to return `dashboard_url` for a Service Instance MUST return it with the initial response to the provision request, even if the service is provisioned asynchronously. |
 | parameters | object | Configuration parameters for the Service Instance. |
+| context | object | Context information provided on provisioning and/or generated by Service Brokers for the Service Instance. |
```
tinygrasshopper, line 1079: I think the use of `context` here is confusing as we use `context` for metadata sent from platform to broker

Samze:
We chatted about this on the call today.

General:
* The spec [already allows brokers and platforms to send arbitrary](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#vendor-extension-fields) fields to eachother off-spec. So this change just gives us a `context` container to place them?
* Do you imagine that we document the fields inside this `context` liek we do in [profile.md](https://github.com/openservicebrokerapi/servicebroker/blob/master/profile.md#context-object)
* Does your usecase for this (setting up network connectivity) be solved by another mechanism? e.g. The networking proposal by @fmui.

Specific:
* We think it should be called something other than `context`, to avoid confusing with the platform to broker `context` object that exists today.
* If we are adding it to Fetch Instance endpoint, we should probably allow brokers to return it as part of the provision response too.

It would be good to chat through these on the next call.

rsampaio:
@Samze Thanks for the feedback!

I agree that the term `context` is overused and even in the spec it has many meanings and the main motivation for adding this is field is to be able to use clients generated from the openapi.yaml to parse this information from brokers as an object. 

> The spec already allows brokers and platforms to send arbitrary fields to eachother off-spec. So this change gives us a context container to place them inside to avoid future naming collisions?
Name collision is a good point and indeed requires a different name than `context` to avoid that, will push a different name suggestion in a moment.

 > Do you imagine that we document the fields inside this context like we do in profile.md? (There are a lot more brokers than platforms so this may be difficult unless we choose some specific fields platforms know about) 
Yes, that is a great idea, with a different name like `attributes` we can definitely identify common fields that platforms are interested and document the use, it is a potential tool to create a platform spec.

> Does your use case for this (setting up network connectivity) be solved by another mechanism? e.g. The networking proposal by @fmui.
Possibly and I believe it is still something we should consider for more complex network setups like VPNs and VPC peering and I believe the field can still be useful for this proposal and still enable information passing.

I propose a  rename to `attributes` since it describes better the intent and avoid collision with user-provided fields and will push changes to this PR shortly to rename the field and also include this field when provisioning is made.

spec.md
rsampaio: @tinygrasshopper I agree, `context` is not a good name for this and I propose something like `attributes` and I updated this PR to reflect this change 👍

rsampaio: I still believe that context should be part of the response and should include the values provided by the client.

spec.md
```diff
@@ -1076,7 +1081,8 @@ For success responses, the following fields are defined:
 | plan_id | string | The ID of the Service Plan from the catalog that is associated with the Service Instance. |
 | dashboard_url | string | The URL of a web-based management user interface for the Service Instance; we refer to this as a service dashboard. The URL MUST contain enough information for the dashboard to identify the resource being accessed (`9189kdfsk0vfnku` in the example below). Note: a Service Broker that wishes to return `dashboard_url` for a Service Instance MUST return it with the initial response to the provision request, even if the service is provisioned asynchronously. |
 | parameters | object | Configuration parameters for the Service Instance. |
-
+| context | object | Context information provided on provisioning for the Service Instance. |
```
fmui, line 1084: Why should the broker reflect the context object back to the platform? It should know the content anyway. Additionally, it would require that brokers must store the original context object, which makes building stateless brokers even more difficult.

rsampaio: @fmui indeed this is a good point for stateless brokers and would force them to store data. I will remove it for now and stick to attributes only in this PR.

rsampaio: @fmui isn't the case for the parameters as well? it definitely requires the broker to reflect back and I assume it would have to save information as well.

spec.md
```diff
@@ -1004,10 +1004,15 @@ For success responses, the following fields are defined:
...
 | operation | string | For asynchronous responses, Service Brokers MAY return an identifier representing the operation. The value of this field MUST be provided by the Platform with requests to the [Polling Last Operation for Service Instances](#polling-last-operation-for-service-instances) endpoint in a percent-encoded query parameter. If present, MAY be null, and MUST NOT contain more than 10,000 characters. |
 | metadata | [ServiceInstanceMetadata](#service-instance-metadata) object | An OPTIONAL object containing metadata for the Service Instance. |
+| attributes | object | An OPTIONAL object containing attributes generated by the Broker, attributes will be available in the response body when fetching information about a Service Instance. |
 ```
 {
   "dashboard_url": "http://example-dashboard.example.com/9189kdfsk0vfnku",
   "operation": "task_10",
+  "attributes": {
+    "platform.generated.attribute1": "value1",
```
fmui, line 1013: Shouldn't these be called `broker.generated.attributeX`?

spec.md
```diff
@@ -1076,7 +1081,7 @@ For success responses, the following fields are defined:
 | plan_id | string | The ID of the Service Plan from the catalog that is associated with the Service Instance. |
 | dashboard_url | string | The URL of a web-based management user interface for the Service Instance; we refer to this as a service dashboard. The URL MUST contain enough information for the dashboard to identify the resource being accessed (`9189kdfsk0vfnku` in the example below). Note: a Service Broker that wishes to return `dashboard_url` for a Service Instance MUST return it with the initial response to the provision request, even if the service is provisioned asynchronously. |
 | parameters | object | Configuration parameters for the Service Instance. |
-
+| attributes | object | Attributes generated by the platform relevant to the Service Instance. This field will only have attributes generated by the platform and cannot be updated. |
```
fmui, line 1084: Attributes are generated by the broker. Right?

rsampaio: that is correct @fmui

spec.md
rsampaio: Yes good catch, let me fix this.

spec.md
rsampaio: I just fixed this sentence, thanks and good catch!

spec.md
```diff
@@ -1004,10 +1004,15 @@ For success responses, the following fields are defined:
...
 | operation | string | For asynchronous responses, Service Brokers MAY return an identifier representing the operation. The value of this field MUST be provided by the Platform with requests to the [Polling Last Operation for Service Instances](#polling-last-operation-for-service-instances) endpoint in a percent-encoded query parameter. If present, MAY be null, and MUST NOT contain more than 10,000 characters. |
 | metadata | [ServiceInstanceMetadata](#service-instance-metadata) object | An OPTIONAL object containing metadata for the Service Instance. |
+| attributes | object | An OPTIONAL object containing attributes generated by the Broker, attributes will be available in the response body when fetching information about a Service Instance. |
```
Samze, line 1007: Do you imagine returned attributes would impact platform behaviour changes? If not we could use the existing `metadata.labels` functionality, if the data is ok to be exposed by users.

If `metadata.labels` isn't suitable then I suggest we still key attributes under `metadata` and add some text on when it makes sense to use metadata.attributes vs metadata.labels.

rsampaio: @Samze In the use-case mentioned in this PR the platform would receive a piece of important information in this field to complete the network handshake, this is the case for AWS PrivateLinks and the information we are sending is the `aws_service_name` and I imagine this being used by services that require a second step using information generated by the broker. I believe we could use metadata for this and we would have to add the `object` type to the definition since `metadata.labels` is not suitable in my opinion. Is metadata a better place for this information? I tried to avoid sticking important information as `metadata` and I am concerned that it will mix important information with `required` information.

rsampaio: @Samze I was just looking at how we return `metadata` and looks like we never return as part of the `ServiceInstanceResource`, metadata is only present in the response of the provisioning. We could use metadata but then I believe we should normalize and include it in the response of both the resource and the provisioning.

Samze: > ...a piece of important information in this field to complete the network handshake, this is the case for AWS PrivateLinks and the information we are sending is the aws_service_name and I imagine this being used by services that require a second step using information generated by the broker. 

Makes sense that `metadata.labels` isn't suitable if this is non-user facing information. I would be in favour of still placing it under `metadata` with a different key, seems redundant to have two places for this "broker" additional information,.

> @Samze I was just looking at how we return `metadata` and looks like we never return as part of the `ServiceInstanceResource`, metadata is only present in the response of the provisioning. We could use metadata but then I believe we should normalize and include it in the response of both the resource and the provisioning.

Goodpoint, I think `metadata` makes sense to be returned as part of GET on the instance resource. It is at the same level

rsampaio: @Samze I believe I completed moving everything to `metadata` and I added a section explaining what attributes are supposed to mean in this context and I believe this is ready.

fmui (approved):

fmui (approved):

Samze (approved):

tinygrasshopper (approved):

PR #722, opened 2020-07-28
Add maintenance info to the response of fetching a service instance

### LINKED ISSUE #714: Support MaintenanceInfo refresh (e.g. triggered by broker-side maintenance)
**What is the problem?**

As a service consumer (a.k.a developer):
* In order to trigger automatic maintenance upgrade of a service instance outside business hours
* I need to be able to specify a scheduled maintenance window for my service
* And that the platform displays correctly only remaining candidate upgrades after a scheduled upgrade was done unattended

As a service brother author:
* In order to support both scheduled maintenance windows, and user-triggered upgrades
* I need the `GET /v2/service_instances/:instance_id` endpoint to return a `maintenance_info` field that platforms can regularly poll to discover about out-of-brands updates triggered.

**Who does this affect?**

This affect service broker authors, and platform authors (this would be mostly transparent to developers)

The `GET /v2/service_instances/:instance_id` endpoint to return a `maintenance_info` field that platforms can regularly poll to discover about out-of-brands updates triggered.

**Additional context**

Having the platform poll the  `GET /v2/service_instances/:instance_id` endpoint will also enable service broker upgrades to return an *updated* dashboard url without requiring a user to explicitly trigger a service instance upgrade.

gberche-orange:
Also provides rationale for platform to poll the "fetching a service instance" endpoint

closes #714

**Checklist:**


rsampaio (dismissed):

spec.md
```diff
@@ -1076,6 +1078,7 @@ For success responses, the following fields are defined:
 | plan_id | string | The ID of the Service Plan from the catalog that is associated with the Service Instance. |
 | dashboard_url | string | The URL of a web-based management user interface for the Service Instance; we refer to this as a service dashboard. The URL MUST contain enough information for the dashboard to identify the resource being accessed (`9189kdfsk0vfnku` in the example below). Note: a Service Broker that wishes to return `dashboard_url` for a Service Instance MUST return it with the initial response to the provision request, even if the service is provisioned asynchronously. |
 | parameters | object | Configuration parameters for the Service Instance. |
+| maintenance_info | [Maintenance Info](#maintenance-info-object) | If a Service Broker provides maintenance information for a Service Plan in its [Catalog](#catalog-management), a Broker MAY return the maintenance information currently associated with the Service Instance. Any field except for `maintenance_info.version` should be ignored by the Platform. This field can be used by the Platform to inform the end-user of unattended maintenances that were triggered out-of-band and applied to the Service Instance. |
```
alexwo, line 1081: - It's a little bit confusing that **maintenance_info** will indicate "available upgrades..?" in the context of getting service instance details. normally incase the maintenance_info diverge it's expected for the platform to upgrade the service catalog, as detailed in:
If a Service Broker provides maintenance information for a Service Plan in its Catalog, a Platform MAY provide the same maintenance information when provisioning a Service Instance. Any field except for maintenance_info.version will be ignored. This field can be used to ensure that the end-user of a Platform is provisioning what they are expecting since maintenance information can be used to describe important information (such as the version of the operating system the Service Instance will run on). If a Service Broker's catalog has changed and new maintenance information version is available for the Service Plan being provisioned, then the Service Broker MUST reject the request with a 422 Unprocessable Entity as described below.

- In case there are multiple upgrade versions, it can be useful to return a list of all upgrade options
- Perhaps it's more efficient to query a new dedicated endpoint that will return all-new available upgrade options by service plan catalog id (avoid going to every instance)
- the platform can query this endpoint during the selected maintenance window
- Getting the current instance version of the provisioned instance is interesting to determine what version is currently used?

I have prepared a modified proposal to support unattended service instance updates.
https://github.com/openservicebrokerapi/servicebroker/issues/726

lease let me know if can addresses your current use case at:
https://github.com/openservicebrokerapi/servicebroker/issues/714

gberche-orange: @alexwo thanks for your feedback

> It's a little bit confusing that maintenance_info will indicate "available upgrades..?" in the context of getting service instance details. normally incase the maintenance_info diverge it's expected for the platform to upgrade the service catalog

Is the mismatch only related to the naming of this field being `maintenance_info.version` instead of something like `current_version` ? I was trying to reuse the existing Maintenance_Info structure and naming, hence my initial suggestion in this PR.

> In case there are multiple upgrade versions, it can be useful to return a list of all upgrade options

The `GET /v2/service_instances/:instance_id` endpoint returns the current state of the service instance to platforms. It is then up to the platform to compare it against the catalog endpoint to infer available upgrades. Could you please share specific examples/use cases for supporting multiple upgrades surfaced to users ? This is currently out of scope of this focused PR.

> Getting the current instance version of the provisioned instance is interesting to determine what version is currently used?

Yes precisely, and avoid surfacing to users upgrades that were already applied off band by brokers themselves.


> I have prepared a modified proposal to support unattended service instance update:
#726

If I understand correctly you're suggesting in #726 to push to the broker the responsibilities of comparing maintenance_info versions between current in a service instance and available in catalog. This seems to me #726 is completely changing the maintenance_info spec well beyond unattended version updates triggered by **brokers** (without platform involvments).

alexwo: You misunderstood me. 

The OpenServiceBroker specification doesn't intend for the broker to be able to do unattended-upgrades. Reading the spec carefully you will see that it explicitly forces update operations to fail if the maintenance_info structure changes in the Catalog relative to the version specified in the service instance.

What #726 attempts to do is add the required endpoints to the Broker specification so that the Platform can manage unattended-upgrades. The platform is the buffer between the developer and the broker.  It is responsible for triggering operations on the broker when needed. #726 adds an endpoint for the broker to expose available upgrade information to the platform, and also a way for the platform to request an upgrade operation be triggered.

gberche-orange: I understand that Platforms can already perform unattended upgrades with 2.16 spec version: available upgrades are today inferred by platforms (see see https://docs.cloudfoundry.org/devguide/services/managing-services.html#upgrade for CF)  by observing the existing 2.16 catalog endpoint and keeping history of maintenance_info they passed to the provisionning endpoint. Platform may dedide to offer user opt-in for unattended upgrades without changes to the OSB specs.

My use case at #714 is indeed related to broker-triggered unattended maintenance, likely according to use opt-ins that specify maintenance windows in their service instance params. 

For example, an AWS RDS service instance would surface to user the [RDS managed service maintenance window](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_UpgradeDBInstance.Maintenance.html), allowing some upgrades to be user triggered, and some other to be brokered-triggered

alexwo: Thanks for your feedback 👍 

1. The **maintenance_info.version**, can be returned as part of the get service instance API call to indicate a new version is currently running on a particular service instance? 

2. To detect **"out-of-brands"** update done by a broker, a platform should poll the service instance endpoint to compare the maintenance_info.version with a cached copy of the last instance maintenance_info.version (can be fetched from the catalog on instance create/update) what will allow it to display only the remaining candidates for an upgrade after a scheduled upgrade was done unattended by the broker?

gberche-orange: You're welcome.

> The maintenance_info.version, can be returned as part of the get service instance API call to indicate what version is currently running on a particular service instance?

yes

> To detect "out-of-brands" update done by a broker, a platform should poll the service instance endpoint to compare the maintenance_info.version with a cached copy of the last instance maintenance_info.version (can be fetched from the catalog on instance create/update).

yes

> this should allow to display only the remaining candidates for an upgrade after a scheduled upgrade was done unattended by the broker?

yes

alexwo (approved):
LGTM

fmui (approved):

Samze (approved):

Samze:
LGTM

spec.md
Samze, line 1081: should -> SHOULD

Samze (changes requested):
Just need to capitalise SHOULD

spec.md
gberche-orange: thanks. This is fixed in https://github.com/openservicebrokerapi/servicebroker/pull/722/commits/da16ac134308fcfe441ece0c5402fb36f835f216

Samze (approved):

fmui (approved):

rsampaio (approved):

PR #732, opened 2020-12-09
Orphan mitigation should be performed for 202 malformed responses as …

pivotal-marcela-campo:
…the platform would not be able to poll if the broker requires an operation to be sent back in the last_operation request.

As from OSB API spec 2.14 brokers may return an operation id for async requests. That operation id should be included in subsequent requests to fetch last_operation.
In the event of a 202 response, if the response body is malformed, then the platform cannot get the operation id to send back when fetching last_operation and then would find itself unable to keep track of the operation.
Resources might have been created by the broker and these would need to be cleaned up.

**Checklist:**
- [ ] The [swagger.yaml](swagger.yaml) doc has been updated with any required changes - No changes needed
- [ ] The [openapi.yaml](openapi.yaml) doc has been updated with any required changes - No changes needed

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1952,6 +1952,8 @@ by the Platform:
 | [Provisioning](#provisioning) | 201 with malformed response | Failure | Yes | No |
 | [Binding](#binding) | 201 with malformed response | Failure | No | Yes |
 | _All_ | 202 | Success | No | No |
+| [Provisioning](#provisioning) | 202 with malformed response | Failure | Yes | No |
+| [Binding](#binding) | 202 with malformed response | Failure | No | Yes |
 | [Provisioning](#provisioning)/[Deprovisioning](#deprovisioning) | All other 2xx | Failure | Yes | No |
 | [Binding](#binding)/[Unbinding](#unbinding) | All other 2xx | Failure | No | Yes |
 | [Updating a Service Instance](#updating-a-service-instance) | All other 2xx | Failure | No | No |


Samze:
Thanks for raising this @pivotal-marcela-campo. Yeah it looks like 202 malformed is missing.

Out of interest, what is the current behaviour of CF for malformed 202 responses? And @jhvhs what about service-catalog?

As a side note, what constitutes a malformed response is interesting here, since the 202 provision/bind response fields are optional. As we allow additional non-spec keys, the only malformed responses here would non `application/json` or using a key with an invalid value or type. E.g. {"operation": true}. Is that what you had in mind or are there other cases?

Also, I think we should make 202 malformed response consistent with how we represent 201 malformed, with different row entries for provision and binding instead of a single row.

pivotal-marcela-campo:
Thanks @Samze for the feedback. I've made the suggested change.

Current CF behaviour for binding/provisioning is to perform OM mitigation under this scenario (for both `v2` and `v3` api versions).

It seems we are inline as to what constitutes a malformed response. For async `binding` and `provisioning` (202 responses) we are considering a malformed response as one that is either not JSON or doesn't adhere to the expected JSON schema (example [schema](https://github.com/cloudfoundry/cloud_controller_ng/blob/234f065a7445d016a174dbf06e955d81f8d80d45/lib/services/service_brokers/v2/response_parser.rb#L235)). 

Do let me know if you need more information!

Samze (approved):
LGTM thanks @pivotal-marcela-campo

jberkhahn (approved):

jberkhahn:
:+1:

rsampaio (approved):

